PR #388, opened 2014-05-16
Make droproot say something when successful.

wxsBSD:
I've seen people run into situations where they were using a command like this:

tcpdump -i eth0 -G 500 -w /root/%H%M%S.pcap

The first file would be created successfully but the second file would not
because their version of tcpdump was dropping privs. It was unclear to them
that this was going on and was causing confusion.

At least with this message in there it should become more evident that
privs are being altered and aid in debugging these kinds of problems.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -713,6 +713,9 @@ droproot(const char *username, const char *chroot_dir)
 		if (ret < 0) {
 			fprintf(stderr, "error : ret %d\n", ret);
 		}
+		else {
+			printf("dropped privs to %s\n", username);
+		}
 		/* We don't need CAP_SETUID and CAP_SETGID */
 		capng_update(CAPNG_DROP, CAPNG_EFFECTIVE, CAP_SETUID);
 		capng_update(CAPNG_DROP, CAPNG_EFFECTIVE, CAP_SETUID);


infrastation:
Do you mean the droproot message mostly makes sense if the runtime behaviour is different from the compiled-in default? In this case it would be better to consider `WITH_USER` instead of removing the message completely, would you agree?

msekletar:
I was speaking from perspective of distribution maintainer. We (in Fedora) compile --with-user=tcpdump and it is documented in tcpdump manpage that we do so. Hence, I don't think there is a need to tell the user this information every-time he runs the program. And if user requests drop-root by giving -Z explicitly on the command line, then again, there is no need to tell the user we did what he asked for.

In case that some distribution compiles --with-user=<username> but doesn't indicate this in tcpdump man page then I think it is their bug which should be fixed.

guyharris:
> The first file would be created successfully but the second file would not because their version of tcpdump was dropping privs.

"Dropping privs" as in "relinquishing root privileges to the user specified at configure time with `--with-user=`?

If so, then it appears that, as of tcpdump 4.0, root/root-group privileges are dropped *before* the first dump file is created, so, if you specify `-Z`, or if your tcpdump is configured with `--with-user=`, you should not be able to write a dump file to any directory into which the user specified by `-Z`or, if `-Z` isn't specified, `--with-user=` does not have write privileges.

PR #426, opened 2015-02-10
Allow the use of pcap's immediate_mode when TPACKET_V3 is available.

tbeadle:
If the kernel has TPACKET_V3 support, then libpcap, by default, will use it and
packets may not be delivered as soon as they're received.

This patch allows the user to provide a -o (--set-immediate-mode) flag to enable
immediate_mode in case receiving packets ASAP is required.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -58,6 +58,9 @@
 /* Define to 1 if you have the `rpc' library (-lrpc). */
 #undef HAVE_LIBRPC
 
+/* Define to 1 if you have the <linux/if_packet.h> header file. */
+#undef HAVE_LINUX_IF_PACKET_H
+
 /* Define to 1 if you have the <memory.h> header file. */
 #undef HAVE_MEMORY_H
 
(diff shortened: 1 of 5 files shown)


guyharris:
There's nothing Linux-specific about this; it's been an issue practically since tcpdump first existed.  Therefore, this shouldn't be based on whether there's TPACKET_V3 support; it should be based on whether libpcap has `pcap_set_immediate_mode()` or not.

Also:

* it should default to immediate mode if you're printing packets rather than saving them to a savefile and the standard output is a terminal;
* other options don't have "set-" in front of them, so it should just be "--immediate-mode";
* "-o" is used for other purposes by some other versions of tcpdump, so it shouldn't be used for immediate mode;
* the man page needs to be updated as well.

d71793791fa0823f25fa9adaa2bb99797197337f does it that way.

tbeadle:
Thanks for the fast turnaround!  I like your approach better.

PR #429, opened 2015-02-12
Fixed bug in arp_print() in handling of inverse arp responses

mkirkhart:
The target's hardware and protocol address were being printed instead of the sender's hardware and protocol address.

This is now handled the same way as a ARP response.

I found this bug several years ago, but there was not an easy way for me to submit a patch.  Once I found out this project was now being hosted on GitHub, I took a look and found the bug still existed, and decided to submit a fix.

### DIFF
--- a/print-arp.c
+++ b/print-arp.c
@@ -389,8 +389,8 @@ arp_print(netdissect_options *ndo,
 
 	case ARPOP_INVREPLY:
 		ND_PRINT((ndo,"%s at %s",
-			  linkaddr_string(ndo, THA(ap), linkaddr, HRD_LEN(ap)),
-			  ipaddr_string(ndo, TPA(ap))));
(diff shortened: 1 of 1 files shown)


guyharris:
Given that it's printing "{hardware address} at {protocol address}", shouldn't *both* of those be source addresses?  I.e., the machine sending the reply gives its hardware and protocol addresses in the source address fields, indicating that the machine with the specified hardware address (the machine sending the reply) has the protocol address in question.

mkirkhart:
Guy

I believe that is what I did.  THA (target hardware address) was changed to SHA (sender hardware address), and TPA (target protocol address) was changed to SPA (sender protocol address).  Am I missing something?  Would a test capture help?

guyharris:
Sorry, misread the diff.

PR #430, opened 2015-02-12
support MIP (Minimal Interconnection Protocol)

hawken93:
See http://www.uio.no/studier/emner/matnat/ifi/INF3190/v15/oblig/oblig.pdf

This protocol is used in at least the inf3190 data communication course in the University of Oslo. Probably not used outside the university. It's a mini-ip protocol operating on top of Ethernet with ethertype 0xffff. It has it's own ARP via a bit in the header, it has 8-bit addressing and not much more.

It would be fun if tcpdump had support for it :)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -154,6 +154,7 @@ LIBNETDISSECT_SRC=\
 	print-lwapp.c \
 	print-lwres.c \
 	print-m3ua.c \
+	print-mip.c \
 	print-mobile.c \
 	print-mpcp.c \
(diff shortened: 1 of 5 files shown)


infrastation:
Ethertype 0xffff is reserved for different purposes. For local experimental works like this it is correct to use ethertypes 0x88b5 and 0x88b6.

hawken93:
Okay :)

infrastation:
This sounds very reasonable, thank you.

PR #431, opened 2015-02-18
Fix issues with the setting of libcap-ng capabilities.

tbeadle:
- Only set the SETUID/SETGID capabilities if required (i.e. we'll be changing
  the effective UID).
- Only set the CAP_DAC_OVERRIDE capability if writing out to a file (i.e. the -w
  flag was provided).
- Fix the calls to capng_clear to pass CAPNG_SELECT_BOTH so that the traditional
  and bounding capabilities are set.
- Only remove CAP_DAC_OVERRIDE from the permitted set after opening the write
  file if neither -G nor -C flag was provided.  We always drop it from the
  effective set immediately after opening the write file.  During file rotation,
  we reacquire it immediately before and drop it immediately after opening the
  file.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -750,16 +750,6 @@ droproot(const char *username, const char *chroot_dir)
 		else {
 			fprintf(stderr, "dropped privs to %s\n", username);
 		}
-		/* We don't need CAP_SETUID and CAP_SETGID */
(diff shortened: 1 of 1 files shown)


fxlb:
Thank you for the patch and the detailed explanations !

PR #432, opened 2015-02-18
Fix -C with Capsicum

wxsBSD:
As discussed at http://thread.gmane.org/gmane.network.tcpdump.devel/7056/focus=7057, this fixes a bug when using -C with Capsicum due to a missing capability causing pcap_dump_ftell() to return -1.


infrastation:
(replaced by #433)

PR #433, opened 2015-02-18
Add CAP_FCNTL and use cap_fcntls_limit().

wxsBSD:
Without these changes pcap_dump_ftell() will return -1 and set errno to
ENOTCAPABLE.

This allows you to do:

tcpdump -C 1 -W 5 -w foo.pcap

Without these changes it will never rotate to foo.pcap1 and continue writing
to foo.pcap0 forever.

Discussed at: http://unix.derkeiler.com/Mailing-Lists/FreeBSD/current/2014-09/msg00142.html

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -1814,10 +1814,14 @@ main(int argc, char **argv)
 		if (p == NULL)
 			error("%s", pcap_geterr(pd));
 #ifdef HAVE_CAPSICUM
-		cap_rights_init(&rights, CAP_SEEK, CAP_WRITE);
+		cap_rights_init(&rights, CAP_SEEK, CAP_WRITE, CAP_FCNTL);
 		if (cap_rights_limit(fileno(pcap_dump_file(p)), &rights) < 0 &&
 		    errno != ENOSYS) {
 			error("unable to limit dump descriptor");
+		}
+		if (cap_fcntls_limit(fileno(pcap_dump_file(p)), CAP_FCNTL_GETFL) < 0 &&
+		    errno != ENOSYS) {
+			error("unable to limit dump descriptor fcntls");
 		}
 #endif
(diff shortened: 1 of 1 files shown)


guyharris:
Sigh.

That thread is a bit, err, umm, *interesting*.

In http://unix.derkeiler.com/Mailing-Lists/FreeBSD/current/2014-09/msg00167.html, Patrick Kelsey speaks of tcpdump's use of `fdopen()` as being an issue, as per the message from Andrey Chernov to which he was responding.

*The only reason why tcpdump uses `fdopen()` is that, in the Capsicum case, it uses `openat()` to open a file*; i.e., Capsicum's contribution to this problem is twofold!

In addition, the only reason why FreeBSD's standard I/O library uses `fcntl()` at all is to determine whether the file in question is append-only (which, BTW, it never would be in tcpdump).

If I ran the zoo, I might, in some fashion, allow any program that has Capsicum rights to read and write from a descriptor and seek on a descriptor to, at minimum, inquire whether it's open for appending.  (Dear AT&T: perhaps you should have provided separate APIs for setting and getting the `FD_` flags and getting and setting (some of) the `O_` flags, as well as for doing file locking, rather than adding a new `ioctl()`-style kitchen sink call.  But I digress....)

guyharris:
I've redone it to have a common routine to do all the Capsicum junk; that fixes one place in the code where the fcntl rights *weren't* being set.

I also included a long wailing comment, the executive summary of which is "if you pour Capsicum on your OS, it'll seep throughout the system, so maybe the standard I/O library routines need to know about it and either 1) offer an `fopenat()` call or 2) offer an `fdopen_cap()` call that ensures that *all* rights necessary to make the descriptor actually *usable* by the standard I/O library routines are set on the descriptor."

PR #440, opened 2015-03-05
Initial support for the REdis Serialization Protocol known as RESP.

adarqui:
This commit adds support for RESP as defined in: http://redis.io/topics/protocol.
It also supports inline commands and pipelining. Due to the popularity of RESP,
numerous services are emerging that use this protocol. You may decode RESP packets
on arbitrary ports using the "-T redis" option.

Example captures can be found in tests/resp*.

A simple way to test this parser is to start redis-server and then run
redis-cli commands such as "redis-cli set key value".

Traditionally, redis-cli monitor is used to debug redis. Unfortunately,
the "monitor" command can cause significant load on a redis-server in
production. This parser may be used as a non-invasive alternative to
redis-cli monitor.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -914,6 +914,75 @@ tstamp_precision_to_string(int precision)
 }
 #endif
 
+#ifdef HAVE_CAPSICUM
+/*
+ * Ensure that, on a dump file's descriptor, we have all the rights
+ * necessary to make the standard I/O library work with an fdopen()ed
+ * FILE * from that descriptor.
+ *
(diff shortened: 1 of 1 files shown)


adarqui:
@mcr , @infrastation , @fxlb,

Hi everyone. When I force pushed my "single commit", it closed out my previous PR. Here is the new pull request with one single commit as requested.

Thanks.

infrastation:
Thank you for the update. I cannot do a thorough proofreading right now, hopefully later about next week if it helps.

PR #449, opened 2015-04-08
Add support for TCP Fast Open official IANA option number

Longinus00:
https://tools.ietf.org/html/rfc7413

TCP Fast Open has left draft status and now has an IANA designated option number. This patch adds support for that.


infrastation:
`netdissect_options` usually comes first.

Longinus00:
I swapped the ordering of the arguments and also changed datalen from int to u_int.

infrastation:
Thank you!

PR #450, opened 2015-04-08
Remove an unnecessicary bit shifting loop

brooksdavis:
The inner loop serves no purpose and depends on signed integer overflow behavior which is undefined.  Our compiler based on clang-3.6 miscompiles the shift operation resulting in memory corruption if multiple flags are set.  I noticed the bug because our hardware enforced memory safety code trapped rather than overwriting memory.

### DIFF
--- a/util.c
+++ b/util.c
@@ -358,26 +358,14 @@ bittok2str_internal(register const struct tok *lp, register const char *fmt,
 {
         static char buf[256]; /* our stringbuffer */
         int buflen=0;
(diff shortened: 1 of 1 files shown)


brooksdavis:
Hmm, those test failures are quite odd.  I'll track them down and resubmit.

brooksdavis:
Updated commit to allow 0 as a valid set of flags.  Now passes all tests that master passes.

PR #451, opened 2015-04-08
Remove an unnecessicary bit shifting loop

brooksdavis:
The inner loop serves no purpose and depends on signed integer overflow behavior which is undefined. Our compiler based on clang-3.6 miscompiles the shift operation resulting in memory corruption if multiple flags are set. I noticed the bug because our hardware enforced memory safety code trapped rather than overwriting memory.

This version passes all tests that master passes.

### DIFF
--- a/util.c
+++ b/util.c
@@ -358,26 +358,14 @@ bittok2str_internal(register const struct tok *lp, register const char *fmt,
 {
         static char buf[256]; /* our stringbuffer */
         int buflen=0;
-        register int rotbit; /* this is the bit we rotate through all bitpositions */
-        register int tokval;
         const char * sepstr = "";
 
 	while (lp != NULL && lp->s != NULL) {
-            tokval=lp->v;   /* load our first value */
-            rotbit=1;
-            while (rotbit != 0) {
-                /*
-                 * lets AND the rotating bit with our token value
-                 * and see if we have got a match
-                 */
-		if (tokval == (v&rotbit)) {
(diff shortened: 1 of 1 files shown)


guyharris:
The bittok2str routines are, for better or worse, currently being used for bitfields that *aren't* flags, e.g. the "recursion count" field in GRE (or "recursion control", as it's called in [RFC 1701](http://tools.ietf.org/html/rfc1701)).  They don't *work as one might expect* for that, as the routine assumes that each entry refers to something that's either true or false.

Should we change the routines to take a table in which each entry contains an optional format string and, if the entry has a format string, the bitfield's value is extracted (including shifting it right appropriately) and printed using the format string, so that the recursion count would show up as `recursion count` *n*?  Or should we fix the tables so that they only contain single-bit values?

brooksdavis:
I don't have a strong opinion which is the right approach.  It does seem that the current bittok2str is quite hard to understand so separating flags printing from other functionality seems like a good idea.

FWIW. The simplest fix to the miscompliation and corruption I saw is to use an unsigned int rather than an int for rotbit since the overflow is defined in that case.

guyharris:
Change 30b028f9da2948a88207da7ae4c1d34e7c0a0f09 in the trunk and 8f0170452f5ebc2aae97544f296e8984e942a5cb in the 4.7 branch use unsigned ints there and elsewhere.

brooksdavis:
miscompiliation resolved.

PR #452, opened 2015-04-15
Backend separation

brooksdavis:
This change reduces the set of public function interfaces to the internal libnetdissect library to the five defined in print.h.  This greatly simplifies the task of putting dissection in a strong sandbox.

This change has no functional effect and passes the tests that passed before the change.  Accepting this change would greatly simplify the maintenance required next time I update CheriBSD (our research fork of FreeBSD) to a version that includes a new version of tcpdump.


mcr:
I'm excited as I think I hear that you will next submit something to split up the code...

brooksdavis:
I've gone significantly further in our CheriBSD project https://github.com/CTSRD-CHERI/cheribsd/, but those require an experimental CPU so it probably doesn't make sense to upstream those changes.  I might well take a look at a simple capsicum sandbox split between the frontend and backend at some point, but I'm not sure when I'd do that.  The main thing is the way and the moment is sharing of gndo.  If we could do away with the global that would simplify things nicely.

PR #453, opened 2015-04-22
Add support for the Marvell Extended Distributed Switch Architecture header

lunn:
This header can be used with Marvell switches to direct packets in/out
of a specific port in a tree of interconnected switches. The header
uses its own Ethertype of 0xdada.

By default, only brief output is printed, showing the switch device,
port, and vlan the packet is to/from. However if -e is given, to print
the link-level header, all fields are printed.


print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+
+#include <string.h>
+
+#include "interface.h"
+#include "addrtoname.h"
+#include "ether.h"
+#include "ethertype.h"
+#include "extract.h"			/* must come after interface.h */
+
+static const char tstr[] = "[|DSA]";
```
fxlb: s/DSA/EDSA/

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+
+/*
+ * Extended Distributed Switch Archiecture.
+ *
+ * A Marvell propriatary header used for passing packets to/from
+ * specific ports of a switch.
+ *
```
fxlb: URL of EDSA specification ?

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+	{ TAG_TO_CPU, "To_CPU" },
+	{ TAG_FROM_CPU, "From_CPU" },
+	{ TAG_FORWARD, "Forward" },
+	{ 0, NULL },
+};
+
+static const struct tok code_values[] = {
+	{ CODE_BDPU, "BDPU" },
+	{ CODE_IGMP_MLD, "IGMP/MLD" },
+	{ CODE_ARP_MIRROR, "APR Mirror" },
```
fxlb: s/APR Mirror/APR_Mirror/
(avoid spaces)

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+	{ 0, NULL },
+};
+
+static void edsa_print_full(netdissect_options *ndo,
+			    const struct edsa_pkthdr *edsa,
+			    u_int caplen)
+{
+	u_char tag = TAG(edsa);
+
+        ND_PRINT((ndo, "%s",
```
fxlb: s/<8 spaces>/<tab>/

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+void
+edsa_print(netdissect_options *ndo,
+	   const u_char *bp, u_int length, u_int caplen)
+{
+	register const struct ether_header *ep;
+	const struct edsa_pkthdr *edsa;
+	u_short extracted_ether_type;
```
fxlb: extracted_ether_type is unused

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+#define TRUNK(edsa)	((edsa->bytes[3] >> 2) & 0x01)
+#define CODE(edsa)	((edsa->bytes[3] & 0x06) |	\
+			 ((edsa->bytes[4] >> 4) & 0x01))
+#define CODE_BDPU	0
+#define CODE_IGMP_MLD	2
+#define CODE_ARP_MIRROR	4
+#define CFI(edsa)	(edsa->bytes[3] & 0x01)
+#define PRI(edsa)	(edsa->bytes[4] >> 5)
+#define VID(edsa)	(((u_short)(edsa->bytes[4] & 0xf) << 8 |						edsa->bytes[5]))
+	u_short ether_type;
```
fxlb: Line too long

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+};
```
fxlb: I suggest to put the `#defines` out of the struct, for a better readability.

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * A Marvell propriatary header used for passing packets to/from
+ * specific ports of a switch.
+ *
+ */
+struct  edsa_pkthdr {
+	u_char bytes[6];
+
+	/* Bytes 0 and 1 are reserved and should contain 0 */
+#define TAG(edsa) 	(edsa->bytes[2] >> 6)
```
fxlb: no space before tab

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+#define TAG_TO_CPU	0
+#define TAG_FROM_CPU	1
+#define TAG_FORWARD	3
+#define SRC_TAG(edsa) 	((edsa->bytes[2] >> 5) & 0x01)
```
fxlb: no space before tab

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+ * Lawrence Berkeley Laboratory and its contributors.'' Neither the name of
+ * the University nor the names of its contributors may be used to endorse
+ * or promote products derived from this software without specific prior
+ * written permission.
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND WITHOUT ANY EXPRESS OR IMPLIED
+ * WARRANTIES, INCLUDING, WITHOUT LIMITATION, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#define NETDISSECT_REWORKED
```
fxlb: The NETDISSECT_REWORKED macro is no more used

fxlb:
Thanks for the patch.
Where are the specifications of EDSA?

fxlb:
Is this is specific to Marvell, a better name could be `mdsa` or `medsa`?

fxlb:
The two files tests/edsa.pcap and tests/edsa-e.pcap are identical.
Only one is useful.

print-edsa.c
```diff
@@ -0,0 +1,183 @@
...
+		edsa_print_full(ndo, edsa, caplen);
+
+	bp += 8;
+	length -= 8;
+	caplen -= 8;
+
+	ether_type = EXTRACT_16BITS(&edsa->ether_type);
+	if (ether_type <= ETHERMTU) {
+		/* Try to print the LLC-layer header & higher layers */
+	  if (llc_print(ndo, bp, length, caplen, ESRC(ep), EDST(ep)) < 0)
```
fxlb: `s/<2 spaces>/<tab>/`

print-edsa.c
lunn: There is no public specification. If you have a Marvell NDA, it is specified in the data sheets for most Marvell switches.

This protocol is a bit different to most, in that it is only spoken between the CPU and the switch. See  https://lwn.net/Articles/302333/. You should not see these packets anywhere else on the network. Hence it can be kept "secret". But if you are developing software on the CPU using this header, which is what i'm doing, it is nice to have tcpdump decode them.

print-edsa.c
lunn: You might of figured this out, but i copied print-arp.c as a starting point. It mixes its structure definition and #defines like this. But i can move them out.

print-edsa.c
lunn: Left over from print-arp.c. Will remove.

lunn:
> Is this is specific to Marvell, a better name could be mdsa or medsa?

Yes, this is specific to Marvell. The Broadcom switches have a similar protocol, again proprietary.

The Marvell data sheets call it EDSA, and there is also an old, simpler DSA protocol. I'm happy to rename to medsa, but then it no longer matches the data sheet naming. Let me know what you want.

print-edsa.c
fxlb: Yes, please.

fxlb:
Yes, please, rename it to medsa.

lunn:
Just rename the file to print-medsa.c, or should i do a full s/edsa/medsa?

fxlb:
Yes, file name, s/edsa/medsa/ and s/EDSA/MEDSA/

lunn:
Here is v2.

I think i covered all your comments.

fxlb:
Merged, Thank you!

fxlb:
@lunn it seems that this document (pages 139-143) give some specifications:
http://www.marvell.com/embedded-processors/kirkwood/assets/FS_88F6180_9x_6281_OpenSource.pdf
Do you confirm?

lunn:
This is documenting DSA, not EDSA. They are very similar, but EDSA uses an EtherType so you know it is there, where as DSA does not have anything identifying if it is there or not, you just have to know.

The fields are however very similar, so the DSA specification is useful background, but you should not 100% trust it.

fxlb:
Not sure I understand. doc states
 "The GbE port supports both extended (8 bytes) and non-extended (4 bytes) DSA tag formats."
and tables 45 and 46 gives fieds for 2 words of 32 bits, thus 8 bytes.

lunn:
Maybe we have a terminology problem. But look at figure 28. We have Destination Address, Source Address, and then the 8 byte tag. This is what i can DSA.

The tcpdump patch i submitted has Destination Address, Source Address, Length/Type of 0xdada, two unused bytes, than four bytes which correspond to the right of figure 28. This is what i call EDSA. The linux kernel also uses this terminology.  The second word in table 44 is an extension, not shown in figure 28, and i've never seen it on the wire.

fxlb:
Thanks for your explanations.

PR #455, opened 2015-04-24
Remove global netdissect_object and remove non-dissection related fields

brooksdavis:
This set of changes further improves the split between the frontend and backend by eliminating a shared global netdissect_object and removing several members from the struct that deal with frontend packet processing details.


guyharris:
Any reason not to simply allocate ndo on the stack?

brooksdavis:
Putting it on the stack and zeroing it would be fine.  I'd done that initially and switched to calloc for no particularly good reason.

PR #463, opened 2015-06-07
3 fixes : ambiguous else, int overflow, trailing whitespaces.

stevekay:
Fixed the ambiguous else issue, fixed int overflow issue, fixed trailing whitespace.


guyharris:
Umm, this isn't a fix for some existing issues, it's a change adding a new option.  Should this be a further commit to another pull request adding that new option, rather than a pull request of its own?

stevekay:
cancelled.

Issue #445, opened 2015-03-25
[RFC] More user-friendly verbose output for DNS flags


msekletar:
Hi everyone,

I've received complaint from user in bugzilla that it is difficult to distinguish DNS response which has f.e. AD bit set. Right know tcpdump will produce following output for DNS response having AD bit set,

``
09:35:59.377000 IP localhost.domain > localhost.43144: 20972$ 3/6/13 SSHFP, RRSIG, RRSIG (3012)
``

Note $ character appearing after the transaction ID. Basically question is if tcpdump should provide alternative output to this, i.e. ``20972 [AD]`` if verbose is set?

I think it would be reasonable to do that because dissector output is not documented anywhere afaik and you have to look to the source code of the dissector to find out what $ character actually means.

https://bugzilla.redhat.com/show_bug.cgi?id=1164048

infrastation:
Man page maybe?

infrastation:
Allright, the original report explains that tcpdump only displays 2 of the 3 flags, and for DNS responses only and requests that tcpdump displays all 3 flags and for both directions and uses the existing AD/DO/CD namespace instead of its own.

msekletar:
Should I prepare patch adding support for 3rd flag and introduce new ouput then?

infrastation:
That is a good idea, this at least can put the intended changes into the discussion scope.

infrastation:
Update: tcpdump actually prints the EDNS DO bit if given `-vv` argument.

PR #467, opened 2015-06-18
For DNS messages, represent the "DNSSEC OK" bit as "DO" instead of "OK"

wessels:
"DO" is a better description of the DNSSEC OK bit in the EDNS0 OPT record.  See RFC 3225 section 3.  There may be other bits allocated in the future that call themselves "Something OK".


infrastation:
This change is one of those discussed in bug #445. As far as the specification goes, "DO" is the right name for this specific bit. Thank you!

PR #470, opened 2015-07-22
Add Value 13 "VLAN" to Tunnel-Type RADIUS attribute

qnet-herwin:
I haven't included any unit tests, because the trivial change this is. It has been tested locally and it has been seen working there. It doesn't break any existing unit tests either.

Fun fact: this value is used to assign a VLAN to a port or WLAN user, I wouldn't be surprised if it's the most common used value for Tunnel-Type.

### DIFF
--- a/print-radius.c
+++ b/print-radius.c
@@ -37,6 +37,10 @@
  * RFC 2869:
  *      "RADIUS Extensions"
  *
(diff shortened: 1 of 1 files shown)


infrastation:
Thank you!

PR #471, opened 2015-07-22
Translate UDP/1700 as RADIUS

qnet-herwin:
Used for Cisco CoA/Disconnect, as shown on https://supportforums.cisco.com/discussion/11719766/ise-112-change-authorization-avaya-switches-5520. The RFC says port 3799, but Cisco does this otherwise, which means this is not an officially reserved port for RADIUS.

Before merging, I'd like to know if this kind of non-official ports would be allowed in the tcpdump source. If that's okay, it would still need unit tests (but I'd rather wait with them until I'm sure that this code is okay)


infrastation:
Would `-T radius` solve this?

qnet-herwin:
Only partially. Currently I'm testing with disconnecting/reauthenticating clients from an access point, which involves two protocols: RADIUS and SNMP. Using `-T radius` tries to parse the SNMP data als RADIUS. I can work around that by having two tcpdump instances run at the same time in different consoles, but it's far from perfect.

In general, I think the `-T` option would be more useful if it also required a filter. That way one could say things like `-T radius 'udp port 1700'`. Another option would be to not force the protocol to the given option, but using it as a fallback: try all protocol dissectors, if we didn't find a dissector specific for this protocol/port, fallback to the option provided by `-T` (actually, using a new and different argument for this behaviour would be a better idea, but you get the point). But I guess I'm drifting away from the case here ;)

infrastation:
It looks like Cisco has failed to comply with the specification (RFC 3576/5176) in their default configuration and their documentation states that in some products this port number cannot be changed. Could be better of course though recording 1700 as another exemption looks workable.

The other suggestions you have made look reasonable and it would be interesting to try a working prototype should you have one.

qnet-herwin:
> The other suggestions you have made look reasonable and it would be interesting to try a working prototype should you have one.

Nope, don't have anything other than the idea. Should I create a separate feature request for it?

infrastation:
Could you rebase on the current master branch and add the test case in case you have one?

qnet-herwin:
@infrastation The branch has been rebased, and I've added a unit test. It is a pretty small test, just one packet with a single RADIUS attribute in it, but it looked good enough as test to me.

infrastation:
Thank you. For the other features you have discussed could you open a new feature request?

PR #474, opened 2015-08-05
Implement basic CoAP support

vtrushkin:
CoAP is a simple binary protocol, but it has non-trivial options packaging. It's not easy to parse binary protocol representation in mind, so, it would be great to have CoAP support in tcpdump. This implementation supports:

* CoAP Messages Types
* Request/Response Messages Codes
* Options and data types listed in RFC 7252

The implementation doesn't interpret payload. In moot points I rely on this https://github.com/obgm/libcoap implementation. Test files were built based on interaction between example coap-client and coap-server provided by that library.


infrastation:
I have briefly reviewed and will need at least one more round to complete proof-reading. Meanwhile could you rebase on the current master branch and catch up with the changes made in commits 3fb4560, 99c91c3 and c1c3c77? Thank you.

vtrushkin:
Done. I've also added a few content types which mentioned here https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/26 Though new content types haven't been registered yet (as I know), but they are already used in existing implementations.

infrastation:
Thank you, the new code compiles now and passes the check. However, `touch .devel; ./configure; make clean all` produces lots of warnings for `print-coap.c`, could you review?

vtrushkin:
Sure.

2015-11-16 13:59 GMT+03:00 Denis Ovsienko <notifications@github.com>:

> Thank you, the new code compiles now and passes the check. However, touch
> .devel; ./configure; make clean all produces lots of warnings for
> print-coap.c, could you review?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/the-tcpdump-group/tcpdump/pull/474#issuecomment-156992854>
> .
>



-- 
Best,
Vyacheslav

vtrushkin:
oprint_XXX functions are used to print value of an option. They all have
common interface to be called through a pointer oprint_func_t. But in some
of them there is no need to use all arguments. I've applied common WA
(void) on that arguments and fixed const casting issues which were
graciously mentioned by compiler :)

2015-11-16 14:17 GMT+03:00 Vyacheslav NT <vyacheslav.trushkin@gmail.com>:

> Sure.
>
> 2015-11-16 13:59 GMT+03:00 Denis Ovsienko <notifications@github.com>:
>
>> Thank you, the new code compiles now and passes the check. However, touch
>> .devel; ./configure; make clean all produces lots of warnings for
>> print-coap.c, could you review?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/the-tcpdump-group/tcpdump/pull/474#issuecomment-156992854>
>> .
>>
>
>
>
> --
> Best,
> Vyacheslav
>



-- 
Best,
Vyacheslav

infrastation:
I have reviewed this again and will post the comments soon.

infrastation:
Those are the issues I see at the moment:
1. In `coap_print()` TKL values 9~15 must be interpreted as a message format error, TKL value 0 should be printed without the token or with an empty token (right now as "0x00" -- same as would be with TKL=1).
1. In `coap_print_options()` there is no check if `delta` is going to overflow 16-bit `option`, there are no `ND_TCHECK` checks for `option` and `length`.
1. In `oprint_opaque()` there is the same issue with 0 length being printed as a 0x00 byte.
1. In `oprint_string()` the string should be printed with `fn_printn()`, which will handle non-printable bytes (and detect end of buffer, which ideally should be accounted for).
1. `OPT_STRING_MAXLEN` is defined less than `OPT_PROXY_URL_MAXLEN`, is this on purpose?

Would you like to take care of this?

vtrushkin:
Some points make me confused after 5 months. I'll check them soon. Thank you.

infrastation:
No problem, take your time.

dogonthesun:
> In `coap_print()` TKL values 9~15 must be interpreted as a message format error.

Actually a token may have length >9 octets. I use `COAP_TKL` macro to get 4 bits following the type.

> TKL value 0 should be printed without the token or with an empty token.
> In `oprint_opaque()` there is the same issue with 0 length being printed as a 0x00 byte.

Fixed. Thanks.

> there are no `ND_TCHECK `checks for option and length

Fixed.

> `OPT_STRING_MAXLEN` is defined less than `OPT_PROXY_URL_MAXLEN`

Options have following formats https://tools.ietf.org/html/rfc7252#section-3.2
But each option has own maximum value length https://tools.ietf.org/html/rfc7252#section-5.10
Lengths of integer options are bound to wide used types (except of the Observe option https://tools.ietf.org/html/rfc7641#section-2 ). And many string options have the same maximum length. I didn't want to introduce new custom data structures to store limits for each option. Length checking is performed in `get_oprint_func()` with two exceptions for Proxy URI and Observe options.

infrastation:
I did not have (and still don't have) the time to complete this review and the original author is now gone but a copy of his work has been saved in the "vtrushkin-coap-support" branch of my repository infrastation/tcpdump. Closing the request for the time being.

PR #500, opened 2016-02-07
Don't drop CAP_SYS_CHROOT before chrooting.

atsampson:
If tcpdump is configured with both libcap-ng and --with-chroot, then it
needs to keep CAP_SYS_CHROOT until droproot is complete.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -577,12 +577,13 @@ droproot(const char *username, const char *chroot_dir)
 		exit(1);
 	}
 #ifdef HAVE_LIBCAP_NG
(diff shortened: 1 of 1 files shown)


fxlb:
Thanks.
`capng_updatev` and `capng_update` return 0 on success and -1 on failure.
it seems better to test the return values.

guyharris:
> `capng_updatev` and `capng_update` return 0 on success and -1 on failure.
It seems better to test the return values.

Yes, but that problem existed before this change, so we should fix all those calls.

Issue #503, opened 2016-02-19
Exiting after file limit reached doesn't print statistics


mpitt:
I'm using the combination of `-G <sec> -W 1` to terminate the capture after some time. When the capture terminates this way, tcpdump doesn't print the statistics on packets (captured, received by filter, dropped by kernel) as usual.

Found in version `4.5.1`, confirmed in version `4.7.4`.

stevekay:
Pull request submitted.
https://github.com/the-tcpdump-group/tcpdump/pull/506

guyharris:
Fixed in bb6a08174a1cb1e6281f6bcf7126a7a94e211a47.

PR #506, opened 2016-03-04
Fixes issue where statistics not reported when -G and -W options used

stevekay:
Added missing call to `info()`.

https://github.com/the-tcpdump-group/tcpdump/issues/503


guyharris:
Version with the conflicts resolved committed as bb6a08174a1cb1e6281f6bcf7126a7a94e211a47.

Issue #495, opened 2015-12-16
Feature request: conditional run dissector on traffic


qnet-herwin:
It has been mentioned briefly in #471: with the option `-T radius` it is possible to dissect traffic over non-RADIUS ports as RADIUS, but this means all traffic will be dissected as RADIUS.

A short use case as an example: FreeRADIUS has the possibility to use a RESTful server for authorization. My RESTful daemon will send a CoA/Disconnect if this user still has another session, and my NAS requires that I send it to port 1700 instead of the standard port. I want to see all the traffic that is relevant, which means UDP port 1812 for RADIUS authentication, TCP port 80 for the RESTful traffic and UDP port 1700 for RADIUS CoA/Disconnect.

Until yesterday, UDP port 1700 was not marked as RADIUS, which means that would not be dissected. If I use `-T radius`, all traffic would be dissected as RADIUS, so the HTTP traffic would be mangled.

It would be nice if there was a possibility to conditionally mark traffic as a certain protocol, like saying `-T "udp port 1700 = radius, tcp port 4080 = http"`

mcr:
> It has been mentioned briefly in #471: with the option -T radius it is
    > possible to dissect traffic over non-RADIUS ports as RADIUS, but this
    > means all traffic will be dissected as RADIUS

    > A short use case as an example: FreeRADIUS has the possibility to use a
    > RESTful server for authorization My RESTful daemon will send a
    > CoA/Disconnect if this user still has another session, and my NAS
    > requires that I send it to port 1700 instead of the standard port I
    > want to see all the traffic that is relevant, which means UDP port 1812
    > for RADIUS authentication, TCP port 80 for the RESTful traffic and UDP
    > port 1700 for RADIUS CoA/Disconnect

    > Until yesterday, UDP port 1700 was not marked as RADIUS, which means
    > that would not be dissected If I use -T radius, all traffic would be
    > dissected as RADIUS, so the HTTP traffic would be mangled

    > It would be nice if there was a possibility to conditionally mark
    > traffic as a certain protocol, like saying -T "udp port 1700 = radius,
    > tcp port 4080 = http"

Being able to dynamically map ports -> protocols in TCPDUMP would indeed be a
nice thing to have.    I think that squishing it all into -T is too hard.
I'd rather have a file that describes the mapping, and just bake a file like
that in as the default.  As for a syntax...
I was thinking that maybe we could use pcap filters as the matchers.

--
[...]

PR #510, opened 2016-03-18
GUE dissection support

Wilm0r:
Hello,

Here's a patch adding support for GUE (Generic UDP Encapsulation) to tcpdump. The spec can be found at: https://tools.ietf.org/html/draft-ietf-nvo3-gue-02 (header format in 2.2).

This is fairly basic but will show dissection of inner IPv4/IPv6 packets assuming GUE is indeed configured on the officially allocated port 6080. Sadly this is not going to be a firm standard (for example one may want to use a privileged <1024 port instead for security reasons), so I wonder whether, for example, using getservbyname() instead of hardcoded port numbers would be acceptable? There's no prior art for this in your code yet from what I can tell which is why I haven't done this yet.

Kind regards,

Wilmer van der Gaast.


Wilm0r:
I've merged recent changes, this PR should apply cleanly against master now.

print-gue.c
```diff
@@ -0,0 +1,164 @@
...
+		return;
+	}
+	vers = GUE_VERS(EXTRACT_16BITS(bp));
+	ND_PRINT((ndo, "GUEv%u", vers));
+
+	switch(vers) {
+	case 0:
+		gue_print_0(ndo, bp, len);
+		break;
+	default:
```
infrastation: Would you like to add support for version 0x1?

print-gue.c
```diff
@@ -0,0 +1,164 @@
...
+		prot = 0;
+	} else {
+		prot = GUE_PROTO(flags);
+	}
+	
+	/* Length of additional headers */
+	hlen = GUE_HLEN(flags);
+	hlen *= 4;
+
+	len -= 2;
```
infrastation: Is it for sure len > 2 before this point?

print-gue.c
```diff
@@ -0,0 +1,164 @@
...
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*
+ * netdissect printer for GUE - Generic UDP Encapsulation
+ * Currently described in:
+ * https://tools.ietf.org/html/draft-ietf-nvo3-gue-02#section-2.2
```
infrastation: Revision 03 of the Internet-Draft has been published.

print-udp.c
```diff
@@ -646,6 +646,8 @@ udp_print(netdissect_options *ndo, register const u_char *bp, u_int length,
 			geneve_print(ndo, (const u_char *)(up + 1), length);
 		else if (IS_SRC_OR_DST_PORT(LISP_CONTROL_PORT))
 			lisp_print(ndo, (const u_char *)(up + 1), length);
+		else if (dport == GUE_PORT)
```
infrastation: It seems from the spec that the reponse from the middlebox to the faraway host would have UDP ports the other way round (from GUE_PORT to any other port), should it be accounted here as well?

infrastation:
Regarding the port numbers, `getservbyname()` use would not be a good idea because its return value generally may be (and practice will be) volatile and inconsistent across different hosts. tcpdump may be run on a host different from the one that did tunnel termination and the one that did the packet capture.

Section 5.6.2 of the Internet-Draft discusses a use of a magic number. That may be a more reliable means of telling the packets apart but the spec would need to have it as a normative statement rather than as a potential future approach. Maybe you would like to discuss this with the document authors as the root of the problem seems to be there. The magic number support in tcpdump should not be difficult per se.

print-udp.c
Wilm0r: Hrm, I'll have to look more closely for that, I forgot about that case. Could that be metadata/OOB-only with no actual IP payload inside?

print-gue.c
Wilm0r: Happy to add another check for cleanliness if it helps. I don't like writing "secure only by pure luck" code :-)

print-gue.c
Wilm0r: I'll need some dumps to do so, I don't have any so far which makes writing (and testing) that a little bit hard. Will update this PR once that (and the other changes) are done.

Wilm0r:
Regarding magic numbers, I'm told that's most likely going to get removed from the next draft version, so better not to add support for it.

So what I'll do is add support for version 1 and squash the commits back into one.

print-udp.c
infrastation: I am not sure which kind of packet exactly but generally packets in tunnels tend to go both ways. Unless, of course, in this protocol the return traffic uses a completely unrelated source and destination ports. I did not study the spec so well.

Wilm0r:
Sorry for the long delay on this :-(

So magic numbers are going to get removed, I think it should be okay to not support those then?

And for GUEv1, I've tried to get example dumps but no response so far. One may be able to generate them using a very recent Linux kernel, but I'd rather not do this myself, not being all that familiar with GUEv1 in particular..

infrastation:
Thank you for the update. It would make it easier to review further with the new code rebased on the current master branch.

Wilm0r:
But of course, I was planning to do a merge but wanted to ask first how strongly you feel about the GUEv1 support.

Merged now. Not all tests pass, but the failing ones are not GUE and are failing in the parent branch as well. :-(

I'd squash this branch but I think Github can do that for you at merge time now?

fxlb:
Some warnings here: https://travis-ci.org/the-tcpdump-group/tcpdump/jobs/165389888
```
./print-gue.c: In function ‘gue_print_0’:
./print-gue.c:119:2: warning: C++ style comments are not allowed in ISO C90 [enabled by default]
./print-gue.c:119:2: warning: (this will be reported only once per input file) [enabled by default]
./print-gue.c:93:8: warning: unused variable ‘control’ [-Wunused-variable]
```

Wilm0r:
Oops, yeah those didn't appear during my local build. :-/ Fixed and the current build at https://travis-ci.org/the-tcpdump-group/tcpdump/jobs/165633760 looks clean now.

Wilm0r:
So is it okay to merge this one with just GUEv0, not v1? I've not been able to get dumps of GUEv1 traffic so far..

infrastation:
Could you rebase this on the current master branch?

infrastation:
Sure, take your time.

Wilm0r:
Oh, thanks for the hint. I was still trying to update the range checking to use your ND_T(TEST|CHECK) macros as well but couldn't fully figure them out. :-( (CHECK vs TEST and also what to do for the check on length>=sizeof(struct ip))

infrastation:
@Wilm0r, if you wish to resume this work, please update this pull request within 14 days. Otherwise it will be closed.

fxlb:
FYI, this is an expired Internet-Draft.
draft-ietf-nvo3-gue-02 -> draft-ietf-intarea-gue-09
(https://datatracker.ietf.org/doc/draft-ietf-intarea-gue/)

Wilm0r:
Correct, there seem to be two versions now. I'll see whether I can get this PR back into decent shape, and how to deal with the versions/support both of them at all. Don't know which one's more common.

I *think* the two protocols aren't really compatible nor are they detectable easily except maybe with heuristics. Would the usual tcpdump solution be to just name them separately, for example gue0 and gue1 ?

fxlb:
> there seem to be two versions now.

Do you have specifications about these versions/variants?

infrastation:
@Wilm0r, in case you don't have the motivation to follow the developments of this protocol design closely, it is OK not to implement it, and to get something else done instead. Often it is better not to have a particular protocol support at all than to have out of date code and the cost of its maintenance.

Wilm0r:
Sorry for losing track of this again.

I agree, I don't want to contribute code that is prone to rot. What I will try to do is figure out the status of GUEv0 vs v1 and see what is wise here. AFAIK (and I should add that at work I'm not actually that close to this project/development (anymore)) GUE is definitely used more widely including by some hardware so support *would* be useful, but I'll need to figure out which of the two versions.

Feel free to close this PR if you've lost patience, I can reopen/open a new one if needed. I'll obviously need to deal with a pretty intense merge anyway.

BTW, my main worry for this one is that GUE can be on ~any port#. (We use a privileged <1024 port instead of the IANA-reserved one ourselves.) When I wrote this patch, tcpdump didn't really support "use this port#" arguments for decoders (while Wireshark does have "reinterpret this stream as protocol X" for example), possibly this has changed?

infrastation:
The traditional workaround to decode a protocol of interest on a non-standard UDP port would be along the lines of `tcpdump -T gue udp port NNNN`, which would obviously exclude decoding all other traffic, but would be better than nothing.

infrastation:
If in 14 days it is still not clear how to progress this matter in a constructive way, it will be better to close this request.

infrastation:
For reference, a potential alternative way to apply dissectors to non-standard ports was discussed in #495.

infrastation:
It is time.

Issue #516, opened 2016-05-04
Possible crash caused by option 20 (SCPS-TP)


rbgarga:
We have a software on pfSense, called filterlog [1] and this software is basically made using part of tcpdump code. We received an email from Patrik Lindquist patrik.lundquist@gmail.com reporting a crash on filterlog and after a quick look at tcpdump code I believe this same problem may occour on it so I decided to come here and share all investigation he did. Following you can see his email content:

```
what seemed to be a rarely occurring odd bug turned out to be quite interesting.

Once I had caught the bug with a filterlog debug build (attached) I began analyzing the TCP options in the TCP SYN packet. Filterlog crashed on option 20 (0x14).

0x02 0x04 0x05 0xac = MSS = 1452
0x01                = NOP
0x03 0x03 0x02      = Window Scale = 2
0x01 0x01           = NOP NOP
0x14 0x02           = SCPS Capabilities   <--- !!!
0x00 0x00 0x00 0x00 = EOL

https://www.iana.org/assignments/tcp-parameters/tcp-parameters.xhtml

SCPS-TP is a space communications version of TCP and quite unexpected on the internet.

https://en.wikipedia.org/wiki/Space_Communications_Protocol_Specifications

– TCP for Transactions (RFC 1644, reference [13])—reduces the handshaking necessary to start a TCP connection and provides ‘reliable datagram’  operation to handle command-response traffic, for very long delay environments in which it is desirable to begin data transfer without waiting for a connection handshake;

http://public.ccsds.org/publications/archive/714x0b2.pdf

See section 3.2.3 on page 20 in the PDF.

SCPS Capabilities Option shall be sent in a i TCP SYN packet and shall be four fields of one byte each and begin with 0x14 0x04 (Kind=20, Length=4).

The offending packet has curiously omitted the two mandatory fields Capabilities Option Bit-Vector and Connection ID and have accordingly adjusted the option length to 0x02, which breaks the SCPS-TP standard but follows the TCP Options standard in general.

The packet will therefore pass TCP option parsers that don't specifically care about SCPS-TP.

So why does filterlog crash, or even care about SCPS-TP? See sysutils/filterlog/files/print-tcp.c.

1. TCPOPT_AUTH is incorrectly defined to option 20 while TCP-AO actually is option 29. This in itself isn't a problem since only the option kind is logged and the option length is correct.

#define TCPOPT_AUTH             20      /* Enhanced AUTH option */


[...]

PR #517, opened 2016-05-05
Correct TCP option Kind value for TCP Auth and add SCPS-TP.

pLuster:
Fixes first problem in issue #516 while the second one isn't broken in tcpdump.

https://www.iana.org/assignments/tcp-parameters/tcp-parameters.xhtml

See section 3.2.3 on page 20 in http://public.ccsds.org/publications/archive/714x0b2.pdf

Review carefully because I haven't been able to test it with TCP Auth and SCPS-TP packets.


infrastation:
Minor correction: the SCPS spec has [moved](https://public.ccsds.org/Pubs/714x0b2.pdf).

infrastation:
I have studied the spec and found that the proposed change implements the SCPS part of the issue discussed in #516 right. Thank you!

PR #518, opened 2016-05-08
Add autoconf option to enable a stripped down build.

biot:
The --enable-minimal option to configure reduces the list of dissectors to a smaller set, the minimum needed to bring up an embedded system.

This produces a smaller tcpdump binary, something that is very useful on small embedded systems, such as SOHO routers. Size reduction is 45-50%, depending on architecture. OpenWrt currently has a similar patch for an older version, but cannot upgrade to current tcpdump with it. Other embedded distributions like Buildroot and Yocto would greatly benefit from this patch as well.

The patch necessarily introduces a lot of #ifdefs to various dissectors. Those can be reduced by grouping together non-minimal functions in dissectors files. I did it this way in print.c for example. Please let me know if that's your preference, happy to respin.


infrastation:
This idea looks useful to me. One nit: it would be better to rename the macro to `ND_MINIMAL` or even to something like `ND_PROFILE`, in the latter case defaulting its value to "full" and allowing to set it to "minimal" (or "embedded"), such that in future it would be easier to add another profile that sits somewhere in between. Would you like to respin this change?

biot:
I've rebased the patch to master, and changed it to ND_MINIMAL. Binary size is about half with the minimal build.

mcr:
We intend to merge this as soon as some other more fragile security related changes have been merged.

infrastation:
I have to some extent rebased the proposed change in my working copy, it is now 230 commits ahead (was August 2018, now March 2019), but this pull request does not allow me to force-push into the biot:mini branch. I can make another rebase round later or I can hand the intermediate result over to whoever has time and interest to do it now.

PR #530, opened 2016-07-11
Geneve TLV Option Classes

jessegross:
The latest versions of the Geneve draft (https://tools.ietf.org/html/draft-ietf-nvo3-geneve-02) have additional assignments for option classes. This adds support for printing them by name instead of by hex value.

### DIFF
--- a/print-geneve.c
+++ b/print-geneve.c
@@ -183,7 +183,7 @@ geneve_print(netdissect_options *ndo, const u_char *bp, u_int len)
 
     if (len < opts_len) {
         ND_PRINT((ndo, " truncated-geneve - %u bytes missing",
-                  len - opts_len));
+                  opts_len - len));
         return;
     }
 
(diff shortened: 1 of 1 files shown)


fxlb:
Hello,
In https://travis-ci.org/the-tcpdump-group/tcpdump/jobs/143953102, there is a warning:
```
./print-geneve.c: In function ‘format_opt_class’:
./print-geneve.c:93:9: warning: comparison is always true due to limited range of data type [-Wtype-limits]
```
Could you fix it ?

jessegross:
Sorry for missing the warning, I just uploaded a new version that fixes it. In terms of the test, there actually is an existing test case that exercises this (geneve-v). It uses one of the older option class assignments but it's still handled by the same code and I guess testing one class is as good as any other.

infrastation:
You are right, I have nothing else to add here.

fxlb:
Merged, Thank you!

Issue #522, opened 2016-06-04
Numeric adapter names can't be used for capture


nbastin:
```
$ tcpdump -i 3703570267332                                                                                                                                                                                                                                                               
tcpdump: Invalid adapter index
```

tcpdump should try to open the name before it decides it isn't there:

```
31231: 3703570267332: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast master ovs-system state UP mode DEFAULT group default qlen 1000
    link/ether c2:22:16:68:ec:d4 brd ff:ff:ff:ff:ff:ff
```

Also, adapter indices aren't `ifindex` numbers (this is clear-ish from the man page, but not from the error message), they're some version of a number that tcpdump makes on its own, which you can only get by running `tcpdump -D`, which takes literally *minutes* on systems with lots of interfaces:

```
$ time sudo tcpdump -D
...
real    2m10.018s
user    0m0.048s
sys     0m4.680s
```

infrastation:
The reason for the `tcpdump -D` interface number not being the OS ifindex is that libpcap supports other capture channels besides the common network interfaces:
```
$ ./tcpdump -D
1.enp0s3 [Up, Running]
2.any (Pseudo-device that captures on all interfaces) [Up, Running]
3.lo [Up, Running, Loopback]
4.nflog (Linux netfilter log (NFLOG) interface)
5.nfqueue (Linux netfilter queue (NFQUEUE) interface)
6.usbmon1 (USB bus number 1)
```
In the list above only numbers 1 and 3 have a corresponding ifindex.

That said, when a network interface name is a number, the existing convention does not work. As the comment in `tcpdump.c` around line 930 goes:
```c
			/*
			 * If the argument is a number, treat it as
			 * an index into the list of adapters, as
			 * printed by "tcpdump -D".
			 *
			 * This should be OK on UNIX systems, as interfaces
			 * shouldn't have names that begin with digits.
			 * It can be useful on Windows, where more than
			 * one interface can have the same name.
			 */
```
In this case either the existing convention should be reviewed or the system in question should use a more traditional network interface naming pattern.

infrastation:
How many network interfaces does it take to slow `tcpdump -D` down?

guyharris:
The number is not, and has never been intended to be, an ifindex.  It is, and has always been intended to be, an ordinal number in the `tcpdump -D` list of interfaces.

[...]

PR #533, opened 2016-07-17
Issue #522: Solution: Attempt pcap_create() first

bijalthanawala:
This addresses issue:
**Numeric adapter names can't be used for capture**
(https://github.com/the-tcpdump-group/tcpdump/issues/522)

- Attempt pcap_create() first and only on failure numeric indexes are scanned
- Handle use-case: Do not do the above when interface  (-i) was not specified by the user
- Handle special case  (negative zero: -i -0)

Todo / Question:
Do we need to modify tcpdump man page to explain this behaviour ?

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -753,6 +753,71 @@ set_dumper_capsicum_rights(pcap_dumper_t *p)
 }
 #endif
 
+
+#ifdef HAVE_PCAP_FINDALLDEVS
+static char *
+search_numeric_adapter_name(char *device, char *ebuf, size_t size_ebuf)
+{
+	/*
+	 * If the argument is a number, treat it as
+	 * an index into the list of adapters, as
+	 * printed by "tcpdump -D".
+	 *
+	 * This should be OK on UNIX systems, as interfaces
+	 * shouldn't have names that begin with digits.
+	 * It can be useful on Windows, where more than
+	 * one interface can have the same name.
+	 */
+
+	register int i;
+	int devnum;
+	char *end;
+	pcap_if_t *dev, *devlist;
+
+	devnum = strtol(device, &end, 10);
+	if (device != end && *end == '\0') {
+		if (devnum <= 0) {
+			/* Return error for negative number input.
+			 * Additionally, handle ZERO because strtol
+			 * converts input of "-0" to a ZERO
+			 */
+			strncpy(ebuf,"Invalid adapter index", size_ebuf-1);
+			return NULL;
+		}
+
+		if (pcap_findalldevs(&devlist, ebuf) < 0) {
+			/* ebuf contains error message from pcap_findalldevs */
+			return NULL;
+		}
+
+		/*
+		 * Look for the devnum-th entry in the
+		 * list of devices (1-based).
+		 */
+		for (i = 0, dev = devlist;
+		    i < devnum-1 && dev != NULL;
+		    i++, dev = dev->next)
+			;
+
+		if (dev == NULL) {
+			strncpy(ebuf,"Invalid adapter index", size_ebuf-1);
+			return NULL;
+		}
+
+		free(device); /* device is a strdup'ed string at this stage. Free it first*/
+		device = strdup(dev->name);
+		pcap_freealldevs(devlist);
+	}
+	else {
+		/* Leave the original message in ebuf (as on entry) untouched */
+		return NULL;
+	}
+
+	return device;
+}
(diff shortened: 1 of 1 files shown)


nbastin:
This is sortof pedantic, but I think important.  Can we remove the comment that says:

```
+	 * This should be OK on UNIX systems, as interfaces
+	 * shouldn't have names that begin with digits.
```

This is an opinion, and has no basis in the POSIX specification or AT&T UNIX documentation.  I don't know why the original author of this comment thought that interfaces "shouldn't" have names that begin with digits, but this has been legal on many systems for decades.  Future maintainers of this code shouldn't be led to believe that this is an invalid configuration that they don't need to handle (which is possibly how we got here in the first place).

guyharris:
> but this has been legal on many systems for decades

"Shouldn't" can either be prescriptive or descriptive; in this case, it's descriptive, because, until Linux or whoever decided to start giving purely-numerical names to interfaces, no UN\*X systems actually did have purely-numerical names for interfaces - not 4.xBSD, not SunOS, not HP-UX, not AIX, not Tru64 UNIX, not any of the BSDs descended from 4.x, etc..

Given that, *now*, there are UN\*X systems that have all-numeric interface names, the comment can be removed, but it's not as if all-numeric interface names have been a thing for decades.

tcpdump.c
```diff
@@ -1371,8 +1405,29 @@ main(int argc, char **argv)
 #endif /* _WIN32 */
 #ifdef HAVE_PCAP_CREATE
 		pd = pcap_create(device, ebuf);
-		if (pd == NULL)
+		if (pd == NULL) {
```
guyharris: You are assuming here that `pcap_create()` will fail if an interface name is supplied that does not correspond to an interface on the system.  It is not guaranteed to do so on all platforms.  You have to wait for `pcap_activate()` to fail with `PCAP_ERROR_NO_SUCH_DEVICE` (other errors mean there *is* such a device but you can't open it).

This also doesn't handle the case of tcpdump with an older version of libpcap without `pcap_create()`.

nbastin:
> "Shouldn't" can either be prescriptive or descriptive; in this case, it's descriptive, because, until Linux or whoever decided to start giving purely-numerical names to interfaces, no UN*X systems actually did have purely-numerical names for interfaces - not 4.xBSD, not SunOS, not HP-UX, not AIX, not Tru64 UNIX, not any of the BSDs descended from 4.x, etc..

My comment very clearly said it was "legal" - and still is. Leading someone in the future to believe they can rely on such a restriction would create more problems.  Neither POSIX nor AT&T ever said you shouldn't do it, as X.25 PAD interfaces commonly did - and probably others before that - which makes "shouldn't" in the comment prescriptive.  Interfaces have long been named numerically, the fact that PARC didn't do it for ethernet doesn't make it less valid.  Certainly it's even (potentially) more common today than it was in the past (as I believe that under Linux this was in fact not legal for quite some time, while it was always legal in most other UNIX derivatives).

Common practice as known to a developer in their experience is not a spec, it's just one practice.

guyharris:
> which makes "shouldn't" in the comment prescriptive

What would make it prescriptive or descriptive would be the intent of the person who wrote it; the person who wrote it meant it descriptively, which makes it descriptive.

> Interfaces have long been named numerically

On which operating systems that supported libpcap and tcpdump were the names used in libpcap calls, and on the tcpdump command line, numerical?  If we look at [an older version of the FreeBSD `net/ifvar.h`](https://svnweb.freebsd.org/base/head/sys/net/if_var.h?revision=21259&view=markup), a `struct ifnet` has `if_name` and `if_unit` members; the convention was to have `if_name` be a short name for a particular model of network adapter and `if_unit` be a 0-based ordinal for the particular interface in the set of all interfaces of that model, with `if_name` not being purely numerical (or even ending with a digit), as that makes it hard to tell where the name ends and the unit number begins.  Many other UN\*Xes followed BSD in that convention; Linux and OS X used a similar convention, but with `if_name` not being different for different adapter types, and with the Linux loopback interface not having a unit number.

Interfaces may have been named numerically in *other* contexts, but, for tcpdump, the only context that matters is the "argument to the `-i` flag" context.

guyharris:
Fixed instead by 0bc8e0e01ef0b14ed8663f37ce224f35f5fb496b.

guyharris:
> Do we need to modify tcpdump man page to explain this behavior ?

Done in d4876278a77ed6576073cc6fd81d2ec43eb85733.

bijalthanawala:
Thanks guy. Reviewing it.

PR #534, opened 2016-07-28
HNCP support (RFCs 7787 and 7788)

cryhot:
Add support for the Home Network Configuration Protocol (HNCP) described in RFCs 7787 and 7788.


guyharris:
`"%016lx"` isn't guaranteed to correctly print a 64-bit quantity; use `"%016" PRIx64` instead.  Fixed in 174b3ec974cf30a02328193c5ef50d2ed56afa91.

PR #537, opened 2016-08-18
HNCP: fix incorrect time interval format

MisterDA:
In HNCP/DNCP a time interval is stored in milliseconds on a 32-bits value (see [RFC 7787#7.3.2](https://tools.ietf.org/html/rfc7787#section-7.3.2)). The function `format_interval` was incorrectly taking an `uint16_t` and made nonsense with the conversion.
The first commit fixes this bug and the second updates the test file.

I am the co-author of the HNCP dissector with @cryhot.

### DIFF
--- a/print-hncp.c
+++ b/print-hncp.c
@@ -163,14 +163,12 @@ format_256(const u_char *data)
 }
 
 static const char *
-format_interval(const uint16_t n)
+format_interval(const uint32_t n)
 {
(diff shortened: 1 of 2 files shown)


MisterDA:
Hi guys, any news on this ? It's been two weeks, and it's just a five-lines fix.

fxlb:
Merged. Thank you.

PR #544, opened 2016-10-26
New Early IANA Allocation

pierky:
https://www.ietf.org/rfcdiff?url2=draft-ietf-idr-large-community-05


infrastation:
The previously added test case still uses a .pcap file with the old allocation and now fails (try `make check` in the working directory).

pierky:
The .pcap file has been updated with a new dump containing the new code. Tests are fine now.

job:
thanks @pierky

fxlb:
Thanks, I will merge an updated patch
(see fxlb@e2adb16)
- with correct TCP checksum
-  only one commit
- an updated commit message

fxlb:
Merged.

Issue #545, opened 2016-10-26
Update Large BGP Communities Path Attribute value


job:
In commit https://github.com/the-tcpdump-group/tcpdump/issues/543 support for Large BGP Communities was added, but IANA assigned a new codepoint because the attribute value was squatted on.

The new value is `32` http://www.iana.org/assignments/bgp-parameters/bgp-parameters.xhtml#bgp-parameters-2 (instead of 30).

pierky:
It should be fixed with PR #544.

PR #546, opened 2016-10-26
Update Large BGP Communities path attribute value

job:
This fixes https://github.com/the-tcpdump-group/tcpdump/issues/545


pierky:
PR #544 updates the IANA code and (now) passes tests too.

fxlb:
Thanks, I will merge an updated patch
(see https://github.com/fxlb/tcpdump/commit/e2adb166ecaf59555e18787e86ac8c828efa0acc)
- with correct TCP checksum
- only one commit
- an updated commit message

PR #570, opened 2016-12-01
rotate when signal USR2 is received.

franky-m:
With this patch the signal USR2 triggers rotation of the output file.

I see two scenarios where this function might be useful:

1) clean shutdown: 
Do not leave an unrotated partial file when exiting due to CTRL-C or a signal
(e.g. # pkill -SIGUSR2 tcpdump; sleep 1; pkill -KILL tcpdump; rm $partial_file)

2) externalize decision on when to rotate: 
You no longer have to decide when to rotate before tcpdump is started. This way rotation intervall 
can also be adopted to changing traffic without restarting tcpdump and missing packets.

If 2. is not a missed feature I could also change the patch so that tcpdump simply does a clean shutdown
when the signal is sent. 

Any feedback is appreciated!

Franky

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -133,6 +133,7 @@ static int Bflag;			/* buffer size */
 static int Cflag;			/* rotate dump files after this many bytes */
 static int Cflag_count;			/* Keep track of which file number we're writing */
 static int Dflag;			/* list available devices and exit */
+static int rotate_now = 0;              /* user forced rotate after next packet */
 /*
  * This is exported because, in some versions of libpcap, if libpcap
  * is built with optimizer debugging code (which is *NOT* the default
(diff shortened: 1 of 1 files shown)


infrastation:
The proposed change looks OK to me, but needs to be rebased on the current master branch, and needs to disable itself when the OS does not have SIGUSR2. Other reviews are welcome.

infrastation:
As it turns out, SIGUSR2 is in POSIX, Linux, FreeBSD, NetBSD, OpenBSD and illumos, so that particular point is moot.

infrastation:
(Also in Solaris 9, 10 and 11; AIX 7.1 and 7.2.)

infrastation:
Rebased on the current master branch and cleaned up. Not tested. It remains to be established whether this change addresses a sound problem and whether the solution is correct.

infrastation:
Unfortunately, `SIGUSR2` was taken for flushing the packet buffer in commit 0e79762 in April 2018.

PR #573, opened 2016-12-08
Pull request for Manufacturer Usage Descriptions

elear:
These files contain updates for new dhcpv4 and v6 options.  They also correct what I believe is an error in print-bootp.c.  I'll update the pull request with the issue # shortly.


elear:
Will reissue pull request (TESTFILE got missed.

Issue #572, opened 2016-12-08
Error in print-bootp.c: TAG_PRINTER_NAME and TAG_MDHCP_SERVER


elear:
The values for these bootp options do not match the IANA assignments, which refer to the POSIX_TCODE and PCODE, respectively (see RFC 4833).  I've got a patch ready for this that I'll issue a pull request against shortly, along with other stuff (it contains a new option, which is how I noticed).

guyharris:
Fixed by pull request #574, right?

PR #574, opened 2016-12-08
Updated Pull Request for Manufacturer Usage Descriptions

elear:
This pull request adds support for draft-ietf-opsawg-mud or manufacturer usage descriptions.  These are DHCP options that devices use when they want ACLs employed by the network to provide some level of protection.  See the draft for more details.

This pull request also provides some support for RFC 4833 options.  There is also an issue opened related to timezones in print-bootp.c that this pull request addresses.


print-bootp.c
```diff
@@ -212,15 +212,17 @@ struct bootp {
 #define	TAG_CLIENT_GUID		((uint8_t)  97)
 #define	TAG_LDAP_URL		((uint8_t)  95)
 #define	TAG_6OVER4		((uint8_t)  96)
-#define	TAG_PRINTER_NAME	((uint8_t) 100)
```
elear: This patch addresses Issue #572

infrastation:
In `print-dhcp6.c` and `print-lldp.c` it would be wrong to print a string using "%s" because the contents of the memory may be generally arbitrary. It may contain control characters and is not guaranteed to contain a null character before `printf()` runs beyond the allocated memory. The established safe way of printing a string from a packet is to use one of the following functions: `fn_print()`, `fn_printn()`, `fn_printzp()`. Please check the comments at the declarations of those functions to see the subtle but important difference between them, so you can use the right one for each of the decoders concerned.

elear:
Hi and thanks for your review.  I'll recommit along the following lines if it sounds right to you:
                        ND_PRINT((ndo,"="));
                        fn_printn(ndo,tp,(int) optlen,NULL);
                        ND_PRINT((ndo,")"));

PR #575, opened 2016-12-10
delete judge

hhktony:
No need to judge


guyharris:
Fixed in 3976b4067c496bf310dbcff0b16908546bf8e3ff, with updates to a comment.

PR #585, opened 2017-02-01
A bunch of changes that we got accumulated in the FreeBSD tree.

glebius:
Fixes compilation at high warning levels.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -352,6 +352,7 @@ EXTRA_DIST = \
 	packetdat.awk \
 	pcap_dump_ftell.c \
 	print-pflog.c \
+	print-pfsync.c \
 	print-smb.c \
 	send-ack.awk \
 	smbutil.c \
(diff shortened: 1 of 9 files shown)


guyharris:
Please split this into separate pull requests for the `addrtostr.c` change (which we'll probably do differently - there's a *reason* why we made the loop index unsigned; see the changelog for `addrtostr.c`) and the pfsync enhancement.

glebius:
I have hard time with github. Initially I created pull request for addrtostr.c. Then I pushed commit for pfsync into my fork, and it automatically was added to the pull request.

guyharris:
You probably need separate branches for each change.  (I keep separate *source trees* for each change I'm working on.)

glebius:
I see. I will try to redo this.

infrastation:
Pushing a commit to a branch makes it appear on the pull request, that is the traditional behaviour on GitHub. This way the requestors can address some of the points raised and then get back to the original discussion. Separate branches usually do the job.

PR #587, opened 2017-02-01
Add parser for pfsync packets.

glebius:
Add parser for pfsync packets, a protocol that synchronizes two (or more pf firewalls.  The parser was originally developed in OpenBSD, then was developed in the FreeBSD tree. Has been used for over a decade.

The compilation is dependable on presense of pfvar.h, just like the pflog parser.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -352,6 +352,7 @@ EXTRA_DIST = \
 	packetdat.awk \
 	pcap_dump_ftell.c \
 	print-pflog.c \
+	print-pfsync.c \
 	print-smb.c \
 	send-ack.awk \
 	smbutil.c \
--- a/configure
+++ b/configure
@@ -4239,7 +4239,7 @@ fi
 done
 
 	if test "$ac_cv_header_net_if_pflog_h" = yes; then
-		LOCALSRC="print-pflog.c $LOCALSRC"
+		LOCALSRC="print-pflog.c print-pfsync.c $LOCALSRC"
 	fi
 fi
 for ac_header in netinet/if_ether.h
--- a/configure.in
+++ b/configure.in
@@ -42,7 +42,7 @@ if test "$ac_cv_header_net_pfvar_h" = yes; then
 	#include <net/if.h>
 	#include <net/pfvar.h>])
 	if test "$ac_cv_header_net_if_pflog_h" = yes; then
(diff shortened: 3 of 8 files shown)


guyharris:
IP protocol number 240 is unassigned, according to [the Assigned Internet Protocol Numbers list](http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml).  Is 240 being used privately for pfsync-over-IP?

glebius:
Yes, that's correct.

guyharris:
So, if there's a protocol that sends these over the wire, is the format either 1) frozen or 2) extensible, so that we don't have to worry about the format of the messages being different in different versions of different OSes?  That's the reason why we require the OS's header file in `print-pflog.c`.

glebius:
Hi! Well, the pfsync protocol comes from OpenBSD and in the first years it was changing very rapidly. Once we took it to FreeBSD, we don't change it, but it is still subject to change. AFAIK, OpenBSD has no intents to standardize it. :( That's why I decided that it will be lesser evil to include system;s header for the protocol definition. At least we will be protected against some change in the protocol, e.g. additional fields or values, but not from renames or removals.

infrastation:
A practicable way to add and maintain pfsync support could be to introduce `-T pfsync` (it would be necessary anyway because protocol number 240 is unassigned, that is, not purposed for private use or experimentation, and can be allocated to something else at any time) and to give the current protocol encoding a name (e.g. "pfsync-2020", "freebsd12" or whatever else makes sense) and to state that this pfsync decoder decodes only this specific variety of the pfsync protocol. For example:

    tcpdump -n -T pfsync ip proto 240

Then when and if the next protocol revision comes into existence (supposedly without the on-the-wire version number again), the old code will still work with old packets, and to disambiguate between the versions it will be enough to add a new `--pfsync-version` option with a sane default value. For example:

    tcpdump -n -T pfsync ip proto 240
    tcpdump -n -T pfsync --pfsync-version=freebsd13 ip proto 240

Once again, the compilation must not depend on FreeBSD and should specify the protocol snapshot in sufficient detail to be maintainable long-term. There should be a new test case with a .pcap file that tries as many code points as possible.

@glebius, please review and tell if you are willing to work on this.

infrastation:
Looks like this is not going to converge in a mutually acceptable way, hence closing.

PR #592, opened 2017-02-22
Make non-privsep tcpdump a configure fail

coypoop:
...unless --without-user is passed
Hopefully more people will be aware of this option as a result.


coypoop:
Feel free to do it any other way, I'm not an autoconf expert.

PR #595, opened 2017-03-20
add NSH ethertype

sirspinach:
tcpdump currently can only read NSH headers that follow VXLAN GPE headers. NSH also has an IEEE allocated Ethertype that allows it to be inserted after the Ethernet layer.

https://tools.ietf.org/html/draft-ietf-sfc-nsh-12#section-12.1

If this pull request needs a test case, I'd be happy to write it. But it might be simple enough that it's not required.


fxlb:
Thanks, please add a test case.

sirspinach:
@fxlb Added a test case. Please let me know if there's anything this patch needs before it can be merged!

infrastation:
Thank you!

PR #597, opened 2017-03-26
IEEE 802.15.4 printer which understands frame version 2 frames, and a…

kivinen:
…lso knows how to print some mac commands and IE contents. Also includes the zep printer to decode ZigBee Encapsulation Protocol frames


mcr:
Could you add some pcap files with some ZBE packets in it, and a make check result?

fxlb:
Thanks. I think you need to 'touch .devel' in your working directory to get more warnings (see CONTRIBUTING file) and see https://travis-ci.org/the-tcpdump-group/tcpdump/jobs/215255654.

kivinen:
Ok, fixed the warnings now. Used the __attribute__ ((unused)) to get rid of the warnings of arguments of ieee802_15_4_frag_frame not being used, as the function is still empty (implementing it is bit tricky, and as I have never seen example frames of that type, there is no point of implementing it).

guyharris:
> Used the attribute ((unused)) to get rid of the warnings of arguments of ieee802_15_4_frag_frame not being used

`__U__` is used elsewhere in tcpdump.

kivinen:
Hmm.. don't find any cases of __U__ in code, and I do not see it being defined anywhere.

guyharris:
> Hmm.. don't find any cases of _U_ in code

(Presumably meaning "don't find any cases of `_U_` in code" - that's "underscore U underscore", not "Markdown markup for boldface U".)

    $ pwd
    /Users/gharris/src/cmd/tcpdump.org/work/tcpdump
    $ egrep _U_ *.c
    machdep.c:abort_on_misalignment(char *ebuf _U_, size_t ebufsiz _U_)
    print-decnet.c:print_elist(const char *elp _U_, u_int len _U_)
    print-esp.c:static void esp_init(netdissect_options *ndo _U_)
    print-esp.c:    _U_
    print-esp.c:    _U_
    print-esp.c:    _U_
    print-fddi.c:fddi_smt_print(netdissect_options *ndo, const u_char *p _U_, u_int length _U_)
    print-forces.c:              uint16_t op_msk _U_, int indent)
    print-forces.c:               uint16_t op_msk _U_, int indent)
    print-forces.c:               uint16_t op_msk _U_, int indent)

        ....

> and I do not see it being defined anywhere.

    $ egrep _U_ aclocal.m4 CMakeLists.txt 
    aclocal.m4:  # so we define _U_ to an empty string.
    aclocal.m4:  V_DEFS="$V_DEFS -D_U_=\"\""
    aclocal.m4:  V_DEFS="$V_DEFS -D_U_=\"__attribute__((unused))\""
    aclocal.m4:  V_DEFS="$V_DEFS -D_U_=\"\""
    CMakeLists.txt:    add_definitions( "-D_U_=" )
    CMakeLists.txt:    add_definitions( "-D_U_=__attribute__((unused))" )
    CMakeLists.txt:    add_definitions( "-D_U_=" )

kivinen:
Ah, U with only one underscores around it, I of course grepped with double underscores, as that is what gcc normally uses. Fixed now to use U with underscores...

PR #599, opened 2017-04-20
Fix two possible issues in decoding BGP Shutdown Communication

job:
1) the condition where the shutdown communication length field claims to be longer then the BGP NOTIFICATION Cease message actually can accommodate.

2) if the shutdown communication payload contains trash, like \0 (NUL) we should be able to print that

### DIFF
--- a/print-bgp.c
+++ b/print-bgp.c
@@ -2679,7 +2679,7 @@ bgp_notification_print(netdissect_options *ndo,
 	     * draft-ietf-idr-shutdown describes a method to send a communication
 	     * intended for human consumption regarding the Administrative Shutdown
 	     */
-	    if((bgpn.bgpn_minor == BGP_NOTIFY_MINOR_CEASE_SHUT ||
+	    if ((bgpn.bgpn_minor == BGP_NOTIFY_MINOR_CEASE_SHUT ||
 		bgpn.bgpn_minor == BGP_NOTIFY_MINOR_CEASE_RESET) &&
 		length >= BGP_NOTIFICATION_SIZE + 1) {
 		    tptr = dat + BGP_NOTIFICATION_SIZE;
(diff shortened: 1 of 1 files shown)


print-bgp.c
```diff
@@ -2679,15 +2679,16 @@ bgp_notification_print(netdissect_options *ndo,
...
 		bgpn.bgpn_minor == BGP_NOTIFY_MINOR_CEASE_RESET) &&
 		length >= BGP_NOTIFICATION_SIZE + 1) {
 		    tptr = dat + BGP_NOTIFICATION_SIZE;
 		    ND_TCHECK2(*tptr, 1);
 		    shutdown_comm_length = *(tptr);
 		    remainder_offset = 0;
 		    /* garbage, hexdump it all */
-		    if (shutdown_comm_length > BGP_NOTIFY_MINOR_CEASE_ADMIN_SHUTDOWN_LEN) {
+		    if (shutdown_comm_length > BGP_NOTIFY_MINOR_CEASE_ADMIN_SHUTDOWN_LEN ||
+			shutdown_comm_length > (length - BGP_NOTIFICATION_SIZE) + 1) {
```
fxlb: Should it be `shutdown_comm_length > length - (BGP_NOTIFICATION_SIZE + 1))` ?

print-bgp.c
```diff
@@ -2698,7 +2699,7 @@ bgp_notification_print(netdissect_options *ndo,
 		    else {
 			    ND_TCHECK2(*(tptr+1), shutdown_comm_length);
 			    ND_PRINT((ndo, ", Shutdown Communication (length: %u): \"", shutdown_comm_length));
-			    safeputs(ndo, tptr+1, shutdown_comm_length);
+			    safeputlchars(ndo, tptr+1, shutdown_comm_length);
```
fxlb: No need for a new print function, you should use `fn_printn(ndo, tptr+1, shutdown_comm_length, NULL);`

print-bgp.c
job: `>` is correct, it is legal for trailing data to exist after a shutdown communication

fxlb: ok

print-bgp.c
```diff
@@ -2698,7 +2699,7 @@ bgp_notification_print(netdissect_options *ndo,
 		    else {
 			    ND_TCHECK2(*(tptr+1), shutdown_comm_length);
 			    ND_PRINT((ndo, ", Shutdown Communication (length: %u): \"", shutdown_comm_length));
-			    safeputs(ndo, tptr+1, shutdown_comm_length);
+			    n_printn(ndo, tptr+1, shutdown_comm_length, NULL);
```
fxlb: needed: s/n_printn/fn_printn/

fxlb:
Merged, thank you.

PR #603, opened 2017-05-18
Add support for MACsec (IEEE 802.1AE-2006)

qsn:


print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+ * MACsec ethertype = 0x88e5
+ * TCI: Tag Control Information, set of flags
+ * AN: association number, 2 bits
+ * SL (short length): 6-bit length of the protected payload, if < 48
+ * Packet Number: 32-bits packet identifier
+ * Secure Channel Identifier: 64-bit unique identifier, usually
+ *     composed of a MAC address + 16-bit port number
+ */
+struct macsec_sectag {
+	u_char  tci_an;
```
guyharris, line 63: This should be `nd_uint8_t`.

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+#if __BYTE_ORDER == __LITTLE_ENDIAN
+	u_char short_length:6,
```
guyharris, line 65: This should be done as a single `nd_uint8_t`, and the bits extracted with shifting and masking, rather than by hoping that the host's bit order corresponds to its byte order (that is ***NOT*** guaranteed to be the case in C!).

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+		     unused:2;
+#else /* __BYTE_ORDER == __BIG_ENDIAN */
+	u_char        unused:2,
+		short_length:6;
+#endif
+	uint32_t packet_number;
```
guyharris, line 71: This should be an `nd_uint32_t`.

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+	u_char secure_channel_id[8]; /* optional */
```
guyharris, line 72: This should be an array of 8 `nd_uint8_t`.

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+} __attribute__((packed));
```
guyharris, line 73: `__attribute__((packed))` is not supported by all compilers.  Using the `nd_` types ensures that the items will appear with no padding, so remove `__attribute__((packed))`.

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+
+	if (sectag->unused || sectag->tci_an & MACSEC_TCI_VERSION) {
+		ND_PRINT((ndo, "%s", istr));
+		return hdrlen + caplen;
+	}
+
+	if (ndo->ndo_eflag) {
+		char buf[128];
+		int n = snprintf(buf, sizeof(buf), "an %d, pn %d, flags %s",
+				 sectag->tci_an & MACSEC_AN_MASK,
```
guyharris, line 157: Use `GET_U_1()` to extract the value of `sectag->tci_an` - `GET_U_1(&sectag->tci_an)`.  That will do the necessary bounds checking.

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+				 EXTRACT_32BITS(&sectag->packet_number),
```
guyharris, line 158: Use `GET_BE_U_4(sectag->packet_number)` to extract the value and do the necessary bounds checking.

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+				 bittok2str_nosep(macsec_flag_values, "none",
+						  sectag->tci_an & MACSEC_TCI_FLAGS));
```
guyharris, line 160: And use `GET_U_1()` here as well - `GET_U_1(sectag->tci_an) & MACSEC_TCI_FLAGS`.

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+		if (n < 0)
+			return hdrlen + caplen;
+
+
+		if (sectag->short_length) {
+			int r = snprintf(buf + n, sizeof(buf) - n, ", sl %d",
+					 sectag->short_length);
```
guyharris, line 167: `GET_U_1(sectag->short_length) & 0x3F` (or give `0x3F` a #define name).

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+			if (r < 0)
+				return hdrlen + caplen;
+			n += r;
+		}
+
+		if (sectag->tci_an & MACSEC_TCI_SC) {
```
guyharris, line 173: `GET_U_1(sectag->tci_an) & MACSEC_TCI_SC`.

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+			uint64_t sci;
+			int r;
+			sci = EXTRACT_64BITS(sectag->secure_channel_id);
```
guyharris, line 176: `sci = GET_BE_U_8(sectag->secure_channel_id);`

print-macsec.c
```diff
@@ -0,0 +1,208 @@
...
+			if (r < 0)
+				return hdrlen + caplen;
+			n += r;
+		}
+
+		ND_PRINT((ndo, "%s, ", buf));
+	}
+
+	len = ieee8021ae_sectag_len(sectag);
+	*length_type = EXTRACT_16BITS(*bp + len);
```
guyharris, line 187: `*length_type = GET_BE_U_2(*bp + len);`

print-macsec.c
```diff
@@ -0,0 +1,202 @@
...
+				return hdrlen + caplen;
+			n += r;
+		}
+
+		ND_PRINT((ndo, "%s, ", buf));
+	}
+
+	len = ieee8021ae_sectag_len(ndo, sectag);
+	*length_type = GET_BE_U_2(*bp + len);
+	if (ndo->ndo_eflag && *length_type > ETHERMTU && !(GET_U_1(sectag->tci_an) & MACSEC_TCI_E))
```
fenner, line 182: print-ether.c uses `MAX_ETHERNET_LENGTH_VAL`, defined in tcpdump's <ethertype.h>, instead of `ETHERMTU` which might be in the system's <net/ethernet.h>

guyharris: I'll read the spec to see why an 802.1 standard would *need* to know about a particular 802.x network's maximum packet size.

fenner: The `length_type` can either be a frame length, if it's equal to or smaller than the MTU, or it's an ethertype.  I don't quite know why it's relevant in this branch, but it's checking that length_type contains an ethertype.

guyharris: > The `length_type` can either be a frame length, if it's equal to or smaller than the MTU, or it's an ethertype. I don't quite know why it's relevant in this branch, but it's checking that length_type contains an ethertype.

It's no longer doing so; it's now punting that to the Ethernet dissector, which is as it should be.

fenner: Excellent!

PR #605, opened 2017-05-30
RFC 6608 specifies subcode 0 to be value "Unspecified Error"

job:


infrastation:
The first change is fine, the second will make a problem. The last element on a `struct tok` array must have the second field set to NULL to indicate the end of array for `tok2strbuf()` (and thus `tok2str()`). In that case the first field remains unused, as a matter of code style it is usually 0. Commit da97db5 makes a corrected change on your behalf.

PR #607, opened 2017-06-19
Improvements of NTP printer

mlichvar:
Here are some improvements for the NTP printer to support sub-second polling intervals, MACs using 160-bit digests and reference IDs generated from IPv6 addresses. There is also a fix for printing key IDs.

I'd like to add a test for the NTP printer, but there is a problem that NTP timestamps are printed in local time and won't match the expected output unless the timezone is the same. Would it be ok to set TZ=UTC in tests/TESTonce?

### DIFF
--- a/print-ntp.c
+++ b/print-ntp.c
@@ -309,10 +309,10 @@ ntp_print(netdissect_options *ndo,
 
 	if ( (sizeof(struct ntpdata) - length) == 16) { 	/* Optional: key-id */
 		ND_TCHECK(bp->key_id);
-		ND_PRINT((ndo, "\n\tKey id: %u", bp->key_id));
+		ND_PRINT((ndo, "\n\tKey id: %u", EXTRACT_32BITS(&bp->key_id)));
 	} else if ( (sizeof(struct ntpdata) - length) == 0) { 	/* Optional: key-id + authentication */
 		ND_TCHECK(bp->key_id);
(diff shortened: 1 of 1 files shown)


infrastation:
I have looked through the commits, the three smaller ones are OK but the size checks in the 160-bit digest one don't look quite right. I can cherry-pick the smaller ones now if you would be comfortable rebasing your changes on the master branch afterwards. Then I can make another attempt at the remaining change later.

sjvudp:
I could try to cherry-pick/merge this request into 625.

mlichvar:
Ok. I'm not sure what's wrong with the size checks. I've rebased the patchset on the current master branch and reordered them so the digest patch is last if you want to skip it for now.

infrastation:
The first three commit are now in the master branch, thank you. The remaining one needs another round and will be dealt with later.

sjvudp:
As there seem overlaps with pull request 625 (which is larger), I wonder whether it would be easier to try 625 first. Anyway, I started to manually merge the fixes I agree with to the "f-NTP-mode6" branch.

infrastation:
The 160-bit digest commit is now in the master branch too, thank you. After studying the code once again it became clear that the length checks were OK in this pull request, they were not completely OK in the original file. Commit 0559dc9 addresses that.

NTP timestamps are now always printed in UTC so if you have a test case to contribute it would come handy.

mlichvar:
Thanks! I'll make a new PR for the NTP test.

PR #608, opened 2017-06-30
assorted RADIUS updates

herwinw:
Added various RADIUS RFCs to the dissector. Every change that is more than just a trivial addition the the attribute array contains unit tests.


herwinw:
You're right, fixed that one

infrastation:
Please let us a little more time to review once again.

infrastation:
The first two commits are now in the master branch. RFC 4072 and RFC 5090 commits are OK but I didn't pick them because they depend on the previous commits (can be put in front and merged after a rebase and conflict resolution if anybody is willing), the other 3 commits add new code and require some more attention.

infrastation:
The RFC5176/RFC5580 commit adds custom lookup code instead of reusing `tok2str()`, is there a reason for that?

herwinw:
Since "I didn't know about that" is not really a valid reason: I've updated the change to use `tok2str`

infrastation:
The RFC 3162 commit needs the following issues fixed in the function `print_attr_netmask6()`:
* `ND_TCHECK2()` checks the buffer by a fixed length of 18 but the TLV value field is variable-length because of the Prefix field it includes:

      The Prefix field is up to 16 octets in length.  Bits outside of
      the Prefix-Length, if included, must be zero.

* The TLV value length is not checked to be consistent with the masklen. For example, what should be done if there are 4 bytes of address and a /128 netmask in the TLV?.

infrastation:
The RFC 4818 commit will be OK after the points raised above are resolved. This way, the previous comment describes the only remaining issues for this pull request. If you have time, please also try to fix the compiler warnings these changes introduce (see the Travis CI build log or `touch .devel && ./configure` to see locally). Thank you.

herwinw:
I believe I've fixed everything except the "The TLV value length is not checked to be consistent with the masklen". I'm not really sure what the best option is here, the RFC doesn't this case. They're now filled with 0, maybe an error (or a warning) would be more suitable?

infrastation:
Thank you, see commit 0bc9b44 for some more fixups. As far as I could figure it out, there is no trivial way to test for non-zero bits outside the declared prefix length, so I had left this check unimplemented (it does not affect safety within tcpdump).

PR #609, opened 2017-07-01
Generate "CMakeLists.txt" to open tcpdump project with many IDEs

olibre:
This implementation of "CMakeLists.txt" is NOT a replacement
of automake/autotools/autoconf but a generated file by
"./configure" like the generation of file "Makefile".

The main objective is just to ease the opening of the tcpdump project
using many CMake-friendly IDE as Qt Creator, KDevelop, 
GNOME Builder, NetBeans, CLion, Visual Studio... and many more...

The file "./configure" has been modified because I do not know
how to handle some double quotes `\"\"` within `-DMACRO=""`  :-/
I hope "./configure" is not often generated from "configure.in"...

The "configure.in" has also been modified to add generation
of the "CMakeLists.txt" from "CMakeLists.txt.in".

The new added file "CMakeLists.txt.in" is the mirror of
the reference file "Makefile.in" in order to ease maintenance.
Any modification within "Makefile.in" should be easily ported
to "CMakeLists.txt.in".

However, I have not implemented all features from
files "Makefile.in" and "Makefile-devel-adds".
But I may do it if some people request it...

For the moment, the generated "CMakeLists.txt" is basic
and does not contains all the nice features we can expect
from advanced CMake-based project. 
However, if requested, I may add the nice features 
I have already provided to project tcpflow: 
https://github.com/simsong/tcpflow/pull/145/files

I do not plan to fully replace "./configure"
by a standalone CMake because the "./configure"
is very large and handle so many platforms
and different configurations.
(nearly impossible to test all the combinations)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 *.orig
 *.rej
 Makefile
+CMakeLists.txt
+CMakeLists.txt.user
 *~
 *.o
 libnetdissect.a
--- a/CMakeLists.txt.in
+++ b/CMakeLists.txt.in
@@ -1,0 +1,297 @@
+#  Copyright (c) 2017 The TCPdump team
+#
+#  2-clause BSD license
+
+#
+# The file "CMakeLists.txt" is generated from "CMakeLists.txt.in" by the script ./configure
+# Remember to edit "CMakeLists.txt.in", not "CMakeLists.txt"
+#
+# Please keep correspondance between "Makefile.in" and "CMakeLists.txt.in" to ease maintenance.
+# Please verify your CMake-related changes using these command line options:
+# mkdir -p build ; cd build ; cmake .. -Wdeprecated -Wdev --warn-uninitialized
+#
+
+cmake_minimum_required(VERSION 2.8.12)
+# The most recent feature used in this file is "add_compile_options()"
+# that requires at least CMake-2.8.12 (Oct 2013)
+
+project(tcpdump LANGUAGES C)
+
+add_compile_options(@V_CCOPT@)
+string(REPLACE "-I" " " V_INCLS "@V_INCLS@")
+include_directories(. ${V_INCLS})
+add_definitions(@DEFS@ @CPPFLAGS@ @V_DEFS@)
+add_compile_options(@CFLAGS@)
+
+
+# Standard LIBS
+string(REPLACE " "  ""  LIBS "@LIBS@")
+string(REPLACE "-l" ";" LIBS ${LIBS})
+
+set(CSRC setsignal.c tcpdump.c)
+
+set(LIBNETDISSECT_SRC
+	addrtoname.c
+	addrtostr.c
+	af.c
+	ascii_strcasecmp.c
+	checksum.c
+	cpack.c
+	gmpls.c
+	gmt2local.c
+	in_cksum.c
+	ipproto.c
+	l2vpn.c
+	machdep.c
+	nlpid.c
+	oui.c
+	parsenfsfh.c
+	print.c
+	print-802_11.c
+	print-802_15_4.c
+	print-ah.c
+	print-ahcp.c
+	print-aodv.c
+	print-aoe.c
+	print-ap1394.c
+	print-arcnet.c
+	print-arp.c
+	print-ascii.c
+	print-atalk.c
+	print-atm.c
+	print-babel.c
+	print-beep.c
+	print-bfd.c
+	print-bgp.c
+	print-bootp.c
+	print-bt.c
+	print-calm-fast.c
+	print-carp.c
+	print-cdp.c
+	print-cfm.c
+	print-chdlc.c
+	print-cip.c
+	print-cnfp.c
+	print-dccp.c
+	print-decnet.c
+	print-dhcp6.c
+	print-domain.c
+	print-dtp.c
+	print-dvmrp.c
+	print-eap.c
+	print-egp.c
+	print-eigrp.c
+	print-enc.c
+	print-esp.c
+	print-ether.c
+	print-fddi.c
+	print-forces.c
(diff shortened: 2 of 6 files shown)


guyharris:
For libpcap, the `CMakeLists.txt` file is explicitly intended to provide an alternative to autotools, because a configure script in Bourne shell wouldn't be usable on Windows unless you've installed the Bourne shell, and the necessary tools, and if it manages to figure out how to use `cl` as `$CC`.

For tcpdump, that'd ultimately be what I'd want, too.  Why not just go with a `CMakeLists.txt` file, so that you can use CMake without autotools?

(I'm not planning on deprecating autotools for either project, just providing CMake as an alternative.)

olibre:
> Why not just go with a `CMakeLists.txt` file, so that you can use CMake without autotools?

Because without autotools, we have to write and maintain lots of code in `CMakeLists.txt` to provide an alternative to the generated header `config.h`. 

Question: **Can tcpdump project be compiled without having to generate any `config.h` ?**

I have started to generate the `config.h` for the **tcpdump** project. 
But this is can become difficult in some cases. See my TODO comments about that work in progress:
https://github.com/olibre/tcpflow/blob/8396205b2e254c4c47d70250bab4abe34aae5513/src/CMakeLists.txt#L14

User @Code7R has provided a wonderful work on [**IceWM fork**](https://github.com/bbidulock/icewm/commits?author=Code7R) to propose an standalone CMake alternative to autoconf/automake/autotools:
https://github.com/bbidulock/icewm/pull/21/files

**Is it what you want for tcpdump project?**

olibre:
I have just seen you wrote the alternative for autotools to generate `config.h` for libpcap:
https://github.com/the-tcpdump-group/libpcap/commits/master/cmakeconfig.h.in

Yes we can together provide a standalone CMake alternative in parallel of the classic autotools 😄.
What do you plan?

mcr:
Guy Harris <notifications@github.com> wrote:
    > (I'm not planning on deprecating autotools for either project, just
    > providing CMake as an alternative.)

Do we have a developer Makefile target (or script) that will update the
CMakefile from the Makefile(.in)?

I wouldn't be upset to obsolete autotools myself :-)

olibre:
Hi @mcr  

I think @guyharris wants a standalone CMake alternative that does not depend on `./configure` (for example to be opened on Windows without any Cygwin/MinGW installed) and both auto(conf|tool|make) and CMake have to be maintained in parallel.

My pull request is an intermediate step: the `CMakeLists.txt` still requires `./configure` to be generate. But maintenance is very easy as I have written the `CMakeLists.txt.in` as a mirror of the `Makefile.in` (however in some cases I have adopted the CMake syntax spirit).

Until someone takes time to write a full validated standalone CMake alternative, my pull request can be accepted and merged. It does not break any thing, and future standalone `CMakeLists.txt` will be a replacement.

mcr:
O. Libre <notifications@github.com> wrote:
    > I think @guyharris wants a standalone CMake alternative that does not
    > depend on ./configure (for example to be opened on Windows without any
    > Cygwin/MinGW installed) and both auto(conf|tool|make) and CMake have to
    > be maintained in parallel.

Yes, I want that too.
Once created, I'd like CMake to run without any autotools.

    > My pull request is an intermediate step: the CMakeLists.txt still
    > requires ./configure to be generate. But maintenance is very easy as I
    > have written the CMakeLists.txt.in as a mirror of the Makefile.in
    > (however in some cases I have adopted the CMake syntax spirit).

My idea is that CMakeLists.txt could be generated by a Makefile.
So a developer could update the CMakeList.txt as part of the release
.tar.gz creation process.

guyharris:
@olibre:

> I think @guyharris wants a standalone CMake alternative that does not depend on ./configure (for example to be opened on Windows without any Cygwin/MinGW installed)

Yes.

> and both auto(conf,tools,make) and CMake have to be maintained in parallel.

Yes.  That's how Wireshark does it.

> I do not know if @guyharris is volunteer to perform this double maintenance.

I'm doing a lot of the autotools work anyway, in which case "double maintenance" would be just "change both of them", although I'd prefer to have everybody who changes one of them change both, which is usually how it works with Wireshark.

And most contributions to tcpdump that change the autotools stuff are probably "here's `print-frobozz.c`, which dissects the Frobozz protocol", where you add one line to `Makefile.in`, so we can just reply to pull requests that *don't* add one line to `CMakeLists.txt` with "please add `print-frobozz.c` to `CMakeLists.txt`, too".

@mcr:

> I wouldn't be upset to obsolete autotools myself :-)

That does require anybody building tcpdump or libpcap to have CMake installed, but 1) a binary package is available for macOS, 2) a binary package for "Linux x86_64" is available, 3) I think the major Linux distributions have packages as well, 4) the \*BSDs all seem to have packages, 5) Oracle's package repository for Solaris 11 has a binary package and OpenCSW appears to have one as well, 6) perzl.org appears to have it for AIX 5.1 and later, and 7) if all else fails, you can try building it from source (I couldn't find it at the HP-UX Porting and Archive Centre, and I don't know what the dead UN\*Xes have in the way of third-party package collections).

> My idea is that CMakeLists.txt could be generated by a Makefile.
> So a developer could update the CMakeList.txt as part of the release .tar.gz creation process.

Yes, we could do that - just have `make releasetar` replace `@LIBNETDISSECT_SRC@` in `CMakeLists.txt.in` with the contents of `LIBNETDISSECT_SRC`.

guyharris:
> I think @guyharris wants a standalone CMake alternative that does not  depend on ./configure (for example to be opened on Windows without any Cygwin/MinGW installed)

Done.  It now builds, at least with MSVC 2015 on Windows 7, with a few warnings.

> Once created, I'd like CMake to run without any autotools.

Done.

Issue #586, opened 2017-02-01
Passing same buffer as in and out to EVP_Cipher() in print-esp.c


glebius:
The prototype of EVP_Cipher() doesn't allow to pass const buffer as in. It seems that the feature of passing the same buffer as in and out is undocumented and may disappear. Better to allocate a temporary buffer for out.

At FreeBSD we currently fixed compilation with __DECONST() macro, but this isn't right solution.

msehnout:
The buffer should also have size that is a multiple of a block size of given cipher. It works fine on most architectures, but fails on ppc64le. There is a RH bugzilla for this with more information: https://bugzilla.redhat.com/show_bug.cgi?id=1441963

guyharris:
The master branch has both fixes.

PR #613, opened 2017-07-12
Use separate buffers for encrypted and decrypted content

msehnout:
This issue is described in #586.

It basically just allocates a new buffer, which is passed into the `EVP_Cipher` function. The rest is just debugging, which I'll remove.

Do you agree with this solution?

### DIFF
--- a/print-esp.c
+++ b/print-esp.c
@@ -191,23 +191,23 @@ int esp_print_decrypt_buffer_by_ikev2(netdissect_options *ndo,
 		return 0;
 	if (EVP_CipherInit(ctx, sa->evp, sa->secret, NULL, 0) < 0)
 		(*ndo->ndo_warning)(ndo, "espkey init failed");
-        EVP_CipherInit(ctx, NULL, NULL, iv, 0);
+	EVP_CipherInit(ctx, NULL, NULL, iv, 0);
 
-        /* We need a block size */
-        block_size = EVP_CIPHER_CTX_block_size(ctx);
-        /* We need the buffer size to be multiple of a block size */
-        output_buffer_size = len + (block_size - len % block_size);
-        output_buffer = (u_char *)calloc(output_buffer_size, sizeof(u_char));
+	/* We need a block size */
+	block_size = EVP_CIPHER_CTX_block_size(ctx);
+	/* We need the buffer size to be multiple of a block size */
+	output_buffer_size = len + (block_size - len % block_size);
+	output_buffer = (u_char *)calloc(output_buffer_size, sizeof(u_char));
 	/* EVP_Cipher output buffer should be different from the input one.
-         * Also it should be of size that is multiple of cipher block size. */
+	 * Also it should be of size that is multiple of cipher block size. */
 	EVP_Cipher(ctx, output_buffer, buf, len);
 	EVP_CIPHER_CTX_free(ctx);
 
-        buf_mut = (u_char*) buf;
-        /* Of course this is wrong, because buf is a const buffer, but changing this
-         * would require more complicated fix. */
-        memcpy(buf_mut, output_buffer, len);
-        free(output_buffer);
+	buf_mut = (u_char*) buf;
+	/* Of course this is wrong, because buf is a const buffer, but changing this
+	 * would require more complicated fix. */
+	memcpy(buf_mut, output_buffer, len);
+	free(output_buffer);
 
 	ndo->ndo_packetp = buf;
 	ndo->ndo_snapend = end;
--- a/print-esp.c
+++ b/print-esp.c
@@ -156,7 +156,8 @@ int esp_print_decrypt_buffer_by_ikev2(netdissect_options *ndo,
 {
 	struct sa_list *sa;
 	const u_char *iv;
-	int len;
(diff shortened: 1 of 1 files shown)


infrastation:
Generic observations: this change fixes one warning and introduces two new warnings in `esp_print_decrypt_buffer_by_ikev2()`, it does not fix the warning in `esp_print()`.

As an unrelated note the Travis build fails the test due to some debugging output but `make check` fails to fail after that, at least as far as exit codes (and thus Travis scripts) are concerned. Congratulations, you have found a bug in tcpdump test scripts (just fixed with commit d72adb9).

print-esp.c
```diff
@@ -190,15 +190,29 @@ int esp_print_decrypt_buffer_by_ikev2(netdissect_options *ndo,
...
+        EVP_CipherInit(ctx, NULL, NULL, iv, 0);
+
+        printf("\n\nDebug!\n\n");
+        int block_size = EVP_CIPHER_CTX_block_size(ctx); // We need a block size
+        int cipher_nid = EVP_CIPHER_CTX_nid(ctx);	 // Just for debugging purposes
+        printf("Block size: %d\n", block_size);
+        printf("Cipher NID: %d // 419 = NID_aes_128_cbc\n", cipher_nid);
+        // We need the buffer size to be multiple of a block size
+        int output_buffer_size = len + (block_size - len % block_size);
+        u_char *output_buffer = (u_char *)calloc(output_buffer_size, sizeof(u_char));
```
guyharris: I don't see a call to free() anywhere in this change....

guyharris:
> It basically just allocates a new buffer

Where does it free the buffer?

msehnout:
Thanks for your responses. I'm aware of the warning in `esp_print()`, but I can just copy the same solution once we agree on it. Those warning are easy to fix, if you mean these:
```txt
./print-esp.c: In function ‘esp_print_decrypt_buffer_by_ikev2’:
./print-esp.c:196:9: warning: ISO C90 forbids mixed declarations and code [-Wpedantic]
         int block_size = EVP_CIPHER_CTX_block_size(ctx); // We need a block size
         ^
./print-esp.c:196:58: warning: C++ style comments are not allowed in ISO C90 [enabled by default]
         int block_size = EVP_CIPHER_CTX_block_size(ctx); // We need a block size
                                                          ^
./print-esp.c:196:58: warning: (this will be reported only once per input file) [enabled by default]
./print-esp.c:201:9: warning: ISO C90 forbids mixed declarations and code [-Wpedantic]
         int output_buffer_size = len + (block_size - len % block_size);
         ^
```

The place where to free the buffer is a question to discuss. The previous solution was to use `buf` as a place to store decrypted message. I could just copy the new buffer into `buf`, but this would again discard the const qualifier, but at least it would fix this bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1441963
Other solutions would have to be more complicated because I need to pass this buffer into the `ndo` variable, thus `ndo` cannot outlive this buffer.

Do you have any recommendation on where to free this buffer?

msehnout:
I decided to just copy the temporary buffer into the old one (passed as argument). This does not solve the warning about discarding const qualifier, but should prevent undefined behavior when using OpenSSL api.

msehnout:
Any comments on this PR?

guyharris:
> Any comments on this PR?

1) Please remove the debugging printouts, as they're making the "make check" part of the Travis build fail.

2) There are *two* locations where we're using the same buffer - see also the code beginning around line 714.

3) It's not the *ideal* solution, in which we wouldn't overwrite the packet data; I'm looking at that, but it's a bit more complicated to do that right, so we should probably go with a solution like this for now.

msehnout:
Done

infrastation:
On my system this removes the old warnings about `print-esp.c` and adds some new ones.

guyharris:
The only way to remove all the warnings is not to have decrypted data in the original buffer.  As the comment in the code says, "Of course this is wrong, because buf is a const buffer, but changing this would require more complicated fix.", and as the commit message says, "There is also a second problem, that we write the output to a const buffer, but this issue is not addressed with this PR.", and as my comment says, "It's not the *ideal* solution, in which we wouldn't overwrite the packet data; I'm looking at that, but it's a bit more complicated to do that right, so we should probably go with a solution like this for now."

infrastation:
Yes, but here is something besides that:
```
./print-esp.c:160:23: warning: unused variable ‘cipher_nid’ [-Wunused-variable]
./print-esp.c:626:18: warning: unused variable ‘cipher_nid’ [-Wunused-variable]
./print-esp.c:625:10: warning: unused variable ‘buf_mut’ [-Wunused-variable]
 ```

guyharris:
Yes, I noticed that `cipher_nid` was unused when comparing @msenhout's changes with the ones I was working on (where you *don't* copy back to the buffer).  I've checked changes to remove `cipher_nid` completely (884b1b45c8a5ebd204a8b4aa4db69aea16aef562) and remove `buf_mut` from the routine where it isn't used (8205015f71629c5f11c3e067fcb744bbaecd57da).

infrastation:
Thank you.

PR #614, opened 2017-07-17
Packet sanitization and IP masking

lilchurro:
Built two new flags to sanitize packets as they are written to savefiles:
- 0 zeroes out the packet payload bytes after (the first encountered) TCP/UDP header
- 00 truncates the packet right after the (first encountered) TCP/UDP header
- * [mask_ip] strips the external IP addresses (i.e., addresses outside of the reserved netblocks specified in RFC5735) and replaces it with [mask_ip].

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,1 +1,1 @@
-4.11.0_CyberReboot_mod_r4
+4.11.0_CyberReboot_mod_r5
--- a/VERSION
+++ b/VERSION
@@ -1,1 +1,1 @@
-4.10.0-PRE-GIT
+4.11.0_CyberReboot_mod_r4
--- a/clean_cap_dump.c
+++ b/clean_cap_dump.c
@@ -187,7 +187,7 @@ pcap_mod_and_dump(u_char *user, const struct pcap_pkthdr *h, const u_char *sp,
 	switch(dlt) {
 	case DLT_EN10MB:
 		if ((ip = get_iph_ptr(h, modp)) == NULL)
-		break;
+		    break;
 		p_len -= ETHER_HDRLEN;
 
 		if (mask_ip_flag && maskIP != NULL) {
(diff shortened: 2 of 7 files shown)


tcpdump.c
```diff
@@ -2372,6 +2398,7 @@ static void
 dump_packet_and_trunc(u_char *user, const struct pcap_pkthdr *h, const u_char *sp)
 {
 	struct dump_info *dump_info;
+    int dlt;
```
guyharris: Please use 8-position indentation (tabs or 8 spaces - note that on the operating system where this was first developed, tab stops were every 8 spaces), as is done in the rest of the file.

clean_cap_dump.c
```diff
@@ -0,0 +1,239 @@
...
+#define MAXPACKET   65549
+
+
+/* Returns the pointer of the IP header or NULL if that couldn't be found. */
+u_char *
+get_iph_ptr(const struct pcap_pkthdr *h, u_char *bp) {
+    u_short length_type;
+    u_char *s;
+    const struct ip *ip;
+    const struct ether_header *ep = (const struct ether_header *)bp;
```
guyharris: Not all packets processed by tcpdump are Ethernet packets.

tcpdump.c
lilchurro: Thanks! Will do. Please let me know if there's anything else I need to change. (Amcurrently  writing an email to the mailing list explaining this pull request.)

clean_cap_dump.c
lilchurro: Yes, I know this; this is explained in my email. Sorry for the delay :)

VERSION
```diff
@@ -1 +1 @@
-4.10.0-PRE-GIT
```
guyharris: Don't change our version number, please.

lilchurro:
Opened pull request #615 to use different branch; retiring this pull request.

PR #615, opened 2017-07-18
Packet sanitization and IP masking

lilchurro:
Created a separate branch so I can keep upstream and workgroup README and version strings separate.


infrastation:
As a general note, this problem is not new and similar tools already exist:
* http://scrub-tcpdump.sourceforge.net/docs.php
* https://www.wireshark.org/lists/wireshark-users/201201/msg00106.html
* https://blog.packet-foo.com/2013/07/trace-file-sanitization-for-network-analysts/
* http://www.tm.uka.de/software/pktanon/links/index.html
* https://github.com/thepacketgeek/sanicap

If you still find it better to suggest a new solution, the proposed changes need to be one clean commit, which explains why this specific solution is better.

tcpdump.c
```diff
@@ -87,6 +87,7 @@ The Regents of the University of California.  All rights reserved.\n";
 #include <sys/nv.h>
 #endif	/* HAVE_CASPER */
 #endif	/* HAVE_CAPSICUM */
+#include <arpa/inet.h>
```
gvanem: Does this exist on all platforms? No.

clean_cap_dump.c
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/* Returns 1 if IP address presented in nd_ipv4 falls within reserved IP range;
+ * else returns 0. */
+static int
+is_reserved(nd_ipv4 a) {
+	/* List of all the reserved IPv4 address spaces, per RFC5735,
+	 * ...PROVIDED IN NETWORK BYTE ORDER!! */
+	struct netblock specialblock[] = {
+		{ .netip = 0x00000000, .netmask = 0x000000ff  }, /* 0.0.0.0/8 */
```
gvanem: This C99 (?) syntax is not universally available.

guyharris: For Wireshark, we currently require compilers that support these C99 features:

- flexible array members
- compound literals
- designated initializers
- "//" comments
- mixed declarations and code
- new block scopes for selection and iteration statements (that is, declaring the type in a for-loop like: `for (int i = 0; i < n; i++) ;)`
- macros with a variable number of arguments (variadic macros)
- trailing comma in enum declarations
- inline functions (guaranteed only by use of `glib.h`)

Those features are supported by all versions of GCC and Clang I've dealt with, and I think it's supported by sufficiently-recent versions of UN\*X vendor C compilers.  Before expanding the list of C99 features, I dug through Microsoft's documentation to see which compilers support which features; as I remember, Visual Studio 2013 (the compiler currently used on the Wireshark buildbots) support all of those.  The Wireshark Developer's Guide also mentions VS 2010; I think it might support them as well.

So, are there any compilers that we still want to support and that don't support those features?  If not, perhaps it's time to require compilers that support those features to build tcpdump.

infrastation: @gvanem, could you clarify if it is not available because you use a compiler that does not support C99 in principle or because tcpdump runs the compiler with C99 disabled?

lilchurro:
Hi there!

I'm currently working on the `+#include <arpa/inet.h>` issue, and a better description for the PR request, but I'd like to take a moment to respond to the question of "Why does this PR exist?"

As a former network analyst, I'm certainly aware that tools exist that sort of do these anonymization/privatization and sanitization features, but there are (in my experience) some not-insignificant issues with their adoption. Our motivation for ultimately adding this to tcpdump are the following:

1. **Simplicity and automation**. We are trying to automate the collection and analysis of network traffic coming off of SDN-enabled switches, and we wanted something that did bulk sanitization without requiring multistep or manual processes (i.e., first collecting, then moving, then scrubbing, etc.) because...

2. **Ability to scale**. Most of our work entail large packet captures (e.g., hundreds of GBs), which can pose scalability issues when running subsequent tools, especially if they require writing full packets to memory before sanitizing and anonymizing. We had looked at scapy, for example (which is a backbone to some of these tools), which does not seem to handle scaling challenges particularly well. 

3. **Thorough sanitization**. In the discussions referenced, one will see comments about sanitizing higher layer protocols (e.g., DNS) but potentially omitting or mishandling another (e.g., HTTP). As well, a problem in one layer can undo the anonymity offered by scrubbing a lower layer protocol. This approach may not solve every privacy concern, but completely zeroing out (or truncating) all payload data above TCP/UDP seems to pretty much resolve many of the more conventional privacy issues.

4. **Runs cleanly on Linux**. We require something that can be run on Linux as that's where we do the bulk of our work. We are also finding that Linux is typically the base OS on "bare metal" switches / "disaggregated" network gear. Windows-only based approaches (like TraceWrangler) are challenging for those of us running tcpdump on primarily Linux-based environments or on physical network switches.

5. **Is an active project**. Ideally, we'd like something that is maintained/not orphaned, and is stable. For example, SCRUBtcpdump looks promising, but doesn't seem to have been updated in a decade, and segfaults from the command line with unexpected input. At the risk of stating the obvious, tcpdump's long history and widescale adoption, significant userbase, active community and development base makes it an ideal and attractive method.

Finally, there's the utility of the proposed -00 option which shrinks the size of data collection. This reduction has significant impact for folks doing downstream ingest and analytics of really large network collections. (Sending 1GB to a GPU farm is way better than 10GB!) Having this functionality built into tcpdump opens it up as a really great option not just for traditional networking folks, but for people doing research in machine learning (e.g., our team) in computer networks and security.

All that being said, thanks so much for your input and guidance! I'll work on better defining the description and the code issues pointed out.

tcpdump.c
d-grossman: @gvanem which platforms are you concerned about?

guyharris: Windows (and possibly DOS).

clean_cap_dump.c
gvanem: If it's *designated initializers* that Guy wrote, then MSVC-2013 does not have that. I'm not sure how I can disable C99. So no.

tcpdump.c
gvanem: @guyharris DOS (i.e. Watt-32) do have `<arpa/inet.h>`. But I'm more concerned about Windows off-course.

clean_cap_dump.c
guyharris: > If it's *designated initializers* that Guy wrote, then MSVC-2013 does not have that

"Does not have that" in the sense of "[this page](https://msdn.microsoft.com/en-us/library/81k8cwsz%28v=vs.120%29.aspx?f=255&MSPPError=-2147217396) doesn't mention it" or in the sense of "Microsoft didn't implement it"?  The feature [appears to exist, with some bugs, in VS 2013](https://connect.microsoft.com/VisualStudio/feedback/details/875095/vs-compiler-chokes-on-designated-initializer-that-contains-a-variable-whose-type-corresponds-to-another-struct).

mcr:
lilchurro <notifications@github.com> wrote:
    > 2 Ability to scale. Most of our work entail large packet captures

    > 3 Thorough sanitization. In the discussions referenced, one will see

So, if you are saying that you would typically want to run the sanitization
*as* you capture, then I get why you want it in tcpdump.  That way there
is never a file that is non-sanitized.

mcr:
Guy Harris <notifications@github.com> wrote:
    > which features; as I remember, Visual Studio 2013 (the compiler
    > currently used on the Wireshark buildbots) support all of those.

...
If this compiler is redily available, then I think that covers windows.

My only concern is whether or not someone is still trying to build for
some old HPUX or something like that.  I'm happy to abandon them :-)


-- 
]   Michael Richardson, Sandelman Software Works        | network architect  [ 

lilchurro:
Hello again... I'm having some issues compiling on Windows. (Using Visual Studio 2017, and attempting to compile with headers and libraries as provided by WpcapSrc_4_1_3.zip.)

I'm trying to compile just the master branch of tcpdump (i.e., without my changes), but am stuck with some dependency issues (among other things). It appears to me that windump version 4.9 doesn't compile on Windows anyway, as it refers to the file util.c, which doesn't exist. Also, I'm having issues using wpcapsrc 4.1.3, as the pcap.h included therein contains reference to sys/time.h -- so is there a specific library that I should be using? The win32 readme seems vague about this.

Is there some way I can get assistance in compiling the master branch on Windows? It seems like I should get that working before trying to verify that my changes can compile.

lilchurro:
FYI, I've added new tests to verify the anonymization and packet sanitization for TCP/UDP packet payloads, and have integrated them into TESTrun.sh. I noticed that there wasn't really a way to test for cases where `-w [savefile]` differs from the usual tcpdump output, so the new test file `TESToutfile` checks for such a case. For ease of use, it's basically coded in the same structure as TESTonce and uses TESTOUTLIST to keep track of tests related to savefiles.

If there's something else I can do to move this PR along, please let me know; otherwise, I will consider this about as done as it could be. 🙂

lilchurro:
Ping... is there anything else I can do to move this PR along?

clean_cap_dump.c
```diff
@@ -0,0 +1,245 @@
...
+	if (is_reserved(ip->ip_src) == 0) {
+		nd_ipv4 mask = network_uint_to_nd_ipv4(m);
+		memcpy(&(ip->ip_src), &mask, sizeof(mask));
+	}
+	if (is_reserved(ip->ip_dst) == 0) {
+		nd_ipv4 mask = network_uint_to_nd_ipv4(m);
+		memcpy(&(ip->ip_dst), &mask, sizeof(mask));
+	}
+
+	return 0;
```
d-grossman: 1 if not reserved? where does this occur?

clean_cap_dump.c
```diff
@@ -0,0 +1,245 @@
...
+}
+
```
d-grossman: /* comment of what function does */

clean_cap_dump.c
```diff
@@ -0,0 +1,245 @@
...
+void
+pcap_mod_and_dump(u_char *user, const struct pcap_pkthdr *h, const u_char *sp,
+	      int dlt, int no_payload_flag, int mask_ip_flag, const char *maskIP) {
```
d-grossman: can any of the incoming pointers be Null?

clean_cap_dump.c
```diff
@@ -0,0 +1,245 @@
...
+
+	memcpy(modp, sp, h->caplen);
+
+	f = (FILE *)user;
+	sf_hdr.ts.tv_sec    = h->ts.tv_sec;
+	sf_hdr.ts.tv_usec   = h->ts.tv_usec;
+	sf_hdr.caplen       = h->caplen;
+	sf_hdr.len          = h->len;
+
+	switch(dlt) {
```
d-grossman: will there be more things processed by this switch? yes -> ok 
no -> probably should become an if

clean_cap_dump.c
```diff
@@ -0,0 +1,245 @@
...
+	case DLT_EN10MB:
+		if ((ip = get_iph_ptr(h, modp)) == NULL)
+			break;
+		p_len -= ETHER_HDRLEN;
```
d-grossman: is p_len always >= ETHER_HDRLEN?

clean_cap_dump.c
```diff
@@ -0,0 +1,245 @@
...
+				struct udphdr *u = (struct udphdr *)(ip+ph_len);
+				if (p_len < sizeof(struct udphdr) ||
+				    EXTRACT_16BITS(&u->uh_ulen) < sizeof(struct udphdr)) {
+					break;
+				} else {
+					p_end = (u_char *)(ip + ph_len + sizeof(struct udphdr));
+				}
+			}
+
+			if (no_payload_flag > 1) {
```
d-grossman: what are valid values for no_payload_flag?

README.md
```diff
@@ -1,7 +1,6 @@
 # tcpdump
 
-[![Build
-Status](https://travis-ci.org/the-tcpdump-group/tcpdump.png)](https://travis-ci.org/the-tcpdump-group/tcpdump)
+[![Build Status](https://travis-ci.org/lilchurro/tcpdump.svg?branch=send-upstream)](https://travis-ci.org/lilchurro/tcpdump)
```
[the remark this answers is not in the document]
lilchurro: Not sure about this one, actually, because it specifically tests the changes to my branch.

clean_cap_dump.c
```diff
@@ -0,0 +1,245 @@
...
+#include <pcap/pcap.h>
+#include <ether.h>
+#include "clean_cap_dump.h"
+#include <extract.h>
+#include "ethertype.h"
+#include "ip.h"
+#include "tcp.h"
+#include "udp.h"
+
+#define MAXPACKET   65549
```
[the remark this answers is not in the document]
lilchurro: Adding comment about it being largest IP packet size + eth header.

lilchurro:
😱  Well, that rebase was a harrowing experience. 1 clean commit now tho.

gregs5:
Is this something @lilchurro should consider spending the time to bring up to date? Would be great to get some guidance here!

infrastation:
There were two discussions made in this pull request. One about the code itself, which indeed belongs here. I cannot help completing it before I finish other bits of work, other developers may be available. Another discussion, which would belong more to the tcpdump-workers mailing list, was whether it is right to incorporate the masking function into tcpdump as opposed to doing it in a separate binary. I guess the latter one could go into more detail to reach a more pronounced consensus. I, for instance, am not quite convinced yet, but I could be wrong.

gregs5:
Thanks - really appreciate the reply. Understood on your concerns. I'll share that we looked at other methods (and all the tools you referenced last year) and still came to the conclusion that for anyone who wants this functionality, this is the right path.  That said though, I'll start a thread in the tcpdump-workers and we'll see where it takes us. Thanks!

Issue #616, opened 2017-07-20
CVE-2017-11108 - DoS in print-stp.c


anarcat:
As part of the Debian LTS audit process, we are investigating the vulnerability in tcpdump identified with CVE-2017-11108:

    tcpdump 4.9.0 allows remote attackers to cause a denial of service (heap-based buffer over-read and application crash) via crafted packet data. The crash occurs in the EXTRACT_16BITS function, called from the stp_print function for the Spanning Tree Protocol.

There was a discussion surrounding the issue in the RedHat bugtrackers, but it is unclear if the original submitter (in CC) was able to reach out to you:

https://bugzilla.redhat.com/show_bug.cgi?id=1468504

I am taking, however, what was said there at face value, however, and therefore opening a discussion surrounding this issue on this bugtracker to ensure followup.

I can confirm I have reproduced the issue in tcpdump 4.9.0 in Debian (stretch and wheezy), but cannot confirm how easily exploitable it could be. The symptoms are easily triggered by running tcpdump against the proof of concept available in the RedHat bugtrackers, under valgrind or when compiling tcpdump with ASAN.

I'm trying to figure out exactly how to fix this, at this stage. It
doesn't look like upstream responded just yet, but I can already say the
following.

 1. tcpdump does try to derefence a pointer with an address value that
    is user-controlled (mstp_len) in print-stp.c:478

 2. even when tcpdump doesn't crash, it fails to render the packet
    properly. with -vvv, i get this:

        STP 802.1aq, Rapid STP, CIST Flags [Learn, Forward, Agreement], length 175
            port-role Designated, CIST root-id 8000.20:bb:c0:7b:d1:2e, CIST ext-pathcost 0
            CIST regional-root-id 8000.20:bb:c0:7b:d1:2e, CIST port-id 8002,
            message-age 0.00s, max-age 20.00s, hello-time 2.00s, forwarding-delay 15.00s
            v3len 8, [|stp 175]

    notice the trailing comma there: after the `v3len` field, there
    would normally be other things there, but some overflow detection
    kicks in stp_print_mstp_bpdu(), which, in turns, calls the magic
    ND_TCHECK macro which detects the overflow and bails out properly

 3. therefore, -vvv doesn't trigger the bug

 4. conclusion: it may be simply a matter of calling the proper
    ND_TCHECK macro before dereferencing the pointer.

I think that covers most of it. Let me know if you need more information.

infrastation:
[...]

PR #617, opened 2017-07-20
Check mstp_len before dereferencing

anarcat:
During STP processing, we have a special handler for `SPB` (Shortest
Path Bridging) which parses a part of the packet for that specific
protocol. To find the length of that part of the packet (`spb_len`),
it uses the `mstp_len` variable to find where that value is
stored. Since `mstp_len` is unpacked from the packet earlier, it is a
user-controlled value and so, the pointer to `sbp_len` can also be
user-controlled.

We therefore need to check that we can extract it safely, using the
`ND_TCHECK_16BITS` macro, as per *Code style and generic remarks* in
the `CONTRIBUTING` document. The pointer taken to extract `mstp_len`
itself *is* checked, but, oddly, even though the comments say
`sbp_len` is validated as well, the macro call is missing.

This fixes the symptoms described by CVE-2017-11108 in my tests, but
someone with e deeper knowledge of the code should review this to
confirm it is the correct solution.

Closes: #616

### DIFF
--- a/print-stp.c
+++ b/print-stp.c
@@ -475,6 +475,7 @@ stp_print(netdissect_options *ndo, const u_char *p, u_int length)
             if (stp_bpdu->protocol_version == STP_PROTO_SPB)
             {
               /* Validate v4 length */
+              ND_TCHECK_16BITS(p + MST_BPDU_VER3_LEN_OFFSET + mstp_len);
               spb_len = EXTRACT_16BITS (p + MST_BPDU_VER3_LEN_OFFSET + mstp_len);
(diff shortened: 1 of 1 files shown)


anarcat:
also note that there may be other branches of the stp code that are vulnerable to similar issues. i haven't audited the whole thing, just the part reported as vulnerable.

infrastation:
Commit d9e65de is a more complete solution to this problem, it is now available in tcpdump release 4.9.1. Thank you very much for the pull request, please send any future security reports to security@tcpdump.org.

PR #621, opened 2017-07-25
Add decoder for XPF RRDATA

arjenz:


infrastation:
The master branch went significantly ahead, this change needs to be rebased and amended to use the current test/extract macros. Otherwise it does not look like it will take forever to proof-read.

arjenz:
draft-bellis-dnsop-xpf does not appear to be gaining a lot of traction, so no point (currently) to add a decoder for it.

PR #624, opened 2017-08-02
cross-compiling: do not check dlpi.

josealar:
For cross-compiling on Linux platforms, we do not need to check libdlpi
since it is only placed on Solaris.

Signed-off-by: Xin Ouyang <Xin.Ouyang@windriver.com>

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -550,7 +550,9 @@ don't.])
 fi
 
 # libdlpi is needed for Solaris 11 and later.
-AC_CHECK_LIB(dlpi, dlpi_walk, LIBS="$LIBS -ldlpi" LDFLAGS="-L/lib $LDFLAGS", ,-L/lib)
+if test "$cross_compiling" != yes; then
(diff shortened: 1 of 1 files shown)


infrastation:
The comment specifies Linux cross-compilation but the change seems to affect any cross-compilation regardless of the OS, could you check and/or explain? Thank you.

guyharris:
1. Why is this a problem?  This doesn't check whether something *runs*, it checks whether something *compiles and links*, which should work for cross-compiling - and, on OSes other than Solaris (and on Solaris prior to Solaris 11), there *is* no libdlpi, so the test will fail (as it should) and not add libdlpi to the list of libraries.
1. As Denis notes, if you're cross compiling for Solaris (e.g., compiling for SPARC on Solaris x86 or for x86 on Solaris SPARC), you *should* check for libdlpi.

josealar:
Thank you for your comments.

This patch is currently applied in meta-oe. I did test building tcpdump without this patch and it builds correctly. I will send a patch to remove it to meta-oe mailing list.

PR #627, opened 2017-08-30
Handling and decoding of NTP mode 6 Control Packets (rebased)

sjvudp:
I created a new branch equivalent (as of now) with pull request 625 that got a conflict after updating master. This feature allows to detect NTP mode 6 packets (and no longer print them as timestamp messages). In addition most parts of the control packet can be decoded (depending on "-v"). What is not done yet is the interpretation of the "data" part of the control packets (which may be split among multiple packets).


infrastation:
For clarity, this is not a result of `git rebase`, this is a result of `git merge` and it joins up to 5 history paths.

sjvudp:
> For clarity, this is not a result of `git rebase`, this is a result of `git  merge` and it joins up to 5 history paths.

I know about the merge. Interestingly when I did a rebase with manual merging, the branch was still maked as conflicting...

PR #628, opened 2017-08-30
Add NTP test

mlichvar:


sjvudp:
I'd advise the pull requestor strongly to describe the kind of tests added (here, too). Referring to "No description provided"...

mlichvar:
There are several client/server exchanges, using different authentication options: no key, MD5 key and SHA1 key. One response is a crypto-NAK. The commit includes the expected output.

infrastation:
For the avoidance of doubt: is everybody happy for this commit to be merged as it is now?

sjvudp:
Before committing: Maybe can we discuss what type of packets we'd like to see in the pcap file, and what options we want to run tcpdump with?
For time packets, the variants would be different versions of NTP, different modes of operation, authenticated or not, different MAC types, autokey and extension fields.
For control packets I'd like to see single-packet responses, multi-packet responses, authenticated and unauthenticated, and a few different commands.
Maybe we want a mode7 (reserved) packet also.

infrastation:
It helps to have as many different encoding examples as possible to test as many code paths as possible. Also if the code goes into different level of details depending on `ndo_vflag` (as it does or is going to do soon in `print-ntp.c`), there should be a test for each level of detail for each capture. Then much later when somebody changes something and the output becomes different, there is an opportunity to stop and think if that is the intended effect before the change is copied around. If some captures to some extent repeat some others that's OK as protocols mix all the time.

sjvudp:
I agree with the comments, but isn't the hard part to add single packets to an existing pcap file?
The other thing is: Should there be one pcap file for one protocol, or should there be different pcap file, one for each aspect of the protocol? Here NTP time messages and NTP Control messages use quite a different format, and they have different meanings. This is reflected in the code by using different decoding functions. Ideally hacking one should not affect the other, and maybe the tests should reflect that code structure.

infrastation:
Multiple captures per protocol work just fine, please see the already existing tests. As a minimum sufficient requirement, if you add a new test it must cover something that was not covered before. It is nice to see a test that is 100% pure and complete but that is not a requirement, otherwise there would be way fewer tests in the set.

infrastation:
If anybody thinks this change should not be merged please state exactly what improvements need to be made, otherwise this will be in the master branch soon because more samples mean less bugs in the code.

sjvudp:
Well, if you are ready to accept changes later: I still wonder: Should there be one pcap file per protocol, or should there be multiple pcap files for individual features of a protocol. Changes to pcap files don't diff very nicely as they are binary.

infrastation:
Thanks for the contribution.

PR #630, opened 2017-09-07
tcpdump should decode BGP add-path NLRI and withrawn routes correctly

davidcronin-arista:
Added ability to decode bgp add-path for updated and withdrawn routes for ipv4 and
ipv6. Since there is no flag to signal that a message contains add-path content, a
heuristic approach is taken where we assume an add-path format if an add-path
message is sensible and that standard bgp is not. In this way we will only display
as add-path if displaying as regular bgp was in some way incorrect anyway.
Also modified bgp_update_print to correctly decode withdrawn routes as IPv4 as IPv6 
routes are always in MP_UNREACH_NLRI.

The test covers decoding of add path in the NLRI, IPv4 and IPv6 MP_REACH and MP_UNREACH, and withdrawn routes which were not decoded before.

Build / test results ( all pass ) can be seen here https://travis-ci.org/davidcronin-arista/tcpdump


davidcronin-arista:
I should probably point out that the test covers decoding of add path in the NLRI, IPv4 and IPv6 MP_REACH and MP_UNREACH, and withdrawn routes which were not decoded before.

infrastation:
Existing comments, Section 6 of RFC 7911 and commit c9c3d3c may provide some additional context to the reviewer of this change. Also the two commits were probably originally meant to be one.

infrastation:
Could you rebase this work on the current master branch and make it one commit?

davidcronin-arista:
Apologies, I'm not that familiar with git.

It should now be one commit and be rebased against the latest master. 

However I was getting a rake of test failures this time when I ran make check which I didn't see before. The test I added is now failing too although I can't see what the difference is. It's possible this is just a formatting difference? I don't have the time to look into it further now, and if it is a formatting difference I don't have time to go through all the outputs and regenerate them if that's what is needed.

infrastation:
Thank you. Here is a summary of failing tests:
* bgp-as-path-oobr-ssl (now goes into more detail than before)
* bgp-addpath (tabs/spaces indent)
* tok2str-oobr-1 (detects invalid attribute)
* bgp_pmsi_tunnel-oobr (idem)
* bgp_mvpn_6_and_7
* bgp_vpn_rt-oobr

Also there is a warning:
```
./print-bgp.c: In function ‘bgp_update_print’:
./print-bgp.c:2619:17: warning: ISO C90 forbids mixed declarations and code [-Wpedantic]
                 int add_path = check_add_path(p, withdrawn_routes_len, 32);
                 ^
```

A more substantial update on this change remains to be done.

DaveCronin:
Hi, I got around to doing some work on this:
Fixed the warning you mentioned
Corrected the tabbing/indentation for bgp-addpath test
Added the extra detail that was being output for bgp-as-path-oobr-ssl && bgp-as-path-oobr-nossl
(note, bgp-as-path-oobr-nossl runs when I run make check, whereas you pointed it out for bgp-as-path-oobr-ssl from the travis log) 

For the other four tests that are failing I think my output is incorrect as opposed to the tests needing updating. I will look into these further.

Please let me know if you spot any mistakes
Thanks

DaveCronin:
Hi, I've had a look at the last four tests and I believe I have fixed the issues, now all tests are passing.

Let me know what you think

PR #631, opened 2017-09-08
Parse NTP mode 6 messages, detect mode 7 messages

sjvudp:
This is a rebased branch corresponding to `sjvudp:f-NTP-mode6` (pull request 625). This will obsolete pull request 625 (but maybe not its comments).


sjvudp:
I added/merged some minor fixes, most significant is printing of the message digests.

infrastation:
Okay, this is step 1. Your work in progress has been rebased on the current master branch and force-pushed back into the `i2-NTP-mode6` branch in your forked repository.

infrastation:
The rebase has prepended your work with 150 commits and it involved only one conflict to resolve in `tests/TESTLIST`. If you study those 150 commits you will understand better why it is important to check both the declared length from packet headers and the input buffer boundary.

sjvudp:
I did a local sync with upstream before adding another feature to do a basic dump of NTP mode 7 packets. Well, it should have been a separate pull request, but as it's based on the root of this branch (that is a good deal outdated regarding NTP packets), I just merged it in manually (I aborted rebase because of too many conflicts). Maybe yu like it and "forward". I'm aware that my newly added NTP mode 7 checks may conflict with TESTLIST.

infrastation:
Please let me know when you are done adding new commits and will get back to step 1. Thank you.

sjvudp:
At the moment I have no intentions to add more stuff to this branch, unless someone thinks I should fix "this and that". I'm working on other NTP stuff, but that addresses a new topic, and thus I'll create a new branch. However, I'll have to wait for integration (to pull) as the additional work is based on this branch.

infrastation:
Before we proceed further please review the changes made in commit 4dafda3. Some of your code that is already in the master branch prints client/server, length and leap indicator one more time in addition to what was already there. Is this a bug?

fxlb:
Problem on Solaris 10 - 32 bits (i386/SPARC):
https://buildfarm.opencsw.org/buildbot/builders/tcpdump-solaris10-i386
https://buildfarm.opencsw.org/buildbot/builders/tcpdump-solaris10-sparc

sjvudp:
> Some of your code that is already in the master branch prints client/server, 
> length and leap indicator one more time in addition to what was already 
> there. Is this a bug?

It would be helpful if you'd refer to a line in the test output files, preferrably. I've checked with my working version ,and could not find what you are referring to, so I guess it's a temporary (d)effect. Did you check the state for the last commit in the branch?

Examples I found (some lines removed for brevity):
IP6 ::1.123 > ::1.38531: NTP LI=0, VN=2, Mode=6, length=408

IP6 (flowlabel 0x62d8a, hlim 64, next-header UDP (17) payload length: 20) ::1.38531 > ::1.123: [bad udp cksum 0x0027 -> 0x5280!] NTP leap indicator=0 (Nominal), Version=2, Mode=6 (Control Message), length=12

IP 10.1.1.4.123 > 132.199.176.18.123: NTP LI=3, VN=3, Mode=1, length=48

IP (tos 0x0, ttl 127, id 44377, offset 0, flags [none], proto UDP (17), length 76)
    10.1.1.4.123 > 132.199.176.18.123: NTP leap indicator=3 (clock unsync.), Version=3, Mode=1 (symmetric active), length=48

IP (tos 0x0, ttl 127, id 44377, offset 0, flags [none], proto UDP (17), length 76)
    10.1.1.4.123 > 132.199.176.18.123: [udp sum ok] NTP leap indicator=3 (clock unsync.), Version=3, Mode=1 (symmetric active), length=48

infrastation:
The specific line in the test output is highlighted in commit 4dafda3.

sjvudp:
On the problem with commit 4dafda3:
It seems some automatic merge failed, as the offending lines are both old and from the same commit:
90692fcf9 print-ntp.c (Denis Ovsienko          2014-03-20 14:39:44 +0400 267) 	ND_PRINT((ndo, "NTPv%d", version));

infrastation:
Let me tell once again that commit 4dafda3 does not cause the problem, it just makes it visible.

The NTP decoder used to output client/server, length and leap indicator even before the changes made in your commit c2e2a18, which added another copy of client/server, length and leap indicator (for example, line 488 of `print-ntp.c` in the current master branch). That addition was one of the reasons why the NTP test added in commit fdd0cf4 failed (it output was produced before your changes). Commit 4dafda3 fixed that failure but made it apparent there is the duplication now.

The question is: is that duplication intentional or is that a bug?

fxlb:
Same problem on Debain Jessie 32 bits:

Failed test: ntp

33c33
<         Receive Timestamp:    1802554105.693999877 (2093-03-22T03:56:41)
---
>         Receive Timestamp:    1802554105.693999877 (1957-02-13T21:28:25)

guyharris:
> Problem on ... 32 bit...

The file contains a time stamp that's past the Y2.038K point, so attempts to print it will cause an overflow, making it look like a negative time from 1957 rather than a positive time from 2093.

The only fixes would be to 1) have our own 64-bit version of `time_t`, and our own versions of `gmtime()`, `localtime()`, and `strftime()` that can always handle 64-bit time stamps or 2) suppress the "ntp" test if tcpdump is a 32-bit executable.

I'm working on 2).

infrastation:
The 2093/1957 issue reproduces exactly on the following systems (`uname -srp`):
* SunOS 5.10 i386
* FreeBSD 11.1-RELEASE i386
* Linux 4.4.0-96-generic i686

So it seems the NTP tests may fail and should be skipped if `uname -p` prints either i386 or i686.

guyharris:
> So it seems the NTP tests may fail and should be skipped if uname -p prints either i386 or i686.

Or if we have a 32-bit non-x86 binary.  I run some tests on The Written Word's machine, which include SPARC, PA-RISC, and Power ISA machines, and I build both libpcap and tcpdump 32-bit and 64-bit and test both of them, so I'm doing the 32-bit vs. 64-bit tests by running `file` on the tcpdump binary.

sjvudp:
Looking at "`git diff c2e2a1^1 c2e2a1`" it seems the bug was introduced by the request to provide a minimal first commit. The original branch (`git diff e065a5b7850c2c1^1 e065a5b7850c2c1`) had this:
```
+/*
+ * Print NTP time requests and responses
+ */
+static void
+ntp_time_print(netdissect_options *ndo,
+              register const struct ntp_time_data *td, u_int length)
+{
+       ND_TCHECK(td->stratum);
        ND_PRINT((ndo, ", Stratum %u (%s)",
-               bp->stratum,
-               tok2str(ntp_stratum_values, (bp->stratum >=2 && bp->stratum<=15
) ? "secondary reference" : "reserved", bp->stratum)));
+               td->stratum,
+               tok2str(ntp_stratum_values,
+                       (td->stratum >= 2 && td->stratum <= 15) ?
+                       "secondary reference" : "reserved", td->stratum)));
```
So yes, the patch is buggy.

sjvudp:
> Same problem on Debain Jessie 32 bits:
> 
> Failed test: ntp
> 
> 33c33
> <         Receive Timestamp:    1802554105.693999877 (2093-03-22T03:56:41)
> ---
>>         Receive Timestamp:    1802554105.693999877 (1957-02-13T21:28:25)
> 

My guess is signed vs. unsigned interpretation of the time value. The distance from 1970 looks rather equal ;-)

infrastation:
Clearly it does not make sense to print the packet length 2nd time inside `ntp_time_print()`. As far as I my basic understanding of NTP goes the same applies to both mode (client/server) and leap indicator. I am going to make respective fixes to the code in `print-ntp.c` within about an hour.

infrastation:
Alright, commit 4d42441 removes the duplicate output and updates the test output. @sjvudp, please try to organize your commits the same way.

infrastation:
For the avoidance of doubt, I didn't touch anything related to the 32-bit `time_t` effect as @guyharris has a more systematic solution in mind.

guyharris:
> My guess is signed vs. unsigned interpretation of the time value.

`time_t` is signed in most UN\*Xes, which means that, if it's 32-bit, it cannot represent times past 2038-01-19 03:14:07 UTC.  If it were unsigned, the limit would be 2106-02-07 06:28:15 UTC.

If you take an unsigned second-since-the-Epoch value for a date past 2038-01-19 03:14:07 UTC but not past 2106-02-07 06:28:15 UTC, and put it into a 32-bit signed `time_t`, the result will have the uppermost bit set and will thus look like it's negative.

That's what's happening here.

What we probably *should* do is either:

1. on platforms with a 32-bit `time_t`, treat time values more than 2^31-1 as unrepresentable and print them as such

1. have our own `gmtime()` (and our own `localtime()`, e.g. using the IANA time zone code?) that handles 64-bit time stamp values, and use those for all packet time stamps.

infrastation:
As to the original purpose of this branch, I have just tried rebasing it on the current master branch again -- after you have added a merge commit for another series of commits based on one month old repository -- and I could not make it through the conflicts. So we are back to step 0 here.

Out of this large set of commits I will cherry-pick the .pcap files to add more tests -- by now it should be reasonably clear tests are useful. Then I will ask you once again to stop buffering new work, to stop branching and to stop merging (in other words, to stop complicating the matter) and then to prepare **one** first complete change that does **one** useful thing and updates the tests in the same commit. That change will have to be based on the current master branch.

If you can do that I can offer help reviewing that complete proposed change. If you cannot do that I cannot make it through so many commits, I am sorry but the amount of time I can spend on this is limited. As an additional way to improve this situation, please consider learning how to use `git rebase` and how to resolve rebase conflicts.

guyharris:
> What we probably should do is either:

3. Treat the value from the packet as unsigned, convert it to 64-bit signed, subtract the 1900-to-1970 offset from it, stuff the result into a `time_t`, fail if that changes the value, otherwise hand that `time_t` to `gmtime()` and print that.

This seems to work; the time stamp in question is from 1957(!), and, with the change, it shows up as a 1957 date/time on both 32-bit and 64-bit platforms.

infrastation:
Now that you have explained it I realize it has to do with the conversions in Section 6 of RFC 5905 (much of which stems from the choice of 1900 as the epoch, but anyway).

infrastation:
@sjvudp, your 3 packet captures from this branch are now in the master branch, I am looking forward to reduce the amount of unmerged work as described above. Cheers!

sjvudp:
On the "one useful thing": I had made a local feature branch initially and then merged that "non-ff" locally, the committed. So You had the choice (assuming github has not problem with that) between the small commits that lead up to the result, or the one big diff. Unfortunately, and here starts the problem, you were picky about my renaming of `bp` to `td` when the underlying data type also had changed (the name of the function, too, BTW). You insisted on having the old name. That causes most of the merge failures for me. I'll retry...

sjvudp:
I just pushed `i3-NTP-mode6`, a rebase of `i2-NTP-mode6` originating at commit `42208e6a24`.

infrastation:
Thank you, that's great news; I can make a look at the new branch later if nobody else is willing.

sjvudp:
Just in case you want to cherry-pick: If possibly do it chronologically, so another rebase will be easier for me at least.

sjvudp:
For the curious: I just pushed (no pull request yet) branch `i-NTP-EF` (based on the top of this branch) that deals with parsing **NTP Extension Fields**. _I could need some example packets_ for further development and testing... Unfortunately I won't have time the next two weeks, so don't hurry ;-)

sjvudp:
I just pushed another rebase (`i4-NTP-mode6`) based on commit `4943adfb5f2768c06ce08bcf402ec6d9dd226b3c`.

sjvudp:
See pull request  657 (another rebase with some adjustments).

PR #643, opened 2017-11-12
An alternative check for alarm()

sgeto:
MinGW's alarm() implementation is a [stub](https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-crt/misc/alarm.c) and its [declaration is guarded](https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-headers/crt/io.h#L358).

The current check does not verify whether alarm() is usable.
This one doesn't either, but at least it checks whether it is declared in the default include files, which I think is, at least in this case, a more reliable test.

https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Generic-Declarations.html

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -3,9 +3,6 @@
 /* define if you want to build the possibly-buggy SMB printer */
 #undef ENABLE_SMB
 
-/* Define to 1 if you have the `alarm' function. */
-#undef HAVE_ALARM
-
 /* Define to 1 if you have the `bpf_dump' function. */
 #undef HAVE_BPF_DUMP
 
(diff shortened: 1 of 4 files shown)


guyharris:
Is

    #ifndef _WIN32

an even *better* check for `alarm()`?  If it's Windows, we presumably have the multimedia timer stuff, and don't need `alarm()`; if it's not Windows, it's UN\*X (if somebody wants us to support something that's neither of them, they'll have to contribute support that checks for it), and it has `alarm()` (if not, then "1978 called, they want their UNIX back").

guyharris:
> If it's Windows, we presumably have the multimedia timer stuff, and don't need `alarm()`; if it's not Windows, it's UN\*X (if somebody wants us to support something that's neither of them, they'll have to contribute support that checks for it), and it has `alarm()` (if not, then "1978 called, they want their UNIX back").

Done in 20af93776dc9c23b9b1ef53ee7260941231c32d1.

PR #644, opened 2017-11-12
Use WDversion only when building windump

sgeto:
...with the project files in tcpdump\win32\prj\.
This is achieved by checking whether WPCAP is true, which is only defined in tcpdump\win32\prj\WinDump.vcproj.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -2794,7 +2794,7 @@ print_packet(u_char *user, const struct pcap_pkthdr *h, const u_char *sp)
 		info(0);
 }
 
-#ifdef _WIN32
+#if defined(_WIN32) && defined(WPCAP)
 	/*
 	 * XXX - there should really be libpcap calls to get the version
 	 * number as a string (the string would be generated from #defines


guyharris:
If we have a sufficiently recent libpcap, we should use `pcap_lib_version()` to get the libpcap version, whether we're on Windows or UN\*X.  WinPcap 4.1.2 has `pcap_lib_version()`, according to [its list of exported functions](https://www.winpcap.org/docs/docs_412/html/group__wpcapfunc.html), and that may be true of at least some earlier versions.

tcpdump doesn't use `WDversion[]` *at all*, so there's no reason to define it; the same applies to `version[]` and `Wpcap_version[]`.

For `pcap_version[]`, we don't need it if we have `pcap_lib_version()`, and we don't have it if `HAVE_PCAP_VERSION` isn't defined, so we should ignore it if `HAVE_PCAP_LIB_VERSION` is defined, and use it otherwise, and if `HAVE_PCAP_VERSION` isn't defined, define it ourselves, as `"unknown"`.

guyharris:
> tcpdump doesn't use `WDversion[]` *at all*, so there's no reason to define it; the same applies to `version[]` and `Wpcap_version[]`.
> 
> For `pcap_version[]`, we don't need it if we have `pcap_lib_version()`, and we don't have it if `HAVE_PCAP_VERSION` isn't defined, so we should ignore it if `HAVE_PCAP_LIB_VERSION` is defined, and use it otherwise, and if `HAVE_PCAP_VERSION` isn't defined, define it ourselves, as `"unknown"`.

Done in c8ccf8b3e18392e857fd22517752777a022949f3.

PR #647, opened 2017-11-14
Some tweaks for AC_LBL_LIBPCAP

sgeto:
I noticed that the "additional-libs" block in `AC_LBL_LIBPCAP` in never reached.
This "fix" should therefore help in adding any additional libs pcap depends on/was linked with to, `LIBS` (including ws2_32 on windows).

Heads up: I'm not sure if I indented correctly. I don't have syntax highlighting for m4. Sorry in advance.
I also only tried it on windows so let's see what travis has to say to this.

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -584,18 +584,20 @@ AC_DEFUN(AC_LBL_LIBPCAP,
                        pcapH=$dir
                    fi
                 done
-            fi
+        fi
 
             if test $pcapH = FAIL ; then
                     AC_MSG_ERROR(cannot find pcap.h: see INSTALL)
- 	    fi
+            fi
             $2="-I$pcapH $$2"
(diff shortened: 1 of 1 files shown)


guyharris:
> I'm not sure if I indented correctly.

The indentation of `AC_LBL_LIBPCAP` was already a complete mess before your changes, so there's no need to worry about correctness.  I made that macro use 4-space indentation without tabs in 55e0de4ea1456a4c2ef4f5528fec2f1a9f2272f3.

sgeto:
I forgot to push the updated configure script. You need to regenerate.

guyharris:
> I forgot to push the updated configure script. You need to regenerate.

I waited until after I re-indented: 768d80e11f6afdcc8ed088d0723507140a7a633e.

PR #651, opened 2017-11-18
Lowered the minimum length of second field in mkdep's awk command

sgeto:
The awk command that is run last as part of `mkdep` to remove duplicate lines doesn't work on my machine. It removes duplicate lines but leaves their backslashes \

This is an estimation of how I think this can be fixes/improved.
I think that the length of this field can vary so much from OS to OS, that the value set now is as arbitrary as the old one. Though a lower value may be a better choice.

Another way is to add a newline if the minimum length isn't met like so `else rec = rec " \n" $2`

### DIFF
--- a/mkdep
+++ b/mkdep
@@ -112,7 +112,7 @@ awk '{
 		prev = $1;
 	}
 	else {
-		if (length(rec $2) > 78) {
+		if (length(rec $2) > 58) {
 			print rec;
 			rec = $0;
 		}


sgeto:
Boi, what a long title ...

fxlb:
@sgeto Do you always have the problem on your machine?

infrastation:
@sgeto, could you please explain in detail how the problem reproduces?

infrastation:
Bug confirmed.

As far as I understand the awk code in `mkdep`, its purpose is to compress Makefile target dependencies as follows:
```
a.c: b.h
a.c: c.h
```
becomes
```
a.c: b.h c.h
```
When the same target has many dependencies, this may result in more than one compressed output line, each no longer than 78 characters. Other versions of this script from 1988 and 1993 use the same line length limit. In theory, reducing the maximum length _should_ have only a cosmetic effect. In practice, the awk script indeed does not handle certain valid input correctly:
```
a.c: b.h \
 c.h \
 c.h
```
becomes
```
a.c: b.h \
 c.h \ 
```
In this case making the limit small enough short-circuits the logic and the output becomes less compressed, but valid. Reducing the limit would not be a valid solution, as the short-circuit depends on the length of the particular filename that triggers the bug. A proper fix would be fixing the logic of rule compression (or discarding the compression, as some other versions of `mkdep` have).

infrastation:
Commit 8238d6c removes the awk code in tcpdump, let's give this change some time to age. If that works out as expected, it will be very likely that the same change works well in libpcap and tcpslice too because these have fewer source files.

infrastation:
Populated to tcpslice and libpcap, closing as resolved. Thank you for reporting this problem.

PR #653, opened 2017-11-19
Made tests suite work with tcpdump.exe and msys build of Perl

sgeto:
these 15 tests fail:
kday1-8
cve2015-0261_01
pcap-invalid-version-1
pcap-invalid-version-2
pcap-ng-invalid-vers-1
pcap-ng-invalid-vers-2
l2tp-avp-overflow
pktap-heap-overflow

How to interpret their exit code?
A log is attached.
[test_results.txt](https://github.com/the-tcpdump-group/tcpdump/files/1485189/test_results.txt)

### DIFF
--- a/tests/TESTonce
+++ b/tests/TESTonce
@@ -18,6 +18,10 @@ if ($^O eq 'MSWin32') {
     $r = system "..\\windump -n -t -r $input $options 2>NUL | sed 's/\\r//' | tee NEW/$output | diff $output - >DIFF/$output.diff";
     # need to do same as below for Cygwin.
 }
+elsif ($^O eq 'msys') {
+    $r = system "../tcpdump 2>/dev/null -n -t -r $input $options >NEW/$output";
+    #print sprintf("END: %08x\n", $r);
+}
 else {
     # we used to do this as a nice pipeline, but the problem is that $r fails to
     # to be set properly if the tcpdump core dumps.


guyharris:
Needs to have the conflicts resolved.

infrastation:
Although the suggested change is very small, it does not rebase well because the Perl code has changed too much. However, I managed to make some sense of this as follows.

The attached file shows that each of the 15 failed tests had the same reason:
```text
< EXIT CODE 00000100
```
Tcpdump does not exit with status code 256 directly. However, if the Windows-specific comment in `TESTrun` applies, in this case 256 means status code 1 shifted 8 bits left, which would be `S_ERR_HOST_PROGRAM`. The only ways to exit with this status code would be (in the current edition of `tcpdump.c`, that is) through:
* `error()`
* the switch block after `getopt_long()`
* `compress_savefile()`

The only possible route during `make check` is `error()`, which also implies an error message, which, unfortunately, remains unknown because in that revision of the script both Windows and Unix branches silenced stderr output. In the hindsight, the cause likely was one of the several improvements to input file validation.

From the proposed change it is clear that the operating system name is "msys" rather than "MSWin32", and that the main effect of the change is skipping the exit code check (in that revision of the script, same as it did for "MSWin32").

So it looks like "msys" allows to run the binary using POSIX command line, but uses Windows encoding for the exit status. If that's correct, the following change should address the problem in a better way:
```diff
--- a/tests/TESTrun
+++ b/tests/TESTrun
@@ -121,7 +121,7 @@ sub runtest {
             $status = $!;
             printf OUTPUT "FAILED TO RUN: status: %d\n", $status;
         } else {
-            if ($^O eq 'MSWin32') {
+            if ($^O eq 'MSWin32' or $^O eq 'msys') {
                 #
                 # On Windows, the return value of system is the lower 8
                 # bits of the exit status of the process, shifted left
```
That said, I don't have any means to test it, and the underlying problem may still stand, in which case the script should report it in a better way.

Unless anyone knows better, I am going to make the change above in 7 days to resolve this case as much as reasonably practicable.

infrastation:
The issue has been addressed as discussed above instead of the originally proposed changes. Closing.

PR #659, opened 2017-12-30
Add support for IPv6 routing header type 4

amsalam20:
IPv6 routing header type 4 is known as Segment Routing Header (SRH).
It is used by Segment Routing (SR) to steer a packet through a
controlled set of instructions, called segments. It is described in
https://tools.ietf.org/html/draft-ietf-6man-segment-routing-header-07

An extension is added to allow tcpdump to print SRH information.


fxlb:
Thanks, before a review, please add a pcap file with this new header in the tests/ directory.

PR #661, opened 2018-01-22
Fix printing of wildcard retractions and IHU in Babel printer.

jech:
When the AE field is 0, the message applies to any routes or
neighbours.  Tcpdump used to print it as ::/0, which is confusing.

### DIFF
--- a/print-babel.c
+++ b/print-babel.c
@@ -439,16 +439,19 @@ babel_print_v2(netdissect_options *ndo,
                 ND_PRINT(" ihu");
             else {
                 u_char address[16];
+                u_char ae;
                 int rc;
                 ND_PRINT("\n\tIHU ");
                 if(len < 6) goto invalid;
                 txcost = EXTRACT_BE_U_2(message + 4);
                 interval = EXTRACT_BE_U_2(message + 6);
-                rc = network_address(EXTRACT_U_1(message + 2), message + 8,
+                ae = EXTRACT_U_1(message + 2);
+                rc = network_address(ae, message + 8,
                                      len - 6, address);
                 if(rc < 0) { ND_PRINT("%s", tstr); break; }
                 ND_PRINT("%s txcost %u interval %s",
-                       format_address(ndo, address), txcost, format_interval(interval));
+                       ae == 0 ? "any" : format_address(ndo, address),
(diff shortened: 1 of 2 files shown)


print-babel.c
```diff
@@ -556,17 +561,19 @@ babel_print_v2(netdissect_options *ndo,
...
-                rc = network_prefix(EXTRACT_U_1(message + 2),
+                rc = network_prefix(ae,
                                     EXTRACT_U_1(message + 3), 0,
                                     message + 16, NULL, len - 14, prefix);
                 if(rc < 0) goto invalid;
                 plen = EXTRACT_U_1(message + 3) + (EXTRACT_U_1(message + 2) == 1 ? 96 : 0);
                 ND_PRINT("(%u hops) for %s seqno %u id %s",
-                       EXTRACT_U_1(message + 6), format_prefix(ndo, prefix, plen),
+                       EXTRACT_U_1(message + 6),
+                       ae == 0 ? "invalid" : format_prefix(ndo, prefix, plen),
```
infrastation: Would it be more helpful to print something like "invalid AE 0"? Also `MESSAGE_NH` likely needs something similar. I understand the fixup is not required for `MESSAGE_UPDATE_SRC_SPECIFIC`, `MESSAGE_REQUEST_SRC_SPECIFIC` and `MESSAGE_MH_REQUEST_SRC_SPECIFIC` because those TLV disappear after draft-boutier-babel-source-specific-01 (May 2015) and the whole case blocks are now obsolete.

infrastation:
Would you also like to rename MH-Request to Seqno Request as it is in the spec and fix IHU to say "rxcost" as it is in the spec (currently says "txcost")?

jech:
> Would it be more helpful to print something like "invalid AE 0"?

Feel free.  Anything that won't confuse the reader is fine with me.

> Also MESSAGE_NH likely needs something similar.

Yes, you're right.

jech:
> Would you also like to rename MH-Request to Seqno Request as it is in the spec
> and fix IHU to say "rxcost" as it is in the spec (currently says "txcost")?

Yes please (for both).

jech:
Looks good, thanks.

PR #663, opened 2018-03-04
PTP: Print parts of the common PTP header

lunn:
Precession Time Protocol allows network devices to synchronize their
clocks. Each PTP packet has a common header, and then message type
specific extensions. Print out some of the important fields in the
common header.

PTP can be carried directly in Ethernet frames, or UDP
encapsulated. Add support for both.

Signed-off-by: Andrew Lunn <andrew@lunn.ch>


lunn:
It would be nice to add to CONTRIBUTING a list of 'Makefile' like files where new files need to be added in order that they get built in different build systems...

print-ptp.c
```diff
@@ -0,0 +1,97 @@
...
+	uint8_t messageType;
+	uint16_t flags;
+	uint16_t seq;
+	uint64_t clockId;
+	uint16_t srcId;
+
+	if (length < PTP_COMMON_HEADER_LEN)
+		goto trunc;
+
+	messageType = 	EXTRACT_U_1(&bp[0]) & 0xf;
```
guyharris: There should be ND_TCHECK_n() calls before fetching; `length` is the on-the-wire length, not the captured length.

lunn: Done

lunn: Trivial merge conflict resolved

lunn:
Anything i can do to help get this merged?

infrastation:
This decoder should take into account the `versionPTP` field (4 lower bits of octet 1 in the header) and only decode the version it implements (from v2 packet diagrams I infer this code stands for v1). Other than that, it looks very straightforward.

infrastation:
To improve `CONTRIBUTING` as you suggested above, feel free to suggest particular changes through a pull request.

fxlb:
PR #721 is also about PTP.

mcr:
> PR #721 is also about PTP.

I think #721 is probably a better base, do you agree?  Would the two posters like to collaborate?

fxlb:
The implementation was done via PR #721, closing.

PR #667, opened 2018-03-19
Fix fail to capture on RDMA device on FreeBSD

slavashw:
In pcap for RDMA devices, on FreeBSD where we have support for capsicum
but we don't have file descriptor for a live capture we will fail with:
"tcpdump: unable to limit pcap descriptor"
We just fail to set the proper access rights - don't fail the capture.

Signed-off-by: Slava Shwartsman <slavash@mellanox.com>

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -2089,7 +2089,7 @@ DIAG_ON_CLANG(assign-enum)
 	if (pcap_setfilter(pd, &fcode) < 0)
 		error("%s", pcap_geterr(pd));
 #ifdef HAVE_CAPSICUM
-	if (RFileName == NULL && VFileName == NULL) {
+	if (RFileName == NULL && VFileName == NULL && pcap_fileno(pd) != -1) {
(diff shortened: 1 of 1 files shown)


tcpdump.c
```diff
@@ -2089,7 +2089,7 @@ DIAG_ON_CLANG(assign-enum)
 	if (pcap_setfilter(pd, &fcode) < 0)
 		error("%s", pcap_geterr(pd));
 #ifdef HAVE_CAPSICUM
-	if (RFileName == NULL && VFileName == NULL) {
+	if (RFileName == NULL && VFileName == NULL && pcap_fileno(pd) > 0) {
```
guyharris: That should be `pcap_fileno(pd) != -1`, as `pcap_fileno()` returns -1 on error.  It's *unlikely* that the `pcap_t` file descriptor ends up being 0, and that *shouldn't* happen (programs *should* have the standard input, output, and error open, even if they're all directed to `/dev/null`), but we shouldn't assume it *doesn't* happen.

slavashw:
Can I get some more reviews here please?

PR #669, opened 2018-03-27
Add support for HMAC based on sha256 (128b keys).

kjopek:
Add support for HMAC based on SHA256 with 128 bits keys.

It can be used in the following form:
./tcpdump -nr ~/ipsec_dump.pcap -s0 -E "0xXXXXXXXX@A.B.C.D aes256-cbc-hmac-sha256-128:[KEY1],0xXXXXXXXX@E.F.G.H aes256-cbc-hmac-sha256-128:[KEY2]"

Sponsored by: Digital Fingerprint

### DIFF
--- a/print-esp.c
+++ b/print-esp.c
@@ -332,6 +332,20 @@ int espprint_decode_hex(netdissect_options *ndo,
 	return i;
 }
 
+static int
+strendswith(char *str, char *end)
+{
+	char *p;
+
+	if (strlen(str) > strlen(end) &&
+	    !strcmp(str + strlen(str) - strlen(end), end)) {
+		p = strstr(str, end);
+		*p = '\0';
+		return 1;
+	}
+	return 0;
+}
+
 /*
  * decode the form:    SPINUM@IP <tab> ALGONAME:0xsecret
  */
@@ -344,7 +358,7 @@ espprint_decode_encalgo(netdissect_options *ndo,
 	size_t i;
 	const EVP_CIPHER *evp;
 	int authlen = 0;
-	char *colon, *p;
+	char *colon;
 
 	colon = strchr(decode, ':');
 	if (colon == NULL) {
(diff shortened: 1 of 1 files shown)


print-esp.c
```diff
@@ -332,6 +332,20 @@ int espprint_decode_hex(netdissect_options *ndo,
 	return i;
 }
 
+static int
+strendswith(char *str, char *end)
+{
+	char *p;
+
+	if (strlen(str) > strlen(end) &&
+	    !strcmp(str + strlen(str) - strlen(end), end)) {
```
mcr, line 341: perhaps a comment or two here explaining the logic.   A temp variable or two for strlen(XX) would also increase readability.

print-esp.c
```diff
@@ -353,18 +367,13 @@ espprint_decode_encalgo(netdissect_options *ndo,
 	}
 	*colon = '\0';
 
-	if (strlen(decode) > strlen("-hmac96") &&
-	    !strcmp(decode + strlen(decode) - strlen("-hmac96"),
-		    "-hmac96")) {
-		p = strstr(decode, "-hmac96");
-		*p = '\0';
+	if (strendswith(decode, "-hmac96"))
```
mcr, line 370: I see, you inherited this convoluted strlen(). Did I write that 20 years ago?  Maybe.  Thanks for improving upon this.

mcr (changes requested):
Good changes, but it needs some test cases to exercise sha256.
Even if you haven't got any ESP packets, at least, you can instantiate the hash and it can fail to authenticate the packets in question.

fxlb:
And please, rebase on top of the-tcpdump-group:master.

print-esp.c
kjopek: Sure, thanks!

infrastation:
In addition to the previous comment, which still stands (merging from 3 years ago is not rebasing), you might want to read the file `CONTRIBUTING.md` (especially step 4), so you have a compiler feedback loop in your working copy.

kjopek:
Thank you very much for all your comments and I am very sorry for breaking contributors' rules.

From what I see now, after 3+ years, this patch was wrong from the very beginning and here is my explanation why I think so.

The original patch was kind-of quick and therefore dirty hack to allow IPSEC debugging under certain configuration. Particularly, to make sure the encryption works as expected and data is correctly encrypted/decrypted. In my humble opinion, the pull request should have been splitted into at least two smaller ones:

1. First one to introduce `strendswith` function and use it where needed.
2. Second one to introduce additional primitive for packet authentication.

While the first one is rather obvious, the second one is tricky as the whole function relies on `authlen` variable which is used only for keeping length of hmac code and is required only to calculate correct offsets in the packet to **decrypt** it correctly. I see no other usage of authentication-related code in `print-esp.c` file. `sa_list` structure maintains `authsecret` and `authsecret_len` fields which are unpopulated and therefore does not play any role in printing ESP packets. In my humble opinion, tcpdump lacks hmac verification and this functionality could be added separately.

Summing up, my original intention was to enable ESP decryption for different ENC+AUTH scenario, but from my current perspective this pull request should be closed and I should create another one only to implement `strendswith` function. The remaining part (lack of auth support) should be discussed and added separately. I am leaving this PR open to allow further discussion.

infrastation:
Thank you for the detailed comment. If you think it would make more sense to split the change into two commits, split it. If the first commit is useful on its own, it can be merged on its own. Would you like to take more time to state the decryption problem properly?

ZsBT:
+1 for this (and keeping the issue open)

PR #670, opened 2018-04-07
When SIGINFO (or SIGUSR1) signal sent, also flush the buffer

GuyLewin:
Currently, if you want to download .pcap file while sniffing you'd have to either use -u or kill the process.
I want to use the existing SIGINFO handler to also flush the buffer.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -255,6 +255,7 @@ static const struct tok status_flags[] = {
 };
 #endif
 
+static pcap_dumper_t *p = NULL;
 static pcap_t *pd;
 
 static int supports_monitor_mode;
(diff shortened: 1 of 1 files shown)


guyharris:
SIGINFO appears to be an oddly-chosen signal for this purpose; on BSD systems, it's intended as a way to force a program to dump out progress information by typing ^T.

Perhaps SIGUSR2 should be used, instead; that way, 1) it's the same on all UN\*Xes and 2) it separates the "print progress information" functionality from the "force flushing of the capture file" functionality.

GuyLewin:
I see what you're saying @guyharris , I just wanted to make it as simple as possible without ruining your interface.
Would you like me to create that in a new SIGUSR2 handler?

GuyLewin:
@guyharris Does this look better?

PR #673, opened 2018-04-23
Resync SMI list against Wireshark

jeremybrowne:
I realize #655 exists to automate pulling in the entire IANA entireprise-numbers database (only as OUIs, I think, not SMIs), however it's fairly quiet recently and already has conflicts to resolve prior to a future merge.

In the meantime, this change resyncs the SMI list against Wireshark master's current epan/sminmpec.h, theoretically a fairly comprehensive set of the vendors people frequently encounter. I've been patching a few of them into my distribution of tcpdump for years and thought I should upstream.


sgeto:
Like all of Wireshark epan/sminmpec.h is GPL. The codes, of course, aren't. How is that to be interpreted?

infrastation:
It seems to me, macros and array elements do not make enough sense as source code for the licensing differences to matter in this case.

PR #678, opened 2018-05-19
Add option to show packet direction

teknoraver:
Add a `-Q show` option to show the packet direction. Example:

```
$ tcpdump -ni enp9s0 -Q show
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on enp9s0, link-type EN10MB (Ethernet), capture size 262144 bytes
<  18:57:39.586150 IP 192.30.253.125.443 > 192.168.85.6.42512: Flags [P.], seq 3024695923:3024695954, ack 1359064726, win 31, options [nop,nop,TS val 919701565 ecr 2600782278], length 31
 > 18:57:39.586397 IP 192.168.85.6.42512 > 192.30.253.125.443: Flags [P.], seq 1:36, ack 31, win 329, options [nop,nop,TS val 2600842337 ecr 919701565], length 35
<  18:57:39.683849 IP 192.30.253.125.443 > 192.168.85.6.42512: Flags [.], ack 36, win 31, options [nop,nop,TS val 919701590 ecr 2600842337], length 0
<  18:57:39.926715 STP 802.1d, Config, Flags [none], bridge-id 8000.00:90:a9:f0:73:d1.8001, length 43
 > 18:57:40.139927 IP 192.168.85.6.44592 > 192.168.85.34.443: Flags [P.], seq 1812804671:1812804898, ack 1090051569, win 1394, options [nop,nop,TS val 2115702367 ecr 3472791082], length 22
```

Depends on the-tcpdump-group/libpcap#719

### DIFF
--- a/netdissect.h
+++ b/netdissect.h
@@ -198,6 +198,9 @@ struct netdissect_options {
   int   ndo_packettype;	/* as specified by -T */
 
   int   ndo_snaplen;
+#ifdef HAVE_PCAP_SETDIRECTION
+  int   ndo_showdir;		/* show packet direction */
+#endif
 
   /*global pointers to beginning and end of current packet (during printing) */
   const u_char *ndo_packetp;
--- a/print.c
+++ b/print.c
@@ -373,6 +373,22 @@ pretty_print_packet(netdissect_options *ndo, const struct pcap_pkthdr *h,
 	 * bigger lengths.
 	 */
 
+#ifdef HAVE_PCAP_SETDIRECTION
+	if (ndo->ndo_showdir) {
+		switch (h->dir) {
+		case PCAP_D_IN:
+			ND_PRINT("<  ");
+			break;
+		case PCAP_D_OUT:
+			ND_PRINT(" > ");
+			break;
+		default:
+			ND_PRINT("   ");
+			break;
+		}
+	}
+#endif
+
 	ts_print(ndo, &h->ts);
 
 	/*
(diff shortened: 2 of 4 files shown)


infrastation:
The Linux SLL/SLL2 data link types solve this problem by having the direction field in their DLT-specific structure, rather than in the common header.

teknoraver:
@infrastation What are the disadvantages of the SLL link type? Can we always put an interface in cooked mode without issues?

guyharris:
> What are the disadvantages of the SLL link type?

The link-layer header is stripped off (except for, it appears, frames captured on an 802.11 virtual-whatever-it's-called in monitor mode), so, for example, you can't see the Ethernet source and destination address and Ethernet type fields for Ethernet packets.

> Can we always put an interface in cooked mode without issues?

I suspect there would be people raising issues about not seeing link-layer headers.

So, at most, we could offer, for all devices on Linux, an option to select `DLT_LINUX_SLL` and `DLT_LINUX_SLL2` *instead* of the native header; you'd specify that with tcpdump and TShark with the `-y` flag (`-y LINUX_SLL` or `-y LINUX_SLL2`), and specify it in the Wireshark GUI by selecting "Linux cooked capture", or whatever it's called, rather than "Ethernet" or whatever.

teknoraver:
I did it in #719, and the prerequisite patch for libpcap is the-tcpdump-group/libpcap/pull/784

PR #679, opened 2018-05-21
Support crypto (AKA eay) with MSVC as well

sgeto:


sgeto:
At last:
https://ci.appveyor.com/project/guyharris/tcpdump/build/747#L143

gvanem:
Where does `c:\Win32-OpenSSL` and the *ancient* `libeay32.dll` come from? A built-in in AppVeyor? The OpenSSL DLLs have been called `libcrypto.dll` and `libssl.dll` for quite some time now.

sgeto:
It's one of these http://www.slproweb.com/products/Win32OpenSSL.html
And it's very recent (probable OpenSSL-1.0.2o or so).

I don't know why OpenSSL's VC++ libs have a different name. I always thought that naming scheme is like windump vs. tcpdump.
And when I follow OpenSSL building instructions for MSVC (in INSTALL.W32) the libs it spits out are these.

gvanem:
That was changed in April-2017 (?). If I do a:
```
perl Configure VC-WIN32 --prefix=%cd
nmake
```
on current master, I get these:
```
.\libssl-1_1.dll
.\libcrypto-1_1.dll
.\libcrypto.lib
.\libssl.lib
```

sgeto:
um, simply `nmake `after `configure`?
And that works?

According to INSTALL.W32, you should do a `nmake -f ms\ntdll.mak`.
The versioning is different too. I get 1.0.2X and you have 1.1.0X.
Not sure really...

mcr:
Rebase?  Can we be sure that we are building against an openssl that itself is the current release?

PR #680, opened 2018-05-23
 We should use a separate input buffer as well 

mruprich:
The issue #586 still created some problems on ppc64(le) architectures. I think we should use separate input buffer as well and it should be the same size as the output buffer. Also I think we should use memset for both buffers or allocate them with calloc.

### DIFF
--- a/print-esp.c
+++ b/print-esp.c
@@ -255,7 +255,7 @@ int esp_print_decrypt_buffer_by_ikev2(netdissect_options *ndo,
 	memset(input_buffer, 0, output_buffer_size);
 	memcpy(input_buffer, buf, len);
 
-	EVP_Cipher(ctx, output_buffer, buf, output_buffer_size);
+	EVP_Cipher(ctx, output_buffer, input_buffer, output_buffer_size);
 	EVP_CIPHER_CTX_free(ctx);
 
 	/*
--- a/print-esp.c
+++ b/print-esp.c
@@ -193,7 +193,7 @@ int esp_print_decrypt_buffer_by_ikev2(netdissect_options *ndo,
 	unsigned int len;
 	EVP_CIPHER_CTX *ctx;
 	unsigned int block_size, output_buffer_size;
-	u_char *output_buffer;
+	u_char *output_buffer, *input_buffer;
 
 	/* initiator arg is any non-zero value */
 	if(initiator) initiator=1;
@@ -241,7 +241,21 @@ int esp_print_decrypt_buffer_by_ikev2(netdissect_options *ndo,
 		EVP_CIPHER_CTX_free(ctx);
 		return 0;
 	}
-	EVP_Cipher(ctx, output_buffer, buf, len);
+
+	//same size as the output_buffer
+	input_buffer = (u_char *)malloc(output_buffer_size);
+	if(input_buffer == NULL) {
+		(*ndo->ndo_warning)(ndo, "can't allocate memory for decryption buffer");
+		EVP_CIPHER_CTX_free(ctx);
+		return 0;
+	}
+	
+	//I believe we should either use calloc or set both buffers to 0
+	memset(output_buffer, 0, output_buffer_size);
+	memset(input_buffer, 0, output_buffer_size);
+	memcpy(input_buffer, buf, len);
+
+	EVP_Cipher(ctx, output_buffer, buf, output_buffer_size);
 	EVP_CIPHER_CTX_free(ctx);
 
 	/*
@@ -250,6 +264,7 @@ int esp_print_decrypt_buffer_by_ikev2(netdissect_options *ndo,
 	 */
 	memcpy(buf, output_buffer, len);
 	free(output_buffer);
+	free(input_buffer);
 
 	ndo->ndo_packetp = buf;
 	ndo->ndo_snapend = end;


guyharris:
Where is this requirement that the input buffer for `EVP_cipher()` be a multiple of the cipher block size documented?

(For that matter, where is `EVP_cipher()` documented *at all*?)

guyharris:
> To be honest I haven't found the documentation on this function neither. But I have spoken with a colleague of mine in Red Hat who maintains openssl and he advised me to do it like this.

So it's an undocumented requirement or, at least, a not-very-obviously-documented requirement.  They should either lift the requirement or document it more clearly.

> He also said that the EVP_cipher() is kind of a low level function and that functions EVP_CipherInit, EVP_CipherUpdate and EVP_CipherFinal should be used instead.

Those *are* documented, but I saw nothing about any requirement imposed on the *input* buffer.

You should probably put a comment indicating that the reason why we're doing this is that (at least on some platforms) the *input* buffer must be a multiple of the cipher block size, and change the commit message to indicate that the underlying problem isn't that you need a *separate* buffer, the problem is that the buffer must be a multiple of the cipher block size. with zero padding.

guyharris:
> He also said that the EVP_cipher() is kind of a low level function and that functions EVP_CipherInit, EVP_CipherUpdate and EVP_CipherFinal should be used instead.

Could you also ask him whether the *input* buffer size restriction still applies if those routines are used and, if so, 1) whether that restriction is intentional, 2) if it's intentional, why it's not documented and 3) if it's not intentional, whether implementations that impose it (such as the little-endian 64-bit Power Architecture implementation, apparently) should be fixed or whether fixing it will still leave enough systems in the field with it unfixed, so that it should be *made* an intentional requirement *ex post facto*?

mruprich:
> You should probably put a comment indicating that the reason why we're doing this is that (at least on some platforms) the input buffer must be a multiple of the cipher block size

I think that here the problem was that the input buffer simply had a different size than the output buffer.

mruprich:
> Could you also ask him whether the input buffer size restriction still applies if those routines are used and, if so, 1) whether that restriction is intentional, 2) if it's intentional, why it's not documented and 3) if it's not intentional, whether implementations that impose it (such as the little-endian 64-bit Power Architecture implementation, apparently) should be fixed or whether fixing it will still leave enough systems in the field with it unfixed, so that it should be made an intentional requirement ex post facto?

I'll ask him and get back to you.

guyharris:
> I think that here the problem was that the input buffer simply had a different size than the output buffer.

OK, [the OpenSSL 1.1.0 man page for EVP cipher routines](https://www.openssl.org/docs/man1.1.0/crypto/EVP_CipherUpdate.html) says:

> EVP_EncryptUpdate() encrypts inl bytes from the buffer in and writes the encrypted version to out. This function can be called multiple times to encrypt successive blocks of data. The amount of data written depends on the block alignment of the encrypted data: as a result the amount of data written may be anything from zero bytes to (inl + cipher_block_size - 1) so out should contain sufficient room. The actual number of bytes written is placed in outl. It also checks if in and out are partially overlapping, and if they are 0 is returned to indicate failure.
> 
> If padding is enabled (the default) then EVP_EncryptFinal_ex() encrypts the "final" data, that is any data that remains in a partial block. It uses standard block padding (aka PKCS padding) as described in the NOTES section, below. The encrypted final data is written to out which should have sufficient space for one cipher block. The number of bytes written is placed in outl. After this function is called the encryption operation is finished and no further calls to EVP_EncryptUpdate() should be made.
> 
> If padding is disabled then EVP_EncryptFinal_ex() will not encrypt any more data and it will return an error if any data remains in a partial block: that is if the total data length is not a multiple of the block size.
> 
> EVP_DecryptInit_ex(), EVP_DecryptUpdate() and EVP_DecryptFinal_ex() are the corresponding decryption operations. EVP_DecryptFinal() will return an error code if padding is enabled and the final block is not correctly formatted. The parameters and restrictions are identical to the encryption operations except that if padding is enabled the decrypted data buffer out passed to EVP_DecryptUpdate() should have sufficient room for (inl + cipher_block_size) bytes unless the cipher block size is 1 in which case inl bytes is sufficient.

...

> EVP_CIPHER_CTX_set_padding() enables or disables padding. This function should be called after the context is set up for encryption or decryption with EVP_EncryptInit_ex(), EVP_DecryptInit_ex() or EVP_CipherInit_ex(). By default encryption operations are padded using standard block padding and the padding is checked and removed when decrypting. If the pad parameter is zero then no padding is performed, the total amount of data encrypted or decrypted must then be a multiple of the block size or an error will occur.

...

> NOTES
> 
> Where possible the EVP interface to symmetric ciphers should be used in preference to the low level interfaces. This is because the code then becomes transparent to the cipher used and much more flexible. Additionally, the EVP interface will ensure the use of platform specific cryptographic acceleration such as AES-NI (the low level interfaces do not provide the guarantee).
> 
> PKCS padding works by adding n padding bytes of value n to make the total length of the encrypted data a multiple of the block size. Padding is always added so if the data is already a multiple of the block size n will equal the block size. For example if the block size is 8 and 11 bytes are to be encrypted then 5 padding bytes of value 5 will be added.
> 
> When decrypting the final block is checked to see if it has the correct form.
> 
> Although the decryption operation can produce an error if padding is enabled, it is not a strong test that the input data or key is correct. A random block has better than 1 in 256 chance of being of the correct format and problems with the input data earlier on will not produce a final decrypt error.
> 
> If padding is disabled then the decryption operation will always succeed if the total amount of data decrypted is a multiple of the block size.

This *seems* to indicate that, if padding is enabled (which is the default) 1) the input for decryption must be a multiple of the cipher block size and 2) the output block must be of size >= (inl + cipher_block_size).  This doesn't seem to be the same as "they must be the same size".

mruprich:
> This seems to indicate that, if padding is enabled (which is the default) 1) the input for decryption must be a multiple of the cipher block size and 2) the output block must be of size >= (inl + cipher_block_size). This doesn't seem to be the same as "they must be the same size".

You are right, my mistake. I misinterpreted the message from my colleague. The input buffer must be a multiple of the cipher block size just as you say. I am not sure what the size of buf here is but perhaps I should still use the input_buffer and alloc it to the correct size?

guyharris:
> perhaps I should still use the input_buffer and alloc it to the correct size?

If, as noted, the input must be padded, with zeroes, to a multiple of the cipher block size, yes, that's what we need to do.

I've checked in change 4ba56b8c2e1240e2526db357c3d8bed4e9777bad that should fix this - in *both* places where we do decryption.  It also updates comments and variable names, allocates the input buffer before allocating the output buffer, zeroes out only the padding in the input buffer, and ensures that we free the input buffer is freed if we fail to allocate the output buffer.

mruprich:
@guyharris There is one little mistake left in the commit. When calling EVP_Cipher, the last parameter should be the buffer_size not len. I tried that on Power arch and it requires the buffer_size to run correctly. Also looking at the EVP_Cipher function in evp.h header file, even though it is not commented or documented, the function looks like this:
`int EVP_Cipher(EVP_CIPHER_CTX *c, unsigned char *out, const unsigned char *in, unsigned int inl);`
I believe that the inl might stand for "in length".

mruprich:
Any update on the last comment?

guyharris:
> When calling EVP_Cipher, the last parameter should be the buffer_size not len.

Well, as noted, they don't bother documenting `EVP_Cipher()`, but [the documentation for various of the EVP_ routines](https://www.openssl.org/docs/man1.1.0/crypto/EVP_DecryptUpdate.html) says of `EVP_DecryptUpdate()`:

> EVP\_DecryptInit\_ex(), EVP\_DecryptUpdate() and EVP\_DecryptFina\l_ex() are the corresponding decryption operations. EVP\_DecryptFinal() will return an error code if padding is enabled and the final block is not correctly formatted. The parameters and restrictions are identical to the encryption operations except that if padding is enabled the decrypted data buffer **out** passed to EVP\_DecryptUpdate() should have sufficient room for (**inl** + cipher\_block\_size) bytes unless the cipher block size is 1 in which case **inl** bytes is sufficient.

and says of `EVP_EncryptUpdate()`:

> EVP\_EncryptUpdate() encrypts **inl** bytes from the buffer **in** and writes the encrypted version to **out**. This function can be called multiple times to encrypt successive blocks of data. The amount of data written depends on the block alignment of the encrypted data: as a result the amount of data written may be anything from zero bytes to (inl + cipher\_block\_size - 1) so **out** should contain sufficient room. The actual number of bytes written is placed in **outl**. It also checks if **in** and **out** are partially overlapping, and if they are 0 is returned to indicate failure.

I see nothing there to indicate one way or the other whether *inl* should include padding bytes or not.  "In length" could be "size of the input buffer" or "number of bytes of input data, not including padding to a multiple of the block size".

If I look at the NOTES section of the man page in question, it says:

> PKCS padding works by adding **n** padding bytes of value **n** to make the total length of the encrypted data a multiple of the block size. Padding is always added so if the data is already a multiple of the block size n will equal the block size. For example if the block size is 8 and 11 bytes are to be encrypted then 5 padding bytes of value 5 will be added.
> 
> When decrypting the final block is checked to see if it has the correct form.
> 
> Although the decryption operation can produce an error if padding is enabled, it is not a strong test that the input data or key is correct. A random block has better than 1 in 256 chance of being of the correct format and problems with the input data earlier on will not produce a final decrypt error.
> 
> If padding is disabled then the decryption operation will always succeed if the total amount of data decrypted is a multiple of the block size.

So if PKCS padding is used, we need to pad with the number of padding bytes, *not* with zeroes.

So the question is whether the payloads we decrypt use PKCS padding.

ISAKMPv2 encrypted payload is one of them; it's described in [section 3.14 of RFC 7296](https://tools.ietf.org/html/rfc7296#section-3.14):

    o  Padding MAY contain any value chosen by the sender, and MUST have
        a length that makes the combination of the payloads, the Padding,
        and the Pad Length to be a multiple of the encryption block size.
        This field is encrypted with the negotiated cipher.

    o  Pad Length is the length of the Padding field.  The sender SHOULD
        set the Pad Length to the minimum value that makes the combination
        of the payloads, the Padding, and the Pad Length a multiple of the
        block size, but the recipient MUST accept any length that results
        in proper alignment.  This field is encrypted with the negotiated
        cipher.

"MAY contain any value chosen by the sender" seem to indicate that it's *not* PKCS padding, as it's not guaranteed to be padded with bytes with a value equal to the number of padding bytes.

For that case, there appears to be no guarantee, even in valid packets, that the payload will be padded to a multiple of the encryption block size, given that, even though there's a MUST for the length in the first paragraph, there's also a MUST in "MUST accept any length that results in proper alignment" in the second paragraph - presumably "proper alignment" just means "the length of the encrypted IKE payloads plus the length of the padding plus 1 byte for the padding length must be a multiple of 4", so that the Integrity Checksum Data is aligned on a 4-byte boundary.

So, for *that* case, it appears that we should try to decrypt packets where the payload *isn't* padded to a multiple of the block size, can do so by padding with zeroes, and should *not* specify PKCS padding, as the padding shouldn't be checked for "validity".

The other one is ESP packet payloads; that's described in [section 2 of RFC 2406](https://tools.ietf.org/html/rfc2406#section-2):

    2.4  Padding (for Encryption)

    Several factors require or motivate use of the Padding field.

           o If an encryption algorithm is employed that requires the
             plaintext to be a multiple of some number of bytes, e.g.,
             the block size of a block cipher, the Padding field is used
             to fill the plaintext (consisting of the Payload Data, Pad
             Length and Next Header fields, as well as the Padding) to
             the size required by the algorithm.

           o Padding also may be required, irrespective of encryption
             algorithm requirements, to ensure that the resulting
             ciphertext terminates on a 4-byte boundary. Specifically,
             the Pad Length and Next Header fields must be right aligned
             within a 4-byte word, as illustrated in the ESP packet
             format figure above, to ensure that the Authentication Data
             field (if present) is aligned on a 4-byte boundary.

           o Padding beyond that required for the algorithm or alignment
             reasons cited above, may be used to conceal the actual
             length of the payload, in support of (partial) traffic flow
             confidentiality.  However, inclusion of such additional
             padding has adverse bandwidth implications and thus its use
             should be undertaken with care.

    The sender MAY add 0-255 bytes of padding.  Inclusion of the Padding
    field in an ESP packet is optional, but all implementations MUST
    support generation and consumption of padding.

           a. For the purpose of ensuring that the bits to be encrypted
              are a multiple of the algorithm's blocksize (first bullet
              above), the padding computation applies to the Payload
              Data exclusive of the IV, the Pad Length, and Next Header
              fields.

           b. For the purposes of ensuring that the Authentication Data
              is aligned on a 4-byte boundary (second bullet above), the
              padding computation applies to the Payload Data inclusive
              of the IV, the Pad Length, and Next Header fields.

    If Padding bytes are needed but the encryption algorithm does not
    specify the padding contents, then the following default processing
    MUST be used.  The Padding bytes are initialized with a series of
    (unsigned, 1-byte) integer values.  The first padding byte appended
    to the plaintext is numbered 1, with subsequent padding bytes making
    up a monotonically increasing sequence: 1, 2, 3, ...  When this
    padding scheme is employed, the receiver SHOULD inspect the Padding
    field.  (This scheme was selected because of its relative simplicity,
    ease of implementation in hardware, and because it offers limited
    protection against certain forms of "cut and paste" attacks in the
    absence of other integrity measures, if the receiver checks the
    padding values upon decryption.)

    Any encryption algorithm that requires Padding other than the default
    described above, MUST define the Padding contents (e.g., zeros or
    random data) and any required receiver processing of these Padding
    bytes in an RFC specifying how the algorithm is used with ESP.  In
    such circumstances, the content of the Padding field will be
    determined by the encryption algorithm and mode selected and defined
    in the corresponding algorithm RFC.  The relevant algorithm RFC MAY
    specify that a receiver MUST inspect the Padding field or that a
    receiver MUST inform senders of how the receiver will handle the
    Padding field.

The latter seems to suggest that, for ESP payloads, the payload is supposed to be padded as appropriate, and we should not even try to decrypt payloads if the payload isn't a multiple of the block size.  It appears to be the sender's responsibility to pad the payload as appropriate for the encryption algorithm being used, and it's probably the responsibility of the crypto library to do the padding if the encryption algorithm specifies a type of padding.  I'm guessing the "monotonically increasing sequence" padding, however, would *not* be checked by the crypto library, as that's a requirement imposed by ESP, not by the encryption algorithm.  Unless there's a really easy way to determine whether we need to do that or not, using calls to the crypto library, I'm inclined *not* to bother checking the padding.

> I tried that on Power arch and it requires the buffer_size to run correctly.

Then perhaps the OpenSSL people should document that.

guyharris:
At least as I read both [RFC 5996 section 3.14](https://tools.ietf.org/html/rfc5996#section-3.14) and [RFC 4303 section 2.4](https://tools.ietf.org/html/rfc4303#section-2.4), the *sender* should pad the encrypted payload as necessary for decryption, so we don't need to copy the input buffer and pad it, we need to check to make sure the payload length is a multiple of what the cipher requires and reject it if not.

Issue #338, opened 2013-10-31
[Feature] Want an option to exit after specified time


sonots:
I want an option to let tcpdump exit after a specified time passed (seconds, minutes, hours). 

The reason is because I want to start tcpdump only during a specific time like 10:00 - 13:00. I often use crontab to start tcpdump on a specific time (like 10:00) with `-c` option to let tcpdump automatically die, but I can not assure it will not die until a specific time  (like 13:00 here) and will die after the specific time. 

Especially, if the captured server is service-out, it will not receive packets any more, then the tcpdump lives forever. I want to avoid such situations, too

jaunix:
This seems like a fairly trivial feature to add, https://github.com/the-tcpdump-group/tcpdump/compare/master...jaunix:stop-after-time

     $ ./tcpdump -h
     tcpdump version 4.8.0-PRE-GIT_2015_03_16
     libpcap version 1.8.0-PRE-GIT_2015_03_07
     OpenSSL 1.0.1e-fips 11 Feb 2013
     Usage: tcpdump [-aAbdDefhHIJKlLnNOpqRStuUvxX#] [ -B size ] [ -c count ]
                     [ -C file_size ] [ -E algo:secret ] [ -F file ] [ -G seconds ]
                     [ -i interface ] [ -j tstamptype ] [ -M secret ] --limit-time seconds] [ --number ]
                     [ -Q in|out|inout ]
                     [ -r file ] [ -s snaplen ] [ --time-stamp-precision precision ]
                     [ --immediate-mode ] [ -T type ] [ --version ] [ -V file ]
                     [ -w file ] [ -W filecount ] [ -y datalinktype ] [ -z command ]
                     [ -Z user ] [ expression ]
     $ date ; sudo ./tcpdump --limit-time 3 -w /tmp/foo ; date
     Mon Mar 16 15:53:44 PDT 2015
     tcpdump: listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
     5 packets captured
     5 packets received by filter
     0 packets dropped by kernel
     Mon Mar 16 15:53:49 PDT 2015
     $

gvanem:
You want to do it with a call to `signal(SIGALRM,..)`? MSVC/MingW doesn't have that. There must be a  cleaner way to do this.

zhexuany:
[...]

PR #684, opened 2018-06-10
Added a flag (-R) to terminate capture after specified time

steinbrecher:
Context/Background: I needed this feature to be able to capture short bursts 
(single ms to tens of ms) on high-rate interfaces. Googling produced some 
results of people trying to do timed captures using watch(1) or other 
coarser-grained checks. These OS-level checks also relied on the runtime 
of the program, not the time interval of packets captured. Adding this 
check seems more reliable, easier to use, and provides finer-grained control.

The check works by adding a check to the callback function passed to pcap;
it saves the time of the first packet seen and then compares later packets
to that time. Time interval is specified with units of microseconds,
milliseconds, seconds, minutes, or hours. Internally, this gets converted
to an integer number of microseconds.

Updated the generated manpage and the printed usage. Haven't added any
tests; I didn't see any for other command line flags like -c, but can
add a couple if desired.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -192,6 +192,8 @@ static int pflag;			/* don't go promiscuous */
 #ifdef HAVE_PCAP_SETDIRECTION
 static int Qflag = -1;			/* restrict captured packet by send/receive direction */
 #endif
+static int64_t Rflag = 0; /* max time to capture after first packet, in microseconds */
+struct timeval first_packet_time; /* store the first packet time, for use in check_runtime */
 static int Uflag;			/* "unbuffered" output of dump files */
 static int Wflag;			/* recycle output files after this number of files */
 static int WflagChars;
@@ -230,7 +232,10 @@ static NORETURN void show_remote_devices_and_exit(void);
 static void print_packet(u_char *, const struct pcap_pkthdr *, const u_char *);
 static void dump_packet_and_trunc(u_char *, const struct pcap_pkthdr *, const u_char *);
 static void dump_packet(u_char *, const struct pcap_pkthdr *, const u_char *);
+static void check_runtime(const struct pcap_pkthdr *);
 static void droproot(const char *, const char *);
+
+static int64_t parse_runtime_str(const char*);
 
 #ifdef SIGNAL_REQ_INFO
 static void requestinfo(int);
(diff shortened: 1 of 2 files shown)


infrastation:
This seems to implement the request made in #338.

steinbrecher:
Needs a quick cleanup & rebase if the feature is of interest -- @infrastation how is that determined?

infrastation:
As another user points it out, there is the `timeout` command that does the job on the OSes traditional for `tcpdump`, so there are arguments pro and contra embedding `timeout` into `tcpdump`. Requesting more feedback on the tcpdump-workers mailing list would not be a bad idea, it seems to me.

mcr:
Can you use a long option?

tcpdump.c
```diff
@@ -688,6 +754,7 @@ static const struct option longopts[] = {
 #ifdef HAVE_PCAP_SETDIRECTION
 	{ "direction", required_argument, NULL, 'Q' },
 #endif
+  { "max-runtime", required_argument, NULL, 'R', },
```
guyharris: Please use the same indentation as other code.

Note: tab stops are set every *eight* spaces, not every *four* spaces, this having originally been written on UN\*X. :-)  If you're reading it with a program with tab stops set every four spaces, it won't display correctly.

guyharris:
Note that without an interval timer (`setitimer()`/`alarm()`) on UN\*X, and an equivalent on Windows, there's no guarantee, on a network with a low level of traffic, that there won't be an arbitrary amount of time after the time expires when the capture terminates.

tcpdump.c
steinbrecher: Will clean up when I rebase :)

steinbrecher:
> Note that without an interval timer (`setitimer()`/`alarm()`) on UN*X, and an equivalent on Windows, there's no guarantee, on a network with a low level of traffic, that there won't be an arbitrary amount of time after the time expires when the capture terminates.

Yeah, very true. Thoughts on adding an interval timer equal to the selected time +1 second (or something equally "long" relative to a short timer)?

steinbrecher:
> Can you use a long option?

`--max-runtime` is a substitute for `-R` in the code as is

PR #687, opened 2018-06-26
Capsicum-related fixes for tcpdump

hselasky:


infrastation:
It would not harm to have some comments in the commit messages or the source code to explain these changes to the folks in future.

hselasky:
Are you referring to both commits?

infrastation:
Well, yes.

hselasky:
Better now?

infrastation:
Looks much more comprehensible to me now, thank you.

guyharris:
So there are no backends that 1) *do* have a file descriptor but 2) *don't* support Capsicum?

For example, does the [Netmap](http://info.iet.unipi.it/~luigi/netmap/)  support in `pcap-netmap.c` support Capsicum?  If not, testing the result of `pcap_fileno()`, to make sure it's not -1, is not sufficient to eliminate the use of Capsicum on Netmap devices.

Or is it a case of "support Capsicum" really meaning that, to support Capsicum with a given device, you have to enable the *particular* ioctls used by the device, in which case, to fix this, there will have to be support in libpcap (as that's the only place that knows which ioctls have to be enabled)?

hselasky:
For example libibverbs have a bunch of file descriptors associated with the data transport object. None of these are registered by libpcap, so at the moment we enter sandboxing, these transport objects stop working.

From what I understand, the file handle in question is only used for BPF, Berkeley Packet Filter, compatible file handles. If it is -1 it means BPF is not in use. I'll have to dig a bit deeper in the code to verify this. Basically these bits been tested under FreeBSD and verified. And there is also an identical review pending with some more information:

https://reviews.freebsd.org/D15785

guyharris:
> For example libibverbs have a bunch of file descriptors associated with the data transport object.

That's *one particular* example; you need to look at *all* the backends that will run on OSes that support Capsicum (is it FreeBSD-only, or do any other OSes that support libpcap support it?) to determine whether this change is sufficient.

> None of these are registered by libpcap

Are there libibverbs call to fetch the descriptors from the various handles `pcap-rdmasniff.c` opens?  If so, again, perhaps what's needed is a per-backend "Capsicumize" function pointer, which would be null on platforms that don't support Capsicum, and would perform the appropriate operations on platforms that do, combined with "Capsicumize" APIs in libpcap, which tcpdump would use.

> From what I understand, the file handle in question is only used for BPF, Berkeley Packet Filter, compatible file handles.

Not true - perhaps, in my previous comment, I should have *explicitly* indicated that the file handle is set by the Netmap backend, and therefore that `pcap_fileno()` will *not* return -1. 

> If it is -1 it means BPF is not in use.

"The return value of `pcap_fileno()` is -1" implies "BPF is not in use".

"BPF is not in use", however, does not imply "the return value of `pcap_fileno()` is -1".

hselasky:
> Are there libibverbs call to fetch the descriptors from the various handles pcap-rdmasniff.c opens?

This is not possible currently, so capsicum + libibverbs is a no-go.

infrastation:
Please allow some more time for a more thorough review.

emaste:
> Please allow some more time for a more thorough review.

Do you have a rough guess of the timeline? We're approaching code freeze for FreeBSD 12 and would prefer to include a change identical to upstream, but if it will be a while yet we'll incorporate a local patch and revert to upstream later.

infrastation:
To answer one of the Guy's questions: [Capsicum project page](https://www.cl.cam.ac.uk/research/security/capsicum/) says it is supported on FreeBSD, Linux and DragonFlyBSD, if that helps.

mcr:
Sorry this is taking so long.
May I ask you to rebase again?

hselasky:
> Sorry this is taking so long.
> May I ask you to rebase again?

Done.

fxlb:
Any reason to NOT merge this patch?

fxlb:
Perhaps the test should be explained...

infrastation:
@hselasky, a few months ago I made a few Capsicum-related bugfixes in tcpdump (commits 780f86b, ac23514 and 51f9c3b), it would be great if you could confirm these look good to you. Also it would be very useful to add a detailed comment into the proposed change, as it is often difficult to reconstruct the context.

hselasky:
I'll have a look. Thanks for the heads up.

emaste:
Is there anything remaining to do here?

hselasky:
@emaste : I have not observed any problems in FreeBSD-14-main as of recently. Are these patches imported into our code-base? If yes, then this issue can be closed.

emaste:
It looks like the one change under "Commits" is in FreeBSD, but is not in upstream tcpdump.

infrastation:
I had another round at comprehending this change and managed to understand it this time. Excuse us for the delay, there has been a lot of backlog.

PR #689, opened 2018-07-12
Display interface on live captures on DLT_LINUX_SLL2

pevik:
The reason why displaying interface only for live captures is different
index when display pcap file on another machine than the one which was
doing the capture on [1].

See: GH the-tcpdump-group/libpcap#127

[1] https://lists.sandelman.ca/pipermail/tcpdump-workers/2018-July/001019.html

Signed-off-by: Petr Vorel <pvorel@suse.cz>
Suggested-by: Guy Harris <guy@alum.mit.edu>

### DIFF
--- a/print-sll.c
+++ b/print-sll.c
@@ -25,6 +25,10 @@
 #include <config.h>
 #endif
 
+#ifdef HAVE_NET_IF_H
+#include <net/if.h>
+#endif
+
 #include "netdissect-stdinc.h"
 
 #include "netdissect.h"
@@ -395,6 +399,9 @@ sll2_if_print(netdissect_options *ndo, const struct pcap_pkthdr *h, const u_char
 	u_short ether_type;
 	int llc_hdrlen;
 	u_int hdrlen;
+#ifdef HAVE_NET_IF_H
+	char ifname[IF_NAMESIZE];
+#endif
 
 	ndo->ndo_protocol = "sll2_if";
 	if (caplen < SLL2_HDR_LEN) {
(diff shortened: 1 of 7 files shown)


print-sll.c
```diff
@@ -408,6 +415,10 @@ sll2_if_print(netdissect_options *ndo, const struct pcap_pkthdr *h, const u_char
 	}
 
 	sllp = (const struct sll2_header *)p;
+#ifdef HAVE_NET_IF_H
+	if (ndo->ndo_live_capture && if_indextoname(EXTRACT_BE_U_4(sllp->sll2_if_index), ifname))
+		ND_PRINT("IFNAME %s ", ifname);
+#endif
```
infrastation: Most of the time the call will be successful. But when the packet reaches tcpdump after the interface was deleted (which is more likely with low packet rate and TPACKET_V3, which can buffer packets long enough), the ifindex at this point will refer to a non-existent interface:
>       On success, if_indextoname() returns ifname; on error, NULL is returned
>       and errno is set appropriately.

infrastation: Maybe "ifindex %u (%s)" on success and just "ifindex %u" on failure?

pevik: Changed it to suggested "IFNAME %s (%u) " on success on live capture otherwise "IFNAME %u ". 

Do we want to show "IFNAME %s (%u) " also while reading packet? Not sure about it (on the same localhost it's a benefit, but can be misleading).

pevik: When I'm thinking about it, printing always interface name looks better for me. I'll remove `ndo_live_capture` and give a warning when not live capture.

pevik:
Implemented previous suggestion.

pevik:
@guyharris @infrastation this is a shy ping. Any objections or can it be merged?

guyharris:
Fine with me.  Anybody else?

pevik:
I've just removed first commit with updating `configure`. It was wrong (what `autoconf` adds `autoheader` removes + recent commit abe1aa5a shows that it wasn't needed).

infrastation:
Errrm, I would say something like "ifindex %u" and "ifindex %u (%s)" would be easier to parse if anyone needs to, rather than the format in the proposed commit.

pevik:
@infrastation I got your point, reverse it as ifindex is for sure, but ifname not. Make sense, I'll do it. I'll just use upper case (as for other) + keep the warning.

pevik:
@guyharris @infrastation OK, at the end I used output suggested by Denis:
`"ifindex %u (%s) "` and `"ifindex %u "`.

infrastation:
Thank you.

PR #693, opened 2018-07-30
Add check for pcap warning PCAP_WARNING_TSTAMP_TYPE_NOTSUP

steinbrecher:
Previously would silently accept any of "host", "adapter", or
"adapter_unsynced" regardless of whether the interface supported
the option.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -1306,7 +1306,7 @@ open_interface(const char *device, netdissect_options *ndo, char *ebuf)
 			error("%s: Can't set time stamp type: %s",
 		              device, pcap_statustostr(status));
 		else if (status > 0)
-		  warning("Could not set timestamp type '%s' on %s: %s",
+		  warning("When trying to set timestamp type '%s' on %s: %s",
 			  pcap_tstamp_type_val_to_name(jflag), device,
 			  pcap_statustostr(status));
 	}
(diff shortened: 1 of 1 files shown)


tcpdump.c
```diff
@@ -1305,6 +1305,10 @@ open_interface(const char *device, netdissect_options *ndo, char *ebuf)
 		if (status < 0)
 			error("%s: Can't set time stamp type: %s",
 		              device, pcap_statustostr(status));
+    else if (status == PCAP_WARNING_TSTAMP_TYPE_NOTSUP)
```
guyharris: 1) The indentation appears to be different from the check for an error.

2) You might as well check for *any* status > 0, as that means "warning"; that makes it robust against future warnings being added.

tcpdump.c
```diff
@@ -1305,6 +1305,10 @@ open_interface(const char *device, netdissect_options *ndo, char *ebuf)
 		if (status < 0)
 			error("%s: Can't set time stamp type: %s",
 		              device, pcap_statustostr(status));
+		else if (status > 0)
+		  warning("Could not set timestamp type '%s' on %s: %s",
```
guyharris: The warning call is indented differently from the error call.

steinbrecher:
Indentation issues should be fixed -- this is what I get for trying something other than emacs.

infrastation:
Merged as a single commit to keep the history clear. As a side note, indentation isn't quite consistent in that part of the file -- it mixes a few styles (see attached screenshot).
screenshot_2018-07-31_13-50-41

guyharris:
Indentation in that routine cleaned up in 67ac8ff216c90612122a4a4e7ed812993a13be1f.

PR #698, opened 2018-08-15
Feature Request: Added support for EAP dump in RADIUS Messages [RDY]

aravindprasad90:
Usually, for EAP related authentications, RADIUS messages contain EAP payload. Currently, RADIUS dump messages donot display the EAP frame information. Hence, added changes for displaying EAP info within RADIUS dump messages. 

Sample O/P:

1) EAP Request Identity Frame:

Before this change:
```
         EAP-Message Attribute (79), length: 19, Value: .
```
       
After this change:
```
           EAP-Message Attribute (79), length: 10, Value:
                 Response (2), id 1, len 8
                 Type Identity (1), Identity: avi
            0x0000:  0201 0008 0161 7669
```

2) EAP Request Challenge Frame:
Before this change:
```
          EAP-Message Attribute (79), length: 24, Value: ..
            0x0000:  0102 0016 0410 cd32 576a 04bd ea97 14c6
            0x0010:  9e84 ae7c 6de2
```

After this change:
```
          EAP-Message Attribute (79), length: 24, Value:
                 Request (1), id 2, len 22
                 Type MD5-challenge (4)
            0x0000:  0102 0016 0410 cd32 576a 04bd ea97 14c6
            0x0010:  9e84 ae7c 6de2
```

3) EAP Response Challenge frame:

Before this change:
```
          EAP-Message Attribute (79), length: 27, Value: ..
            0x0000:  0202 0019 0410 035f 1f16 c4e2 d71e 1553
            0x0010:  6b55 effc 0a0f 6176 69
```

After this change:
```
          EAP-Message Attribute (79), length: 27, Value:
                 Response (2), id 2, len 25
                 Type MD5-challenge (4)
            0x0000:  0202 0019 0410 035f 1f16 c4e2 d71e 1553
            0x0010:  6b55 effc 0a0f 6176 69
```

4) EAP Success Frame:

Before this change:
```
          EAP-Message Attribute (79), length: 6, Value: ..
            0x0000:  0302 0004
```

After this change:
```
          EAP-Message Attribute (79), length: 6, Value:
                 Success (3), id 2, len 4
            0x0000:  0302 0004
```

### DIFF
--- a/print-eap.c
+++ b/print-eap.c
@@ -148,56 +148,36 @@ static const struct tok eap_aka_subtype_values[] = {
  */
 void
 eap_print(netdissect_options *ndo,
-          const u_char *cp,
+          const u_char *tptr,
           u_int length)
 {
-    const struct eap_frame_t *eap;
-    const u_char *tptr;
-    u_int eap_type, tlen, type, subtype;
-    int count=0, len;
+    u_int type, subtype;
+    int len, count;
 
-    ndo->ndo_protocol = "eap";
-    tptr = cp;
-    tlen = length;
-    eap = (const struct eap_frame_t *)cp;
-    ND_TCHECK_SIZE(eap);
-    eap_type = EXTRACT_U_1(eap->type);
+    ND_TCHECK_1(tptr);
+    type = EXTRACT_U_1(tptr);
+    ND_TCHECK_2(tptr + 2);
+    len = EXTRACT_BE_U_2(tptr + 2);
+    if(len != (int)length) {
+       goto trunc; 
+    }
+    ND_PRINT(", %s (%u), id %u, len %u",
+            tok2str(eap_code_values, "unknown", type),
+            type,
+            EXTRACT_U_1((tptr + 1)),
+            len);
 
-    ND_PRINT("%s (%u) v%u, len %u",
-           tok2str(eap_frame_type_values, "unknown", eap_type),
-           eap_type,
-           EXTRACT_U_1(eap->version),
-           EXTRACT_BE_U_2(eap->length));
+    ND_TCHECK_LEN(tptr, len);
 
-    if (ndo->ndo_vflag < 1)
-        return;
+    if (type == EAP_REQUEST || type == EAP_RESPONSE) {
+        /* RFC 3748 Section 4.1 */
+        ND_TCHECK_1(tptr + 4);
+        subtype = EXTRACT_U_1(tptr + 4);
+        ND_PRINT("\n\t\t Type %s (%u)",
+                tok2str(eap_type_values, "unknown", subtype),
+                subtype);
 
-    tptr += sizeof(struct eap_frame_t);
-    tlen -= sizeof(struct eap_frame_t);
-
-    switch (eap_type) {
-    case EAP_FRAME_TYPE_PACKET:
-        ND_TCHECK_1(tptr);
-        type = EXTRACT_U_1(tptr);
-        ND_TCHECK_2(tptr + 2);
-        len = EXTRACT_BE_U_2(tptr + 2);
-        ND_PRINT(", %s (%u), id %u, len %u",
-               tok2str(eap_code_values, "unknown", type),
-               type,
-               EXTRACT_U_1((tptr + 1)),
-               len);
-
-        ND_TCHECK_LEN(tptr, len);
-
-        if (type == EAP_REQUEST || type == EAP_RESPONSE) {
-            /* RFC 3748 Section 4.1 */
-            ND_TCHECK_1(tptr + 4);
-            subtype = EXTRACT_U_1(tptr + 4);
-            ND_PRINT("\n\t\t Type %s (%u)",
-                   tok2str(eap_type_values, "unknown", subtype),
-                   subtype);
-
-            switch (subtype) {
+        switch (subtype) {
             case EAP_TYPE_IDENTITY:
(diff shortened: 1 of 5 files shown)


aravindprasad90:
Hi @infrastation 

>>Most of the new code in the proposed commit has been copied from print-eap.c, is there a good reason not to call the existing function eap_print() from print_attr_string() instead?

infrastation:
Thank you for reworking this change to reuse existing code, please allow some time to review this change again.

aravindprasad90:
Hi @infrastation 

Yes Ok. Thanks for the support. 

It has been more than 1.5 years since I initially raised the pull request with initial set of changes. Hence, if possible, kindly review it sooner. Thanks for the support.

print-radius.c
```diff
@@ -626,6 +628,7 @@ print_attr_string(netdissect_options *ndo,
                   const u_char *data, u_int length, u_short attr_code)
 {
    u_int i;
+   int len;
```
infrastation: Should this be unsigned?

print-radius.c
```diff
@@ -671,6 +674,13 @@ print_attr_string(netdissect_options *ndo,
            data++;
            length--;
         break;
+      case EAP_MESSAGE:
+           eap_print(ndo, data);                  
+           ND_TCHECK_2(data + 2);
+           len = EXTRACT_BE_U_2(data + 2);
```
infrastation: Should this length be checked to fit within the current structure before calling `eap_print()`, and should it be passed as an argument to `eap_print()` so it can validate its own structures? (The answers are not necessarily "yes", I did not proof-read every step yet.)

aravindprasad90:
Hi @infrastation 

Thanks for the review. 

1. Updated the code to make eap_print() API to return the success(0)/failure(-1). This is required since incase of any failures (type, value errors) in API, nd_print_trunc() would be already called and hence, in the caller, code must immediatley return.
2. Changes type of len to uint
3. Reg question of value checks for length variable, I think they not may be needed.  And further, incase of any error in eap_print() API, we will immediately return at caller too and hence, seperate checks may not be necessary.

Let me know your views. Thanks for the support.

aravindprasad90:
Hi @infrastation 

Thanks a lot again for the review. My bad, sorry for those errors. Corrected them all. 

Kindly review and let me know your views. thanks for the support.

infrastation:
Thank you for the update. Let me try to explain how to get these details right and to keep this code safe.

The function `eapol_print()` knows the amount of bytes allocated for the EAP packet in the EAPOL frame. The function `print_attr_string()` knows the amount of bytes allocated for the EAP packet in the RADIUS attribute. These functions need to pass that amount into `eap_print()`, but they do not need to access into the EAP packet to avoid violation of layers. The function `eap_print()` needs to verify that the amount of bytes is sufficient for the EAP structure in the packet, which has its own length field (keep the optional EAP padding from RFC 3748 Section 4 in mind). If the function `eap_print()` finds that the EAP packet was truncated, it needs to signal this to the calling function.

aravindprasad90:
Hi @infrastation 

Thanks a lot for the explanation. Understood your point. Introduced the length parameter in eap_print() API and added the check for verifying the length field in its own packet and the length passed from outer later (EAPOL/RADIUS). 

Thanks a lot for the support and guidance. Kindly review the changes.

aravindprasad90:
Hi @infrastation 

Kindly review the changes and let me know your views. Thanks for the support.

aravindprasad90:
Hi @infrastation

Kindly review the changes and let me know your views. Thanks for the support.

print-eap.c
```diff
@@ -278,9 +258,50 @@ eap_print(netdissect_options *ndo,
             case EAP_TYPE_EXPERIMENTAL:
             default:
                 break;
-            }
         }
-        break;
+    }
+    rc = 0;
```
infrastation: Obviously `rc` has no purpose.

print-radius.c
```diff
@@ -671,6 +673,15 @@ print_attr_string(netdissect_options *ndo,
            data++;
            length--;
         break;
+      case EAP_MESSAGE:
+           ND_TCHECK_2(data + 2);
```
infrastation: This code block is wrong because it does not check `length` before accessing into `data`, and because it likely does not need to access into `data` in the first place as I tried to explain in August. Also this does not seem to be a good place for this block because when the EAP data is shorter than the RADIUS attribute, the rest of the attribute will be printed as a string (unless I misunderstood the code flow, feel free to correct me); it seems more appropriate for `print_attr_strange()`.

print-eap.c
aravindprasad90: Agreed, addressed.

print-radius.c
aravindprasad90: Thanks for the review again. Yes, agree with your point. 

Made code changes to avoid accessing the data part outside of eap_print() API (for maintained layered approach). 

Also, in cases where the EAP data is lesser than RADIUS attribute, its an erroneous scenario and nd_print_trunc() will be called by eap_print() API itself and hence, as such, no more printing will be required. 

Kindly review the changes and let me know your views.

Thanks a lot for the continued support.

aravindprasad90: Hi @infrastation 

Kindly review the changes and let me know your views.

aravindprasad90: Hi @infrastation @guyharris 

Kindly review the changes and let me know your views.

PR #699, opened 2018-08-23
print nfs write verifier cookie for write and commit operation

viralkmehta:


infrastation:
Do you have a .pcap that shows the effect of the new code?

infrastation:
It would be most helpful if you added a new test case based on a .pcap file that contains the required NFS packet(s) (`tcpdump -#` will show packet numbers and`editcap` can filter these out for you).

viralkmehta:
Yes, I looked at that and I did not find any relevant existing .pcap file for general NFS testing.

I can see below pasted nfs testing .pcap files.
Let me know which one you think is the most appropriate to add new test and I will do that.

[mehtav3@centos72_base tests]$ grep -i nfs TESTLIST
unaligned-nfs-1 unaligned-nfs-1.pcap    unaligned-nfs-1.out     -v
nfs-attr-oobr           nfs-attr-oobr.pcap              nfs-attr-oobr.out
hoobr_nfs_printfh       hoobr_nfs_printfh.pcap          hoobr_nfs_printfh.out
hoobr_nfs_xid_map_enter hoobr_nfs_xid_map_enter.pcap    hoobr_nfs_xid_map_enter.out


Thanks,
Viral

print-nfs.c
```diff
@@ -1435,15 +1435,17 @@ parsecreateopres(netdissect_options *ndo,
...
             const uint32_t *dp, int verbose)
 {
 	u_int er;
 
 	if (!(dp = parsestatus(ndo, dp, &er)))
 		return (0);
-	return parse_wcc_data(ndo, dp, verbose) != NULL;
+	if (!(dp = parse_wcc_data(ndo, dp, verbose)))
+		return (0);
+	return dp;
```
infrastation: this is the same as `return parse_wcc_data(ndo, dp, verbose)`

infrastation:
A new separate test would be the most appropriate.

print-nfs.c
viralkmehta: Yup, that is true, I can make it simpler.

viralkmehta:
Added a new commit which adds new test for the code changes

print-nfs.c
```diff
@@ -1803,8 +1807,14 @@ interp_reply(netdissect_options *ndo,
 
 	case NFSPROC_COMMIT:
 		dp = parserep(ndo, rp, length);
-		if (dp != NULL && parsewccres(ndo, dp, ndo->ndo_vflag) != 0)
+		if (dp != NULL && (dp = parsewccres(ndo, dp, ndo->ndo_vflag)) != 0) {
```
infrastation: `!= NULL` would be more appropriate so long as the test is for a pointer.

print-nfs.c
```diff
@@ -1668,6 +1668,10 @@ interp_reply(netdissect_options *ndo,
 					ND_PRINT(" <%s>",
 						tok2str(nfsv3_writemodes,
 							NULL, EXTRACT_BE_U_4(dp + 1)));
+
+					/* write-verf-cookie */
+					ND_TCHECK_8(dp + 2);
+					ND_PRINT(" verf %" PRIx64, EXTRACT_BE_U_8(dp + 2));
 				}
```
infrastation: The above reads as "4 bytes addressed as dp[1], dp[2], dp[3] and dp[4] stand for the write mode" and "8 bytes addressed as dp[2], dp[3], dp[4], dp[5], dp[6], dp[7], dp[8] and dp[9] stand for the cookie". Clearly there is a 3-byte overlap, could you double-check and confirm?

print-nfs.c
```diff
@@ -1803,8 +1807,14 @@ interp_reply(netdissect_options *ndo,
 
 	case NFSPROC_COMMIT:
 		dp = parserep(ndo, rp, length);
-		if (dp != NULL && parsewccres(ndo, dp, ndo->ndo_vflag) != 0)
+		if (dp != NULL && (dp = parsewccres(ndo, dp, ndo->ndo_vflag)) != 0) {
+			if (ndo->ndo_vflag > 1) {
+				/* write-verf-cookie */
+				ND_TCHECK_8(dp);
+				ND_PRINT(" verf %" PRIx64, EXTRACT_BE_U_8(dp));
```
infrastation: Likewise, here `dp` is an argument to `parsewccres()` and a pointer to the cookie, is this intentional?

infrastation:
If the proposed change follows RFC 7530, it could be helpful to refer to particular sections as that's quite a large document.

infrastation:
Thank you for the comments, now it makes more sense. Do you have references to specific sections of the protocol specification?

viralkmehta:
This one is standard reply for NFS WRITE operation and the verifier cookie is valid for both NFSv3 and NFSv4. So, far it was not being printed and I added the change.

For RFC section, you can look at, 
https://tools.ietf.org/html/rfc1813#section-3.3.7

viralkmehta:
And under the section I mentioned, look at
      struct WRITE3resok {
           wcc_data    file_wcc;
           count3      count;
           stable_how  committed;
           writeverf3  verf;
      };

The last field is what this patch is printing, now.

viralkmehta:
Hi,

Btw, had a question on who merges the code ? And when does that happen generally ?

infrastation:
Thank you for the spec references, that is helpful. One or another maintainer merges pull requests when the changes are ready for production and the maintainer is able to proof-read. The amount of time and work differs, you can look through closed pull requests if interested. This pull request should not take long to merge.

infrastation:
I see how the existing code relates with the WRITE operation definition (Section 3.3.7), in that the structure is different for an OK and for a non-OK status. The COMMIT operation (Section 3.3.21) code does not seem to take the status into account (it does not call `parsestatus()` at all), so something may be wrong in this decoder, but the new code has bounds check, so even if it produces incorrect output, it will do it safely.

PR #702, opened 2018-09-25
Subject: Introduce -nn option

Boucher-Lambert:
This changes the semantics on -n option so only namelookups are skipped. Port
numbers *are* translated to their string representations. Option -nn then has
the same semantics as -n had originally.

This is a partial upstreaming of tcpdump-4.9.2-3 used in CentOs 7.5.

Signed-off-by: Timothee Boucher-Lambert <timothee.boucher-lambert@tessares.net>

### DIFF
--- a/addrtoname.c
+++ b/addrtoname.c
@@ -802,7 +802,7 @@ init_servarray(netdissect_options *ndo)
 
 		while (table->name)
 			table = table->nxt;
-		if (ndo->ndo_nflag) {
+		if (ndo->ndo_nflag > 1) {
 			(void)nd_snprintf(buf, sizeof(buf), "%d", port);
 			table->name = strdup(buf);
 		} else
@@ -1226,7 +1226,7 @@ init_addrtoname(netdissect_options *ndo, uint32_t localnet, uint32_t mask)
 		f_localnet = localnet;
 		f_netmask = mask;
 	}
-	if (ndo->ndo_nflag)
+	if (ndo->ndo_nflag > 1)
 		/*
 		 * Simplest way to suppress names.
 		 */
--- a/tcpdump.1.in
+++ b/tcpdump.1.in
@@ -546,7 +546,11 @@ Use \fIsecret\fP as a shared secret for validating the digests found in
 TCP segments with the TCP-MD5 option (RFC 2385), if present.
 .TP
 .B \-n
-Don't convert addresses (i.e., host addresses, port numbers, etc.) to names.
+Don't convert host addresses to names.  This can be used to avoid
+DNS lookups.
+.TP
+.B \-nn
+Don't convert protocol and port numbers etc. to names either.
 .TP
 .B \-N
 Don't print domain name qualification of host names.
--- a/tests/TESTonce
+++ b/tests/TESTonce
@@ -21,7 +21,7 @@ if ($^O eq 'MSWin32') {
 else {
     # we used to do this as a nice pipeline, but the problem is that $r fails to
     # to be set properly if the tcpdump core dumps.
-    $r = system "../tcpdump 2>/dev/null -# -n -r $input $options >NEW/$output";
+    $r = system "../tcpdump 2>/dev/null -# -nn -r $input $options >NEW/$output";
     if($r != 0) {
         # this means tcpdump failed.
         open(OUTPUT, ">>"."NEW/$output") || die "fail to open $output\n";


infrastation:
I vaguely remember a discussion along those lines, but do not remember why it did not result in such a change. Regardless if it was some code complications or just lack of time, from a user point of view the proposed change looks clear.

infrastation:
Requested feedback on the list.

frisso:
I would prefer to avoid breaking the semantic of already implemented switches.
My proposal would be to create a new switch (e.g., "-na" do not resolve addresses, which implicitly tells the user that port names can be translated) with the additional behavior, and leave "-n" as it is now.

sethhall:
Wow, please don't merge this change.  There are lots of great points being made on the tcpdump-workers list but for me the absolute most convincing one is that it would suddenly make my typical set of command line flags have a different behavior and it would be inconsistent between systems. 

With most tools all I'm ever trying to do is figure out how to disable this behavior since pre-providing the lookup in /etc/services has rarely been helpful to me.

ckreibich:
Please, no! Do not change established command line usage, period. This is a big deal. For the folks already used to -n vs -nn, running -nn already achieves the same thing, so if you really want to now allow distinction of just what doesn't get resolved, do it via a new mechanism, like the one Joerg suggests on the list.

infrastation:
@Boucher-Lambert, could you please clarify the motivation for having the "don't resolve addresses, but resolve service names" mode, regardless of how exactly the user activates it? I could think of potential scenarios where it would be useful, but what is your practical use case?

infrastation:
Would everybody be willing to accept that `-n` remains as it is now (resolve neither hostnames nor service names) and `-nn` becomes "do not resolve hostnames, but resolve service names)?

mcr:
Denis Ovsienko <notifications@github.com> wrote:
    > Would everybody be willing to accept that -n remains as it is now
    > (resolve neither hostnames nor service names) and -nn becomes "do not
    > resolve hostnames, but resolve service names)?

I'm happy with that.

Boucher-Lambert:
> Would everybody be willing to accept that `-n` remains as it is now (resolve neither hostnames nor service names) and `-nn` becomes "do not resolve hostnames, but resolve service names)?

No.

I proposed this change because I noticed a difference in the semantic between the mainstream tcpdump and the tcpdump used in CentOs.

In CentOs, the options are as follow:
       _-n     Don't convert host addresses to names.  This can be used to avoid DNS lookups.
       -nn    Don't convert protocol and port numbers etc. to names either._

With current implementation of `-n` in tcpdump, there is style somehow a compatibility ( since `-nn` will be interpreted as `-n`) with a code written for CentOs.

If  `-nn` becomes "do not resolve hostnames, but resolve service names", the semantic becomes the opposite of what it is in CentOs.

infrastation:
I have studied the change proposed in this pull request and I confirm you have explained its immediate effect correctly. However, an unsolicited default behaviour change in the CentOS tcpdump package (or any other tcpdump package, for that matter) does not on its own make valid grounds for the default behaviour change everywhere. If the proposed change is merged as it is now, it will likely cause a long-time annoyance for the software maintainers for no obvious gain.

As to "code written for CentOS", you still have not updated on the specific use cases, hence from your comments it looks like making a CentOS-specific change in one software and then having to make CentOS-specific changes in other software, i.e. keeping developers busy to achieve what was available in the first place -- original tcpdump and code written for generic Linux/POSIX.

If your goal is to introduce a new mode that disables host name lookups but leaves service name lookups in place, there are other, much cleaner ways to achieve that in CentOS and elsewhere. One of which could be, for example, adding a command-line parameter like `--resolve-service-names` and migrating the code to use it. If you decide to work in that direction, I would be willing to assist in merging such a change. But I believe it would be wrong to accept the change originally proposed in this pull request.

micolous:
This comment is offered _neutrally,_ after trying to troubleshoot some tests in another software package which failed as a result of this patch.

I found this patch [has been included in Red Hat's distributions of tcpdump since 2001](https://src.fedoraproject.org/rpms/tcpdump/c/d42d137a2e259f46cb569ef92f34b43bc6802e03
), as `tcpdump-3.6.1-portnumbers.patch`:

```
* Sat Jan 13 2001 Pekka Savola <pekkas@netcore.fi>            
  - Make -n resolve port/protocol numbers but not hostnames, -nn for no            
    resolving at all
```

The [publicly accessible mailing list archives for tcpdump](https://seclists.org/tcpdump/) only appear to go back to 2002, but I found [a previous upstreaming attempt in 2009](https://seclists.org/tcpdump/2009/q3/110).

It does not appear that this patch has been applied to Debian or Gentoo.  I haven't checked other distributions of `tcpdump`.

PR #703, opened 2018-10-02
Add files generated by CMake on Windows to .gitignore

dmiller-nmap:
These files were generated by running `CMake.exe` and then building the Debug configuration in Visual Studio 2015. I'll push one more commit in a moment for files created by building the other configurations.


guyharris:
So this is for doing an in-tree build, running CMake from the top-level source tree, rather than, for example, making a separate `build` directory and running CMake from that directory, right?

dmiller-nmap:
@guyharris Yes, that's correct. I hadn't considered using a separate build directory (we're a little behind the times like that at Nmap :smile: ). This is pretty low priority, then, considering that, and I'll understand if you want to just close it out.

PR #705, opened 2018-10-02
Change unknown attribute in test

herwinw:
The value 127 is actually assigned to the Location-Information Attribute
(RFC 5580). Change it to a very high value.

This way it won't interfere when we'll add the attributes from RFC5580 to the list.


fxlb:
Merged as a4e008c5c47af5f3713f6b7336970797432ac94a.

fxlb:
Thanks.

PR #706, opened 2018-10-03
WIP: Add RADIUS attributes from RFC5580

herwinw:
I've started decoding the RADIUS attributes from RFC5580, but I'm reaching the point where a first review and some pointers in the right direction might get useful.

The main issue here is [Location-Information](https://tools.ietf.org/html/rfc5580#section-4.2), which contains a lot of inner fields, making it complex to print. My main issues here:

* Snooping a field currently consists of calling an EXTRACT-macro, manually advancing a pointer to `data` and finally decrementing `length` with all the fields already processed. This feels like taking 2 steps too many. Is there an existing way to simplify this?
* The timestamps are the 64 bit format also used in ntp. Is it okay to extract some logic from `print-ntp.c` to a shared module that can be used by the ntp and radius dissectors?

The file `tests/rfc5580.packet` is something I feed to `radclient` to create the packets and will not be included in the final commit (once this is all done)


mcr:
Herwin <notifications@github.com> wrote:
    > The timestamps
    > are the 64 bit format also used in ntp. Is it okay to extract some
    > logic from print-ntp.c to a shared module that can be used by the ntp
    > and radius dissectors?

That's a great idea.
Make a separate pull request.

    > The Travis tests will fail, but that is not really related to this PR
    > (see #705). The file tests/rfc5580.packet is something I feed to
    > radclient to create the packets and will not be included in the final
    > commit (once this is all done)

Put it somewhere else, but please include it, as someone else might want to
use it.

--
]   Michael Richardson, Sandelman Software Works        | network architect  [

herwinw:
@mcr Regarding the radclient-statement: this is more a generic problem than something specific for this issue. I've implemented a number of RADIUS-RFCs and all the pcap-files in the test-folder for RADIUS have been generated that way.
Problem is: these RADIUS-client-files are no foolproof way to regenerate the PCAP files. `radclient` uses a random client port, the UDP packet size may change, the RADIUS authenticator will change if the packet data (or the secret) is changed. I found myself constatly copy-pasting the new generated packet when working on this code.

mcr:
If we have the script, we can at least repeat what you did.
I think that I can teach radclient to use the same port number,
if I go to the effort if spining up a radius server.

print-radius.c
```diff
@@ -995,6 +1030,143 @@ print_attr_mip6_home_link_prefix(netdissect_options *ndo,
...
+   entity = EXTRACT_U_1(data);
+   data++;
+
+   sighting_time = EXTRACT_BE_U_8(data);
+   data += 8;
+
+   time_to_live = EXTRACT_BE_U_8(data);
+   data += 8;
+
+   ND_PRINT("Index %u, code %s, entity %s, sighting time %llu, time to live %llu: ",
```
guyharris, line 1090: Use PRIu64 instead of %llu - that's the C99 way of printing a u_int64_t (which is *not* guaranteed to be an `unsigned long long int`):

    ND_PRINT("Index %u, code %s, entity %s, sighting time %llu, time to live %"PRIu64": ",

fxlb:
Please, use the new GET_ macros instead of EXTRACT_ ones (See 176e182 and ee68aa3).
`s/EXTRACT_/GET_/g`
The new GET_ macros have embedded bounds checks, via get_ inline functions (see extract.h).

fxlb:
You should replace u_intN_t by C99-style uintN_t.

infrastation:
Let me know if you need help rebasing and updating this pull request, it should not be difficult to resolve this before long.

infrastation:
Rebased on the current master and added the trivial fixups requested above (please squash these before/during the merge).

Did **not** review any other aspects of the proposed change.

infrastation:
`print_attr_operator_name()` checked for NUL character when printing the Operator-Name part of the Text value, although Section 4.1 defines that string as counted, not terminated. It also tested if each character is printable. Please see the fixup commit that uses `nd_printn()` instead of the custom code and either object or squash it into the final version.

infrastation:
For `print_attr_location_information()` the same stands as above, plus I have changed the unknown value format strings to avoid duplication. That is another fixup commit.

Besides that, it would be nice to have Sighting Time and Time-to-Live decoded from NTP format (as stated in respective source code comments), although to me Section 4.2 reads as if Time-to-Live value is both absolute (NTP timestamp) and relative ("for how long"). I accept it may not be immediately doable.

Two more attributes yet to review.

infrastation:
For `print_attr_location_data()` it seems to be wrong to print the String field as a string.

Its structure depends on the Code value (which for some reason is stored in the Location-Information attribute); for Civic (0) the data is a variable-length structure with TLVs (RFC 4776 Section 3 less the first 3 octets); for Geospatial (1) the data is a fixed-length structure (RFC 3825 Section 2 less the first 2 octets). For any other code point the structure is unknown.

This way, there is no case where printing the data as a string would produce any useful output, and the sample packet has invalid data for location index 1. Could you double-check?

infrastation:
In `print_basic_location_policy_rules()` the Note Well field is optional, so another fixup commit amends the code accordingly and uses `nd_printn()` as before. Also in Flags the reserved bits must be zero, so I have changed the code to make it obvious when any MBZ bits are set. Also it would be nice to decode Retention Expires since Section 4.4 is sufficiently clear that the value is absolute, but that's not critical. Rebased on the latest master branch.

**Summary:** please see the `print_attr_location_data()` comment above, which is the only outstanding major issue in the proposed changes.

infrastation:
Added another minor fixup to print NTP timestamps (since that is easy after merging #708), which makes it apparent that all three example timestamps are 1900-08-09T22:25:43Z. It would be better to have more realistic values there, also Time-to-Live and Retention are supposed to be greater than Sighting Time. Also letter case and puntuation are more consistent now.

infrastation:
Added a fixup to plug the Location-Data issue explained above. If anybody sees any outstanding issues in the code, please comment before long. I am going to clean the pile of WIP commits up and to merge the final version soon.

infrastation:
This has been merged as the originally proposed commit with one annotated fixup commit on top, thank you everyone for your input.

PR #707, opened 2018-10-03
Fix tab/spaces in GNUmakefile of win32

herwinw:


guyharris:
> Note that neither Travis nor AppVeyor uses this; the only Windows build system files I'm bothering to maintain, at least, are the files for CMake (as they can be used for UN*Xes and Windows).

This means that anybody who finds an issue with the GNUmakefile, or with the Visual Studio project files, will have to fix those files themselves, and submit a pull request if they want the fix to be in the repository.  It also means that they're likely to find issues appearing over time due to code rot....

PR #708, opened 2018-10-03
Extract NTP printing

herwinw:
This method is required to disect some RADIUS attributes (see #706)

The changes in the win32 folder are based on what grep could find for other shared modules, and have not been tested in any way.


mcr:
I think this code will have to be redone, as we have changed the ND* check macros a lot, and so it probably does not merge.  Can you redo this from master?

fxlb:
Rebased, merged as commit 763bef5f53e7776413961e26850f426863d520d3 with 2 changes:
1) fix typo in commit message,
2) update some #include.

Thanks for your patience.

Issue #709, opened 2018-10-05
Missing return value check


Bugcheckers:
Hi,

I have reported this to security@tcpdump.org but no reply yet so I wonder my email might have been missed.

We have found some missing checks :

1. Return value of malloc is not checked here:

https://github.com/the-tcpdump-group/tcpdump/blob/master/tcpdump.c#L1158

host_url = malloc(endp - url + 1);
memcpy(host_url, url, endp - url);
2. Return value of strchr might be NULL but are not checked here, but it is checked other parts of the code.

https://github.com/the-tcpdump-group/tcpdump/blob/master/smbutil.c#L466

p = strchr(++fmt, '}');

https://github.com/the-tcpdump-group/tcpdump/blob/master/smbutil.c#L868

p = strchr(fmt, ']');

3. Return value of strsep might be NULL and it is usually checked but it is dismissed here:

https://github.com/the-tcpdump-group/tcpdump/blob/master/print-esp.c#L579

spistr = strsep(&spikey, "@");

Thank you very much!

Bugcheckers:
#710

Bugcheckers:
#710

PR #710, opened 2018-10-09
return-checks: Added a return check for malloc and for strsep

agnosticdev:
Added a return check for `malloc` in tcpdump.c.
Added a return check for `strsep` in print-esp.c.
Fixed a small typo.
All tests passed.


Bugcheckers:
address issue #709

PR #715, opened 2018-11-06
Include correct capsicum header

rbgarga:
Current included header `sys/capability.h` is deprecated. All
supported FreeBSD versions have now the proper header `sys/capsicum.h`


guyharris:
My PC-BSD 9.1 system doesn't have `sys/capsicum.h`, but my PC-BSD 9.1 system also doesn't have a Flex capable of generating reentrant scanners, so I'm not *that* concerned with supporting FreeBSD 9 or earlier; if somebody wants to add configuration checks to include `sys/capability.h` on older systems, they may do so.

PR #718, opened 2018-11-22
Fix checksum calculation for IPv6 Segment Routing (SRv6) traffic

amsalam20:
The checksum calculation for IPv6 packets is based on a pseudo
header that includes the packets’ final Destination Address (DA).

If the IPv6 packet contains a Routing header, the final DA is the
last element of the Routing header [RFC8200].

Currently, tcpdump supports Routing header types 0 and 2.

IPv6 Segment Routing Header (SRH) is a new Routing header type (4).
SRH is defined in draft-ietf-6man-segment-routing-header [1].

This patch fix the checksum calculation for SRv6 packets. It allows
tcpdump to get the final DA value from SRv6 packets.

[1] https://tools.ietf.org/html/draft-ietf-6man-segment-routing-header-15


ahabdels:
Any news on this one ?

fxlb:
Draft is now '-17', any change ?

ahabdels:
There is no change regarding this as it is an architecture principle. At ingress node, the Segment Routing Header (SRH) is inserted between the IPv6 header and transport layer header. The Destination address (DA) of the packet is updated to be the first SID and accordingly to preserve to the original  destination address is added as the last SID in the SRH SID list. Hence, any node that needs to verify the checksum of packet must retrieve the packet original packet header which is the last as mentioned.

fxlb:
In the commit message you mean "packet's final Destination Address" ?

ahabdels:
Yes

fxlb:
Thanks, I will update that.

fxlb:
Merged, Thank you !

PR #719, opened 2018-11-24
allow to force a cooked capture

teknoraver:
Using a datalink header can be used to show packets direction


mcr:
New static flags like "yflag" are not the way to do stuff.  please add that to the ndo structure, and use a long option to enable it.

teknoraver:
> New static flags like "yflag" are not the way to do stuff. please add that to the ndo structure, and use a long option to enable it.

Done. The long option was already there as `--linktype`

PR #720, opened 2018-12-06
CVE-2018-19519 buffer overread.

pmgdeb:
Here is the patch with the suggested solution for CVE-2018-19519 as mentioned in [0] which initializes buf in print-hncp.c:print_prefix.

[0] https://github.com/zyingp/temp/blob/master/tcpdump.md

### DIFF
--- a/print-hncp.c
+++ b/print-hncp.c
@@ -208,6 +208,8 @@ print_prefix(netdissect_options *ndo, const u_char *prefix, u_int max_length)
     int plenbytes;
     char buf[sizeof("xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx::/128")];
 
+    buf[0] = '\0';
+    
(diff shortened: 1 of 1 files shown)


agnosticdev:
Very interesting. In CVE it states:
> This issue did not affect the versions of tcpdump as shipped with Red Hat Enterprise Linux 5 and 6.

Why is it affecting Red Hat Enterprise Linux 7?
Is this issue present outside of RHEL also?

pmgdeb:
> Why is it affecting Red Hat Enterprise Linux 7?
The affected code was introduced in tcpdump version 4.8.0-bp, see also (https://access.redhat.com/security/cve/cve-2018-19519).

> Is this issue present outside of RHEL also?
The reporter found it in Ubuntu, any distro using tcpdump > 4.8.0 should be affected.

infrastation:
Thank you for suggesting this change, but this problem has been identified and fixed a few months ago, the fix will be available in the next tcpdump release. I am sorry, but the report on which you have based this work is a duplicate (as well as the CVE ID). Closing.

PR #723, opened 2018-12-23
Add dissector for SSH version exchange

x-way:


mcr:
Looks great. Can you rebase?

fxlb:
Please, use new GET_ macros instead of EXTRACT_ ones (See 176e182 and ee68aa3).

print-ssh.c
```diff
@@ -0,0 +1,108 @@
...
+
+static int
+ssh_print_version(netdissect_options *ndo, const u_char *pptr, u_int len)
+{
+	u_int idx = 0;
+	const char *pnp;
+
+	ND_TCHECK_1(pptr+idx);
+	if ( EXTRACT_U_1(pptr+idx) != 'S' )
+		goto trunc;
```
fxlb: `trunc` means a truncation of the packet. It's not the case here and in line 40, 44, 48.

fxlb:
The new GET_ macros do the bounds checks with a setjmp/longjmp logic.
```
/* part of extract.h */

static inline uint8_t
get_u_1(netdissect_options *ndo, const u_char *p)
{
        if (!ND_TTEST_1(p))
                longjmp(ndo->ndo_truncated, 1);
        return EXTRACT_U_1(p);
}

[...]

#define GET_U_1(p) get_u_1(ndo, (const u_char *)(p))
```
Thus you can remove the ND_TCHECK_1 macros.

fxlb:
For a better git history, I suggest you to squash all the commit in one with `git rebase -i HEAD~8` (fixup (f) keyword for the last 7 commits)

fxlb:
Merged. Thank you.

PR #727, opened 2019-01-16
Fix and update reason codes based on 802.11-2016

Mister-X-:
There is an issue with numbering too, as you can see in the first commit, 2aaa600, and tcpdump doesn't display the right one if above 18.

802.11-2016 can be obtained on https://ieee.org

### DIFF
--- a/print-802_11.c
+++ b/print-802_11.c
@@ -930,7 +930,7 @@ static const char *reason_text[] = {
 	"Disassociated because the information in the Supported Channels "
 	  "element is unacceptable",				/* 11 */
 	"Disassociated due to BSS transition management",	/* 12 */
-	"Reserved",						/* 13 */
+	"Invalid Information Element",				/* 13 */
 	"Michael MIC failure",					/* 14 */
 	"4-Way Handshake timeout",				/* 15 */
 	"Group key update timeout",				/* 16 */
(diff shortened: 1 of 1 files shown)


print-802_11.c
```diff
@@ -927,27 +927,30 @@ static const char *reason_text[] = {
...
-	"Reserved",						/* 13 */
+	"Disassociated due to BSS transition management",	/* 12 */
+	"Invalid Information Element",				/* 13 */
 	"Michael MIC failure",					/* 14 */
 	"4-Way Handshake timeout",				/* 15 */
 	"Group key update timeout",				/* 16 */
 	"Information element in 4-Way Handshake different from (Re)Association"
 	  "Request/Probe Response/Beacon",			/* 17 */
 	"Group Cipher is not valid",				/* 18 */
+	"Pairwise Cipher is not valid",				/* 19 */
```
agnosticdev: Should 19 read, "Invalid pairwise cipher?"
reason_codes

Mister-X-: Yes, but I followed how some others were written. If you notice, a few others should be 'invalid' and are written as 'not valid'. Same goes for 'STA', but are written as 'station'.

I can redo this PR and fix all of that if you'd like.

agnosticdev: I will defer to the maintainers on if a new PR is needed, but I wanted to point it out.  Thank you for the follow up.

Mister-X-: Any update from the maintainers?

Mister-X-: Reached any decision on how you want the PR?

agnosticdev: @guyharris and @Mister-X- was there any next steps for the PR?

Mister-X-: There isn't anything in the comments above that tell me what the next step on my side would be. I can see a two possibilities on your side: merge it, or close it (and provide with the changes to do

mcr:
Hi, while I do 6tisch stuff, I don't know much about the reason codes. So whichever language you think is useful.  You may also want to talk to @kivinen .
Can you include some pcap files?  Ideally a packet with each reason code... I realize that might be rather difficult to do, but anything would help with future regression testing of this code.
And please rebase, I anticipate pulling your code in by end of June.

Mister-X-:
How about I close this PR, recreate one with updated and fixed reason codes with the exact same wording as in 802.11-2016?

Creating PCAPs is easy. It's basically knowing how one file is done (starting with reason code 0), then increment one byte until it reaches 255. I'll create a small C program to generate all possible PCAPs.

fxlb:
Difficult to have an avis without access at 802.11-2016. It seems this doc is behind a paywall.

Mister-X-:
There is screenshot attached to a conversation here: https://github.com/the-tcpdump-group/tcpdump/pull/727#discussion_r248641907

Feel free to ask @agnosticdev and I'm sure he can make a more complete screenshot for you.

kivinen:
IEEE has get ieee [1] program that allows access (click on IEEE Xplore's GET Programpage) to the 802 specifications (click GET 802(R) Standards) free of charge, which does include 802.11-2016 [3] also. To download the document you need to create account and then you should be able to download the document.

[1] https://standards.ieee.org/products-services/ieee-get-program.html
[2] https://ieeexplore.ieee.org/browse/standards/get-program/page/series?id=68
[3] https://ieeexplore.ieee.org/document/7786995

fxlb:
> IEEE has get ieee [1] program

Thanks, I will have a look.

PR #728, opened 2019-01-17
Add support for dissecting Broadcom tags

ffainelli:
This pull request adds support for decoding the 4byte Broadcom tag format used by a wide variety of Broadcom Ethernet switches. These tags are present in Ethernet frames between the MAC SA and the EtherType of the original frame, and therefore, we have a bit of guess work to find out whether they are there or not.

There is possibly a better way to do this, via the use of specific DLT/linktype values. I would still like to get feedback on the approach being taken here.


mcr:
In what circumstances would a capture see these tags?

ffainelli:
> In what circumstances would a capture see these tags?

See the discussion on tcpdump-workers@lists.tcpdump.org for specific details: 

https://lists.sandelman.ca/pipermail/tcpdump-workers/2019-January/001169.html

In a nutshell, you would be able to see those tags if you run a capture on the management interface that connects to the Ethernet switch, which is typically just another regular port of the switch, just with added functionality to tell you why you are seeing this packet to begin with.

ffainelli:
@mcr @guyharris can you review these two commits and let me know if the approach is acceptable with respect to allowing to specify the Ethernet header length. Thank you

ffainelli:
This is dependent on https://github.com/the-tcpdump-group/libpcap/pull/796

ffainelli:
Forced push an update which also takes care of dissection pre-pended (DSA_TAG_BRCM_PREPEND). We can use the print_encap_header() callback for that.

netdissect.h
```diff
@@ -435,6 +435,7 @@ extern u_int arcnet_if_print IF_PRINTER_ARGS;
 extern u_int arcnet_linux_if_print IF_PRINTER_ARGS;
 extern u_int atm_if_print IF_PRINTER_ARGS;
 extern u_int bt_if_print IF_PRINTER_ARGS;
+extern u_int brcm_tag_if_print IF_PRINTER_ARGS;
```
ffainelli: That hunk belongs in the next commit, not here.

ffainelli:
I need to make a couple of changes:

- there is one hunk in netdissect.h that belongs in patch 2
- the comment about print_ether_hdr_len() is not quite correct and refers to an earlier implementation where an offset was specified

ffainelli:
Now pushed an update which addresses the two things mentioned earlier as well as adds tests for Broadcom tags, "normal" and prepended.

guyharris:
> You have to add the new file to the `NETDISSECT_SOURCE_LIST_C` variable in `CMakeLists.txt`, too.

Or put the dissectors for Ethernet+switch tag into `print-ether.c`.

ffainelli:
> > You have to add the new file to the `NETDISSECT_SOURCE_LIST_C` variable in `CMakeLists.txt`, too.
> 
> Or put the dissectors for Ethernet+switch tag into `print-ether.c`.

Having a separate file allows for people to easily strip some functionality out if they desire a tiny tcpdump, see for instance OpenWrt's tcpdump-tiny package (which should probably be reworked to have a list of dissectors you can specify at ./configure time):

https://github.com/openwrt/openwrt/blob/master/package/network/utils/tcpdump/patches/100-tcpdump_mini.patch

guyharris:
> Having a separate file allows for people to easily strip some functionality out if they desire a tiny tcpdump

Then you'll have to fix the `CMakeLists.txt` file for this to pass the tests (and thus you'll have to fix that file for this to be merged).

ffainelli:
> > Having a separate file allows for people to easily strip some functionality out if they desire a tiny tcpdump
> 
> Then you'll have to fix the `CMakeLists.txt` file for this to pass the tests (and thus you'll have to fix that file for this to be merged).

Forced push an update with that. AppVeyor appears happy so far.

PR #733, opened 2019-02-12
Extract extend DNS rcode from OPT record.

marka63:
Extend DNS printing

This extracts the extended rcode bits from the OPT record's TTL field and combines it with rcode to display the actual rcode.  This adds rcodes BADVERS (16) and BADCOOKIE (23) to the set of known rcodes.

BADCOOKIE is defined in RFC 6891.   BadCookie is currently being mis-reported as YXRRset (23 - 16).

And a actual DNS COOKIE exchange where BADCOOKIE was returned and the client retries with the new server cookie.

tcpdump: listening on lo0, link-type NULL (BSD loopback), snapshot length 262144 bytes
10:33:16.245522 IP (tos 0x0, ttl 64, id 26511, offset 0, flags [none], proto UDP (17), length 68, bad cksum 0 (->1518)!)
    127.0.0.1.53324 > 127.0.0.1.53: 37225+ [1au] SOA? . (40)
10:33:16.245803 IP (tos 0x0, ttl 64, id 40438, offset 0, flags [none], proto UDP (17), length 84, bad cksum 0 (->dea0)!)
    127.0.0.1.53 > 127.0.0.1.53324: 37225 BadCookie 0/0/1 (56)
10:33:16.246129 IP (tos 0x0, ttl 64, id 34739, offset 0, flags [none], proto UDP (17), length 84, bad cksum 0 (->f4e3)!)
    127.0.0.1.51706 > 127.0.0.1.53: 18053+ [1au] SOA? . (56)
10:33:16.246387 IP (tos 0x0, ttl 64, id 27521, offset 0, flags [none], proto UDP (17), length 159, bad cksum 0 (->10cb)!)
    127.0.0.1.53 > 127.0.0.1.51706: 18053$ 1/0/1 . SOA a.root-servers.net. nstld.verisign-grs.com. 2019021202 1800 900 604800 86400 (131)


fenner:
Is there a way to see that the rcode might be wrong because we hit the snaplen while trying to parse for the OPT record?

marka63:
There is nothing in the header that indicates that a OPT record is present.  You only know that there isn't a OPT record once the additional section is parsed.

print-domain.c
```diff
@@ -47,8 +47,21 @@ static const char *ns_resp[] = {
...
 	" Resp12", " Resp13", " Resp14", " NoChange",
+	" BadVers", "Resp17", " Resp18", " Resp19",
+	" Resp20", "Resp21", " Resp22", " BadCookie",
 };
 
+static const char *
+ns_rcode(int rcode) {
+	static char buf[sizeof(" Resp4095")];
+
+	if (rcode < sizeof(ns_resp)/sizeof(ns_resp[0])) {
```
fxlb: New warning here:
```
./print-domain.c:58:12: warning: comparison of integers of different signs: 'int' and 'unsigned long' [-Wsign-compare]
        if (rcode < sizeof(ns_resp)/sizeof(ns_resp[0])) {
            ~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

fxlb: cc @marka63

PR #734, opened 2019-02-15
Support DLT_IEEE802_15_4_TAP.

jkcko:
New link type for IEEE 802.15.4 with additional meta-data.

See https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=15429 for full specification.

No actual parsing of meta-data TLVs are performed.  The header and TLVs are skipped and the PHY payload is passed to the common IEEE 802.15.4 print function.

### DIFF
--- a/print-802_15_4.c
+++ b/print-802_15_4.c
@@ -230,3 +230,33 @@ ieee802_15_4_if_print(netdissect_options *ndo,
 	ndo->ndo_protocol = "802.15.4_if";
 	return ieee802_15_4_print(ndo, p, h->caplen);
 }
+
+/* For DLT_IEEE802_15_4_TAP */
+/* https://github.com/jkcko/ieee802.15.4-tap */
+u_int
+ieee802_15_4_tap_if_print(netdissect_options *ndo,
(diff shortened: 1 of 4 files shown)


print-802_15_4.c
```diff
@@ -230,3 +230,18 @@ ieee802_15_4_if_print(netdissect_options *ndo,
 	ndo->ndo_protocol = "802.15.4_if";
 	return ieee802_15_4_print(ndo, p, h->caplen);
 }
+
+/* For DLT_IEEE802_15_4_TAP*/
+u_int
+ieee802_15_4_tap_if_print(netdissect_options *ndo,
+                      const struct pcap_pkthdr *h, const u_char *p)
+{
+	uint8_t version = EXTRACT_U_1(p);
```
guyharris: This needs to check whether `h->caplen` is >= 4; if it's not, you're fetching past the end of the packet buffer.

jkcko:
Commit updated.
Added sanity checks and default error printing to ieee802_15_4_tap_if_print parsing.
Add reference to specification on github.

PR #736, opened 2019-03-03
print-medsa: Generalize to also print MDSA

lunn:
Marvell SOHO Ethernet switches use two different tag formats.
EDSA uses an Ethertype, so can be identified from that. DSA lacks
the Ethertype, other information is needed to know it is present.
PCAP can now mark such frames, so generalize the EDSA printer to
also print DSA tagged frames.

Signed-off-by: Andrew Lunn <andrew@lunn.ch>


lunn:
Fixed up the unneeded ;
I assume the tests are failing because the libpcap it is using to build against does not yet have the DSA changes.

print-medsa.c
```diff
@@ -190,3 +202,59 @@ medsa_print(netdissect_options *ndo,
...
+		mdsa_print_full(ndo, mdsa, caplen - ETHER_SA_OFFSET);
+
+	/* We printed the Ethernet header already */
+	ndo->ndo_eflag = 0;
+
+	/* Parse the Ethernet frame regularly telling how big the non
+	 * standard Ethernet header is.
+	 *
+	 * +-----------++-----------++----------------++--------------+
+	 * | MAC DA (6)|| MAC SA (6)||Broadcom tag (4)||Type/Length(2)|
```
guyharris: Presumably you mean "Marvell tag" or "DSA tag" here, rather than "Broadcom tag"?

print-medsa.c
```diff
@@ -190,3 +202,59 @@ medsa_print(netdissect_options *ndo,
 trunc:
 	nd_print_trunc(ndo);
 }
+
+/* The DSA tag is 4 bytes. It has the same content as the last 4 bytes
```
guyharris: *Last* 4 bytes, or *first* 4 bytes?  It looks, from the definition of `struct mdsa_pkthdr`, as if the EDSA tag has 4 bytes of tag and 2 bytes of Ethernet type.

print-medsa.c
lunn: Yep. I Cut/paste from Florians' broadcom code, and did not fix it up.

print-medsa.c
lunn: Last.

EDSA has 2 bytes of Ethertype, 2 bytes reserved, and then 4 bytes which are identical to a DSA tag.
For printing EDSA, when medsa_print() is called, it is because the Ethertype has been matched. The Ethertype has already been pulled off, so medsa_print() only sees 6 bytes of the 8 byte header.
After the EDSA or DSA header comes the real ethertype value for the frame.  So for EDSA, this is the second Ethertype header in the frame.

lunn: Since the ether_type seems to be causing confusion, i removed it from the structure.

The broadcom cut/paste problem has also been fixed.

lunn:
The commit of Vivien code has made this redundant So closing.

PR #739, opened 2019-03-16
icmpv6: Add option for IPv6 only flag 

loganaden:
https://www.ietf.org/id/draft-ietf-6man-ipv6only-flag-05.txt which is close to becoming 
an IETF RFC: https://www.ietf.org/id/draft-ietf-6man-ipv6only-flag-05.txt.


fxlb:
Rebased and merged with an updated git message. Thank you.

mcr:
Would have been nice to have a sample (RA) packet to go with this.

PR #743, opened 2019-04-02
Add shortcuts for setting time stamp precision

rnsanchez:
Rationale: when you need to quickly start collecting packets into a savefile and you happen to need a different time stamp precision as the default, it is easy to confuse whether the proper option it is `--time-stamp-precision`, `--timestamp-precision`, or what.  These shortcuts do not change `tcpdump`'s behavior, they are equivalent to using the `--time-stamp-precision` option.

In order to present the shortcuts in `print_usage()`, it felt adequate to reshuffle the strings and group the ones related to time stamping (i.e., `--time-stamp-precision` and the included shortcuts).

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -670,6 +670,8 @@ show_remote_devices_and_exit(void)
 #define OPTION_IMMEDIATE_MODE		130
 #define OPTION_PRINT			131
 #define OPTION_LIST_REMOTE_INTERFACES	132
+#define OPTION_TSTAMP_MICRO		133
+#define OPTION_TSTAMP_NANO		134
 
 static const struct option longopts[] = {
 #if defined(HAVE_PCAP_CREATE) || defined(_WIN32)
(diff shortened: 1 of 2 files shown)


tcpdump.c
```diff
@@ -3074,11 +3082,13 @@ print_usage(void)
...
 	(void)fprintf(stderr,
-"\t\t[ -r file ] [ -s snaplen ]" TIME_STAMP_PRECISION_USAGE "\n");
+"\t\t[ -r file ] [ -s snaplen ] [ -T type ] [ --version ]\n");
 	(void)fprintf(stderr,
-"\t\t[ -T type ] [ --version ] [ -V file ] [ -w file ]\n");
+"\t\t[ -V file ] [ -w file ] [ -W filecount ] [ -y datalinktype ]\n");
+#ifdef HAVE_PCAP_SET_TSTAMP_PRECISION
 	(void)fprintf(stderr,
-"\t\t[ -W filecount ] [ -y datalinktype ]\n");
+"\t\t[ --time-stamp-precision precision ] [ --micro ] [--nano]\n");
```
guyharris: That should be `"... [ -- nano ]\n"`, with spaces around the argument, as with other flag arguments.

rnsanchez: Thank you for the review, I overlooked this.  Updated now.

PR #744, opened 2019-04-06
Add DLT support for decoding (E)DSA tags

vivien:
Following @ffainelli's [work](https://github.com/the-tcpdump-group/tcpdump/pull/728) on adding DLT support for parsing DSA tagged master interfaces, this PR adds support for decoding the Marvell DSA and Marvell EDSA tags.


print-dsa.c
```diff
@@ -0,0 +1,235 @@
...
+	{ 0x2, "IGMP/MLD Trap" },
+	{ 0x3, "Policy Trap" },
+	{ 0x4, "ARP Mirror" },
+	{ 0x5, "Policy Mirror" },
+	{ 0, NULL }
+};
+
+static u_int
+dsa_if_print_full(netdissect_options *ndo, const struct pcap_pkthdr *h,
+		  const u_char *p, uint taglen)
```
guyharris: `uint` isn't guaranteed to be defined as a type, but `u_int` is; use `u_int`, so this builds on Windows.

PR #745, opened 2019-04-11
Add 4-byte ASN tests, both with -b and without

fenner:
I did some coverage analysis on print-bgp.c, and noticed that "-b" is not exercised at all, nor is the "open" message printer.  These are some packets captured from a 4-byte ASN test.


fxlb:
Rebased, merged, thank you.

PR #748, opened 2019-04-14
Add oss-fuzz support (not yet enabled)

fenner:
This builds on the work done by @catenacyber in #700. I started an independent effort to add oss-fuzz support, then discovered #700, and this is the combination of both efforts.

My version calls into printers directly - I've implemented the plumbing for ether, ip, ip6 and bgp.  This allows targeted testing: "I wrote some code for a new bgp printer, let's fuzz it".  It creates the corpus (example good packets for the fuzzing infrastructure to learn from) by parsing test/*.pcap using Scapy.

The other version uses pcap files, so can end up in any part of the system.  It's good for "what can happen in the universe" but is harder to target.

Adding this plumbing to the tcpdump package is one half of what's needed; the other half is to enable clusterfuzz by submitting to the oss-fuzz repository.  Merging this pull request can be done independently, so that fuzzing on the clusterfuzz infrastructure can be enabled when desirable.

I've tried to document some of how it works at https://github.com/fenner/tcpdump/blob/fuzz/fuzz/README.md .

Closes #700.


catenacyber:
I hope that we can integrate this soon

fenner:
I rebased to get past the travis failure that was present in master.

Issue #637, opened 2017-10-05
permission denied on Solaris when '-w dump.pkt' option is used


Sashan:
running tcpdump on Solaris we get 'Permission denied', when process attempts to write captured packets to file specified by `-w dump.pkt` command line option:

````
root@solaris:~/#
root@solaris:~/# tcpdump -i net0 -w ~/dump.pkt
dropped privs to nobody
tcpdump: /root/dump.pkt: Permission denied
````
The fix we opted for is to let tcpdump to open `~/dump.pkt` file first and then switch from root to nobody. The patch below is against tcpdump-4.9.2 version.
````
--- a/tcpdump.c 2017-09-03 16:17:14.000000000 -0700
+++ b/tcpdump.c 2017-10-05 04:12:47.032285975 -0700
@@ -1786,82 +1786,6 @@
                (void)setsignal(SIGHUP, oldhandler);
 #endif /* _WIN32 */
 
-#ifndef _WIN32
-       /*
-        * If a user name was specified with "-Z", attempt to switch to
-        * that user's UID.  This would probably be used with sudo,
-        * to allow tcpdump to be run in a special restricted
-        * account (if you just want to allow users to open capture
-        * devices, and can't just give users that permission,
-        * you'd make tcpdump set-UID or set-GID).
-        *
-        * Tcpdump doesn't necessarily write only to one savefile;
-        * the general only way to allow a -Z instance to write to
-        * savefiles as the user under whose UID it's run, rather
-        * than as the user specified with -Z, would thus be to switch
-        * to the original user ID before opening a capture file and
-        * then switch back to the -Z user ID after opening the savefile.
-        * Switching to the -Z user ID only after opening the first
-        * savefile doesn't handle the general case.
-        */
-
-       if (getuid() == 0 || geteuid() == 0) {
-#ifdef HAVE_LIBCAP_NG
-               /* Initialize capng */
-               capng_clear(CAPNG_SELECT_BOTH);
-               if (username) {
-                       capng_updatev(
-                               CAPNG_ADD,
-                               CAPNG_PERMITTED | CAPNG_EFFECTIVE,
-                               CAP_SETUID,
-                               CAP_SETGID,
-                               -1);
-               }
-               if (chroot_dir) {
-                       capng_update(
-                               CAPNG_ADD,
-                               CAPNG_PERMITTED | CAPNG_EFFECTIVE,
-                               CAP_SYS_CHROOT
-                               );
-               }
-
[...]

PR #749, opened 2019-04-17
tcpdump should open file, where captured packets are going to be writ…

Sashan:
…ten to,

before it switches to user nobody.

patch has been submitted to upstream:
https://github.com/the-tcpdump-group/tcpdump/issues/637

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -2180,88 +2180,6 @@ main(int argc, char **argv)
 		(void)setsignal(SIGHUP, oldhandler);
 #endif /* _WIN32 */
 
-#ifndef _WIN32
-	/*
-	 * If a user name was specified with "-Z", attempt to switch to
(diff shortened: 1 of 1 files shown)


mcr:
The problem with this is that tcpdump opens files again with -C switches, and so really mechanisms were tcpdump has any privilege are broken.   Your patch is welcome, but we need to decide how to do privilege separation correctly: tcpdump has perhaps three or four levels of trust.

Sashan:
> 
> 
> The problem with this is that tcpdump opens files again with -C switches, and so really mechanisms were tcpdump has any privilege are broken. Your patch is welcome, but we need to decide how to do privilege separation correctly: tcpdump has perhaps three or four levels of trust.

I see. To be honest I have not looked at that deeply enough. I've just crafted the patch, which addresses the most common scenario for Solaris, which is mentioned in #637. I'll try to re-work my patch.

PR #768, opened 2019-06-11
babel: add supports for rfc6126bis

SawHad:
Update babel with rfc6126bis. See: [link](https://tools.ietf.org/wg/babel/).


print-babel.c
```diff
@@ -282,12 +297,11 @@ subtlvs_print(netdissect_options *ndo,
...
         }
-        if(cp == ep)
-            goto invalid;
+        if (cp == ep) goto invalid;
         sublen = GET_U_1(cp);
         cp++;
-        if(cp + sublen > ep)
-            goto invalid;
+        if ((MANDATORY_MASK & subtype) != 0) ND_PRINT(" (M)");
+        if(cp + sublen > ep) goto invalid;
```
mcr: This line should probably be a TCHECK macro.

mcr: Looks like yes, there are some new pcap files, thank you.

maiste: Thank you for your review. 
About the TCHECK macro, I don't see how we can use it. The only one which is suitable for is `ND_TCHECK_LEN` but it goes to trunc label (and not invalid label). Should we write a new one in netdissect.h or let the code like this ?

mcr:
We have more macros that will come out in 5.0, on the master branch.

+        if(cp + sublen > ep) goto invalid;

You need to know if the pointer "cp + sublen" is in the capture buffer, right?
ND_TCHECK_LEN(cp, sublen) does exactly that.

--
]   Michael Richardson, Sandelman Software Works        |    IoT architect   [

maiste:
Sorry, I misunderstood how ND_TCHECK_LEN worked. It's corrected. Thanks for your help.

fxlb:
Many changes are about putting `if` or `else` instructions on one line instead of two and have nothing to do with  the rfc6126bis support. They should be removed to focus the PR about rfc6126bis.

maiste:
I remove the if else corrections from the PR.

print-babel.c
guyharris: The ND_TTEST_ macros do the same tests as do the ND_TCHECK_ macros, but return a Boolean result (true if the test succeeds, false if the test fails), rather than testing the Boolean result and jumping to `trunc` if the test fails.

And the GET_ macros have the tests built in.

guyharris:
@fxlb, @mcr: are there any changes that still need to be made?

guyharris:
There are *some* changes that need to be made, because some builds are failing the "make check" stage; I haven't figure out what they are yet.

print-babel.c
```diff
@@ -42,38 +42,45 @@
...
+                           u_int length, u_short bodylen);
 
 void
 babel_print(netdissect_options *ndo,
             const u_char *cp, u_int length)
 {
+    u_short bodylen;
     ndo->ndo_protocol = "babel";
     ND_PRINT("babel");
-
```
guyharris, line 53: Please don't remove blank lines.

print-babel.c
```diff
@@ -42,38 +42,45 @@
...
         ND_PRINT(" %u", GET_U_1(cp + 1));
     }
-
     switch(GET_U_1(cp + 1)) {
     case 2:
-        babel_print_v2(ndo, cp, length);
+        if (length < 4)
+            goto invalid;
+        bodylen = GET_BE_U_2(cp + 2);
+        ND_PRINT(" (%u)", bodylen);
```
guyharris, line 67: This prints the length twice; that should be done only by babel_print_v2().  That's causing some tests to fail.

print-babel.c
```diff
@@ -89,17 +96,25 @@ babel_print(netdissect_options *ndo,
 #define MESSAGE_ROUTE_REQUEST 9
 #define MESSAGE_SEQNO_REQUEST 10
 #define MESSAGE_TSPC 11
-#define MESSAGE_HMAC 12
-#define MESSAGE_UPDATE_SRC_SPECIFIC 13 /* last appearance in draft-boutier-babel-source-specific-01 */
-#define MESSAGE_REQUEST_SRC_SPECIFIC 14 /* idem */
-#define MESSAGE_MH_REQUEST_SRC_SPECIFIC 15 /* idem */
+#define MESSAGE_HMAC_OLD 12
```
guyharris, line 99: That's still an HMAC TLV.  TLV 16 is a MAC TLV.

print-babel.c
```diff
@@ -280,14 +295,16 @@ subtlvs_print(netdissect_options *ndo,
         cp++;
         if(subtype == MESSAGE_SUB_PAD1) {
             ND_PRINT(" sub-pad1");
+            cp++;
```
guyharris, line 298: A PAD1 subTLV, like a PAD1 TLV, is only one byte - the TLV type field.  `cp` has already been incremented above; it should not be incremented again.

print-babel.c
```diff
@@ -732,6 +803,13 @@ babel_print_v2(netdissect_options *ndo,
         }
         i += len + 2;
     }
+
+    /* Packet trailer */
+    if (i >= bodylen && (length >= (u_int)bodylen+4) && (length-bodylen-4 != 0)) {
```
guyharris, line 808: This should be done by putting the TLV parsing into a separate routine, called once for the body and once for the trailer.  (There should probably be a flag indicating whether it's the body or the trailer, so that TLVs not allowed to be in the trailer should be reported as such.)

guyharris:
A modified version of this, as per my comments, has been committed as dc5b398458db155339754ba97ce9108b09318831.

PR #774, opened 2019-06-21
Support 2 new versions for arista ethertype and print the time in utc for the new versions

niks3089:
This change is [Reviewable](https://reviewable.io/reviews/the-tcpdump-group/tcpdump/774)


fxlb:
Merged, thank you!

PR #776, opened 2019-07-05
An assortment of PIMv2 packets for testing

fenner:
We have a decoder for DF Election packets which I will submit separately - this gives some coverage of PIM packet printers.


fxlb:
Merged, thanks !

PR #778, opened 2019-07-07
Decode the Multiple Labels capability (RFC8277)

fenner:
Decode the Multiple Labels capability message.  Note that the labeled-unicast printer in `decode_labeled_prefix4()` does not support multiple labels, so an update with multiple labels will be decoded as `(illegal prefix length)` as shown in the accompanying test pcap.


fxlb:
Merged, thanks !

Issue #569, opened 2016-11-30
Stop capturing after writing N bytes


kirawrath:
I would like a way to read `n` bytes with `tcpdump` and then exit.
Today we can use `-c` option, but it only counts packets, not bytes.
With the `-C` option I almost have what I want, the problem is that it does not exit after writing the specified size, but instead close that file and start to write a new one.
If I use it in conjunction with the option `-W 1` I still don't get what I need because it will then overwrite the beginning of my file.

This feature does not seem to be complex at all, and I feel many others felt frustrated like I did when I realized I could not accomplish this with `tcpdump` alone (I wrote a wrapper in c++ to overcome this).
I can write this feature myself, I just need to know if there is anything preventing it to be created.

stevekay:
Here's a hack that might suit in the short term.

````
[steve@localhost tcpdump]$ cat foo.sh
#!/bin/sh
pkill tcpdump
rm /tmp/foo.cap1
[steve@localhost tcpdump]$ sudo ./tcpdump -i ens33  -C 5 -W 5 -w /tmp/foo.cap -z ./foo.sh
tcpdump: listening on ens33, link-type EN10MB (Ethernet), capture size 262144 bytes
1916 packets captured
1956 packets received by filter
40 packets dropped by kernel
[steve@localhost tcpdump]$ ls -lh /tmp/foo.cap*
-rw-r--r--. 1 root root 4.8M Jan 18 22:43 /tmp/foo.cap0
[steve@localhost tcpdump]$

mattie47:
Thanks for that workaround @stevekay  👍

tarrenj:
Is this still the recommended way of doing this?  The man page says:

> -W
    Used in conjunction with the -C option, this will limit the number of files created to the specified number, and begin overwriting files from the beginning, thus creating a 'rotating' buffer. In addition, it will name the files with enough leading 0s to support the maximum number of files, allowing them to sort correctly. 
    Used in conjunction with the -G option, this will limit the number of rotated dump files that get created, exiting with status 0 when reaching the limit. 
    If used in conjunction with both -C and -G, the -W option will currently be ignored, and will only affect the file name. 

However I haven't been able to achieve this behavior...

PR #781, opened 2019-07-13
Add limit-capture-size option

DEADBEEFh:
Using --limit-capture-size, tcpdump can now exit after capturing
a specified amount of data.  If used with -w, tcpdump will exit
after writing the specified limit to disk.  Without -w, tcpdump
will exit reaching the limit based on packets*h->len.

Works with -C/-W/-G options.

The method for specifying file size for -C and
--limit-capture-size was updated to allow users to enter
num[k,m,g,ki,mi,gi].  Leaving off units will result in the current
behavior of an argument being multiplied by a default of 1000000.

This commit updates and improves upon work done by @stevekay for
feature request 97 and pull request 464.

### DIFF
--- a/tcpdump.1.in
+++ b/tcpdump.1.in
@@ -249,8 +249,12 @@ savefile and open a new one.  Savefiles after the first savefile will
 have the name specified with the
 .B \-w
 flag, with a number after it, starting at 1 and continuing upward.
-The units of \fIfile_size\fP are millions of bytes (1,000,000 bytes,
-not 1,048,576 bytes).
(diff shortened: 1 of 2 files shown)


DEADBEEFh:
Upon further review, there are a few minor cleanups I'd like to do:
-looks like I left an errant `errno=0` in the case statement
-probably don't need the `sizeflag` and can replace checks with `size_bytes != 0`

I'll wait for other comments and include those before committing.

DEADBEEFh:
Closes #569 and closes #97

systemcrash:
will this go in? @DEADBEEFh can you update the conflicts?

PR #792, opened 2019-08-22
add isis router capability TLV

ckishimo:
add isis router capability TLV per RFC 7981 section 2

```
	    unknown TLV #242, length: 8
	      0x0000:  c0a8 0001 0013 0100
```
```
	    IS-IS Router Capability TLV #242, length: 8
		Router-id: 192.168.0.1
		Flags: [none]
```


print-isoclns.c
```diff
@@ -3307,6 +3315,18 @@ isis_print(netdissect_options *ndo,
             ND_PRINT("\n\t      Sequence number: %u", GET_BE_U_4(tptr));
             break;
 
+        case ISIS_TLV_ROUTER_CAPABILITY:
+        if (tlen < 5)
+            break;
```
fxlb: Should not be silent in case of error, proposed update (also fix indent):
```patch
-        if (tlen < 5)
-            break;
+            if (tlen < 5) {
+                ND_PRINT(" [object length %u < 5]", tlen);
+                nd_print_invalid(ndo);
+                break;
+            }
```

fxlb:
Merged, thank you!

PR #795, opened 2019-09-03
add rsvp capability object rfc5063

ckishimo:
add rsvp capability object rfc5063 (section 4.2)

```
	RSVPv1 Hello Message (20), Flags: [Refresh reduction capable], length: 40, ttl: 1, checksum: 0x7d4d
	  Unknown Object (134) Flags: [ignore silently if unknown], Class-Type: Unknown (1), length: 8
	    0x0000:  0000 0018
```
```
	RSVPv1 Hello Message (20), Flags: [Refresh reduction capable], length: 40, ttl: 1, checksum: 0x7d4d
	  Capability Object (134) Flags: [ignore silently if unknown], Class-Type: 1 (1), length: 8
	    Flags: [none]
	    0x0000:  0000 0018
```


print-rsvp.c
```diff
@@ -1176,6 +1186,24 @@ rsvp_obj_print(netdissect_options *ndo,
...
+        case RSVP_OBJ_CAPABILITY:
+            switch(rsvp_obj_ctype) {
+            case RSVP_CTYPE_1:
+                if (obj_tlen < 4)
+                    return-1;
+                ND_PRINT("%s  Flags: [%s]",
+                       indent,
+                       bittok2str(rsvp_obj_capability_flag_values,
+                                  "none",
+                                  GET_BE_U_4(obj_tptr)));
```
fxlb: "Reserved bits MUST be set to zero on transmission and MUST be ignored on receipt." (rfc5063)
Thus a mask is needed.

fxlb:
Do you have pcap samples with some `capability_flag_values` set?

ckishimo:
@fxlb as I did not have any pcap samples I took the liberty of modifying the current pcap sample
Please let me know if there is any problem with that or the updated code. Thanks

print-rsvp.c
fenner: Protocol speakers must ignore reserved bits.  Protocol *dumpers* might want to show them for 2 reasons
1. to say "Hey this sender is misbehaving, it is setting reserved bits to nonzero"
2. when a new feature comes out it can be debugged without having to change the mask.
That being said, `bittok2str` doesn't output anything for bits that are set that are not in the input values array, so, I'm still not quite sure what the issue is.

fenner:
In your description you say

```
	    Flags: [none]
	    0x0000:  0000 0018
```

Why output the hex dump as well as the parsed flags?

print-rsvp.c
fxlb: > Protocol _dumpers_ might want to show them

Ok, proposed update
```patch
@@ -1193,11 +1193,15 @@ rsvp_obj_print(netdissect_options *ndo,
             case RSVP_CTYPE_1:
                 if (obj_tlen < 4)
                     return-1;
+                uint8_t unused_and_flags = GET_BE_U_4(obj_tptr);
+                if (unused_and_flags & ~RSVP_OBJ_CAPABILITY_FLAGS_MASK)
+                    ND_PRINT("%s  [reserved=0x%x must be zero]", indent,
+                        unused_and_flags & ~RSVP_OBJ_CAPABILITY_FLAGS_MASK);
                 ND_PRINT("%s  Flags: [%s]",
                        indent,
                        bittok2str(rsvp_obj_capability_flag_values,
                                   "none",
-                                  (GET_BE_U_4(obj_tptr) & RSVP_OBJ_CAPABILITY_FLAGS_MASK)));
+                                  (unused_and_flags & RSVP_OBJ_CAPABILITY_FLAGS_MASK)));
                 obj_tlen-=4;
                 obj_tptr+=4;
                 break;
```

fxlb:
>  I took the liberty of modifying the current pcap sample

OK.

print-rsvp.c
ckishimo: Just pushed the proposed changes. Thanks

ckishimo:
> Why output the hex dump as well as the parsed flags?
No particular reason.... In fact as you said the sender is misbehaving as this value should be zero...

fxlb:
Merged, Thanks.

Issue #763, opened 2019-05-09
buffer overread problem in the print_prefix function of print-hncp.c.


FIOpwK:
From: https://github.com/zyingp/temp/blob/master/tcpdump.md

[2018/11/26] Update: CVE-2018-19519 is assigned to the problem.

Set a breakpoint at the print_prefix function, and run tcpdump like tcpdump -ee -vv -nnr id_57.pcap (The special pcap file could be downloaded from https://github.com/zyingp/temp/blob/master/id_57.pcap) Then in the print_prefix function, the prefix variable's value is 0xff, max_length is 39, and the is_ipv4_mapped_address function return No. So the program enters the else-clause and calls decode_prefix6 (defined in print-bgp.c). However, the decode_prefix6 directly returns -1 (since the plen in it is 0xff) and does not fill any C string in the passed-in buf. Then after back to the print_prefix function, the last ND_PRINT("%s", buf) line in print_prefix may overread the 46-byte-long buffer buf (until meets a '\0' byte), since the buf variable is not initialized.

Solution: 
Initialize buf[0] to '\0' in 'print_prefix'.

buf[0] = '\0';

(Just as print-cnfp.c and print-bgp.c do.)

========

Any update on this?

guyharris:
The current tip of the master branch for tcpdump prints

    reading from file /Users/gharris/Downloads/id_57.pcap, link-type EN10MB (Ethernet), snapshot length 32983
    [Invalid header: len(16) < caplen(78)]
    tcpdump: pcap_loop: block total length in header and trailer don't match

If you're going to fuzz captures, fuzz the pcap/pcapng record information, and the packet data, ***SEPARATELY***.  Fuzzing the record information would be done to test libpcap's ability to handle invalid pcap or pcapng files.  Fuzzing the latter would be done to test tcpdump's ability to handle invalid packets.  Fixing bugs in libpcap found by fuzzing the record information may involve changing libpcap in such a way that it will reject the packet, so that programs such as tcpdump won't ever see the fuzzed packet data - this is why fuzzing both produces files that may not be useful for fuzzing tcpdump.

guyharris:
Resolved by commit 511915bef7e4de2f31b8d9f581b4a44b0cfbcf53.

FIOpwK:
Is the issue closed because you created a commit to 'fix' this ticket but are not planning on including the 'fix' in any release? So cherrypicking patches is the idea here?

fxlb:
[...]

PR #799, opened 2019-09-24
If decode_prefix6() returns a negative number, don't print buf.

H2O-2:
If it returns a negative number, it hasn't necessarily filled in buf, so
just return immediately; this is similar to the IPv4 code path, wherein
we just return a negative number, and print nothing, on an error.

This should fix GitHub issue #763.

(backported from commit 511915bef7e4de2f31b8d9f581b4a44b0cfbcf53)


fxlb:
New release with already this fix ongoing.

guyharris:
The fix is in 4.9.3 and in all subsequent releases, including the current one and all other 4.99.x releases.

PR #800, opened 2019-09-30
Fix order when printing PGM_OPT_REDIRECT_FIXED_LEN and opt_len

pmgdeb:
Fix order of PGM_OPT_REDIRECT_FIXED_LEN and opt_len to match the expression when printing in function print-pgm(), case PGM_OPT_REDIRECT.


fxlb:
Merged with commit message updated. Thank you !

PR #807, opened 2019-10-24
Use buffer_size instead of len in EVP_Cipher.

dalegr:
Input data to the EVP_Cipher must be padded with zeros and data size
must be multiple of a block_size [1,2].

For example, in the case of the AES 128 CBC, if inl argument of the
EVP_Cipher is len (not buffer_size), then the result (output_buffer)
will depend on architecture.

This patch does fix ikev2pI2 test on a ppc64le (tested on PowerNV 8247-22L).

[1] https://tools.ietf.org/html/rfc2406
[2] https://tools.ietf.org/html/rfc3602


guyharris:
The code is significantly changed in the master branch - for one thing, we don't call `EVP_Cipher()` any more, we call other routines.  We also require the block being decrypted to already have been padded as necessary.

So these changes 1) no longer apply and 2) are probably not necessary.

PR #762, opened 2019-05-03
Print EDNS options

cdeccio:
This allows the EDNS options to be displayed with `-vv`.

PR #808, opened 2019-10-24
Print EDNS options with data

jacobgb24:
Builds directly on #762 . Adds printing of EDNS options and their data with `-vv`. Format of EDNS options matches format of TCP options.


guyharris:
> Builds directly on #762 .

"Builds directly on" as in "#762 should be checked in first" or as in "replaces #762"?  This includes the same changes to `tests/TESTLIST` that #762 does.

jacobgb24:
It should probably replace #762 (by built I meant that I started from that branch and added commits to it thus obsoleting it). The changes to `tests/TESTLIST` are the same, but the output of the tests themselves are different.

fxlb:
"This branch cannot be rebased due to conflicts".
Closing ?

fxlb:
Rebase needed.

fxlb:
There is a conflict in tests/TESTLIST.
You should :
 `git rebase master` after updating your master branch with `the-tcpdump-group:master`

print-domain.c
```diff
@@ -267,6 +267,181 @@ ns_cprint(netdissect_options *ndo,
...
+static void
+print_eopt_ecs(netdissect_options *ndo, const u_char *cp,
+               u_int data_len)
+{
+    u_int family, addr_bits, src_len, scope_len;
+
+    u_char padded[32];
+    char addr[INET6_ADDRSTRLEN];
+
+    if (!ND_TTEST_2(cp)) {
```
fxlb: Is this test for truncated condition?
If yes, it is now useless because the new GET_ macros have embedded bounds checks, via get_ inline functions (see extract.h).

jacobgb24: The intent of the checks was to ensure that the EDNS option was properly formatted (e.g. ECS should have 2 bytes for family, 1 byte for source length, etc.). So I was marking the option invalid if it didn't have the proper number of bytes.

I don't know if there's a way to differentiate between a malformed option and a truncated message in this case. If you have any suggestions on that, let me know.

jacobgb24:
In terms of rebasing, I'm not super familiar with the process (I've generally avoided rebases). I pulled `the-tcpdump-group:master` into my `master` then did `git rebase master` which said `Current branch master is up to date.`. 

Sorry if that wasn't the correct process and let me know if there's something else I should do

print-domain.c
guyharris: > The intent of the checks was to ensure that the EDNS option was properly formatted (e.g. ECS should have 2 bytes for family, 1 byte for source length, etc.).

I.e., you're checking to make sure that the *option length* is large enough?

That should be done by checking the value of `data_len`.

That's not what `ND_TTEST_2()` does.  It checks whether there are at least two bytes in *the entire packet, as captured*, starting at the byte pointed to by `cp`; by default, that checks against the end of the captured part of the link-layer frame, although, as DNS runs atop UDP or TCP, which run atop IP, it will check against the end of the captured part of the IP datagram.  It doesn't check the remaining length of the option data; that'd be done by checking `data_len`.
 
> I don't know if there's a way to differentiate between a malformed option and a truncated message in this case. If you have any suggestions on that, let me know.

Check against `data_len` and print a message *different* from the one printed if the bounds checks embedded in the GET_ macros fail.  Calling `nd_print_invalid()` will do that, but you could also print "option too short" or something such as that.

fxlb: Someting like :
```
    if (data_len < @minimal_length@) {
        ND_PRINT("[length %u < @minimal_lengtht@]", data_len);
        nd_print_invalid(ndo);
        return;
    }
```
is what you want to do?

fxlb:
As you work in your master branch (not the best way I think), you could do:
git rebase @remote_for_the-tcpdump-group_repo@/master

print-domain.c
jacobgb24: Yeah that looks better. I'll make that change in a minute.

There's also two `ND_TTEST` checks in `eopt_print()` (`print-domain.c:344`). While those are redundant with the new `GET_` would it be best to leave those since the general flow is to pass `NULL` up to `domain_print()` which handles truncation there?

fxlb:
After removing the merge you have done with:
git reset --hard HEAD^
[first you could save your branch with : git checkout -b master-save; git checkout -]

fxlb:
@jacobgb24 
note for the future: your repo is forked from cdeccio/tcpdump, it should forked from the-tcpdump-group/tcpdump.

kimbo:
Would love to see this merged :+1:

fxlb:
Rebased in one commit, merged with 6bb25b4b4c84e5015bdcfcbd688dff512a20d971.
Updated by 285a6e7ab593d19e1e6f967ca7c14d7088537ce1.
Thank you !

PR #821, opened 2020-02-03
Testrun perl

mcr:
This code changes TESTrun to process more complex test cases by loading perl code defining an array.  This replaces the attempt to do this with YAML.pm, which would have required an additional package to run tests.


guyharris:
> This code changes TESTrun to process more complex test cases by loading perl code defining an array.

So the test descriptions are expressed in PON? :-)

guyharris:
Looks good.  Presumably if more than one thing needs to be defined to run the test, you could define it as "THIS && THAT" - presumably that will evaluate to `undef` if either `THIS` or `THAT` is undefined.

guyharris:
AppVeyor is being even more annoying than usual - it's stuck trying to fetch something that's "not a tree", and won't try something that *is* a tree, it appears - and it doesn't matter anyway whether the build succeeds, because we're not yet doing the tests on Windows, so just ignore the issues it's having with "switch from crypto.sh to crypto.tests".  Succeeding with Travis and Cirrus is sufficient.

mcr:
> > This code changes TESTrun to process more complex test cases by loading perl code defining an array.
> 
> So the test descriptions are expressed in PON? :-)

I guess so...?!

mcr:
> Looks good. Presumably if more than one thing needs to be defined to run the test, you could define it as "THIS && THAT" - presumably that will evaluate to `undef` if either `THIS` or `THAT` is undefined.

I plan to support config_set => ['CONFIG_THIS', 'CONFIG_THAT']

fxlb:
(Closed by error, re-open).

infrastation:
Have you considered making the "run/do not run" decisions in the Perl space instead of the declarations? This way you could just set the test's key to a boolean variable or insert/skip items in an `if` branch, thus avoiding the need to invent and implement a declarative way for the conditional tests.

infrastation:
I mean it would be straightforward just to assign a variable like `$test_crypto = $have_libcrypto && ! $have_capsicum;` and use it in the list of test like `'when' => $test_crypto,` instead of implementing an expression parser. And it would be nice to make it a few clean commits instead of exposing all the drafts.

infrastation:
For clarity, it would not be a big problem to use the proposed first implementation of conditional encoding (given it replaces the shell files). I have made a quick look through and a test run on FreeBSD (which finally works, which should enable `make check` in Cirrus CI later) and noticed the following debug output, which probably should not be in the final version:
```
FILE: /usr/home/denis/tcpdump/tests/crypto.tests
    esp1                                    : passed
    esp2                                    : passed
    esp3                                    : passed
    esp4                                    : passed
    esp5                                    : passed
    espudp1                                 : passed
    ikev2pI2                                : passed
    isakmp4                                 : passed
    bgp-as-path-oobr-ssl                    : passed
    bgp-aigp-oobr-ssl                       : passed
bgp-as-path-oobr-nossl ... skipped, HAVE_LIBCRYPTO is set
bgp-aigp-oobr-nossl ... skipped, HAVE_LIBCRYPTO is set
FILE: /usr/home/denis/tcpdump/tests/isis-seg-fault-1-v.tests
isis-seg-fault-1-v ... skipped, no USING_GCC
FILE: /usr/home/denis/tcpdump/tests/lmp-v.tests
lmp-v ... skipped, no USING_GCC
------------------------------------------------
   0 tests failed
 512 tests passed
```
Will make another round tomorrow.

mcr:
> I mean it would be straightforward just to assign a variable like `$test_crypto = $have_libcrypto && ! $have_capsicum;` and use it in the list of test like `'when' => $test_crypto,` instead of implementing an expression parser. And it would be nice to make it a few clean commits instead of exposing all the drafts.

I really don't want to write code in configuration files for tests.
I would prefer to keep the test configuration declarative, which is why I chose YAML at first.
I will be happy to rebase/merge things.

mcr:
> FILE: /usr/home/denis/tcpdump/tests/crypto.tests

Is this the debug that you feel should be omitted?

infrastation:
Thank you. If the other non-indented lines are not debug, it would be nice to print them using the same format (can be fixed later though). I am going to get back to this later today.

mcr:
> Thank you. If the other non-indented lines are not debug, it would be nice to print them using the same format (can be fixed later though). I am going to get back to this later today.

The FILE.. isn't exactly a debug, it is more of a section header.  I think it needs to be there to know where the tests are coming from.  I could prefix the test name with the base name of the tests if you prefer, but I think that would take more space.
So it would look like:
   TESTLIST/foo
   TESTLIST/bar
   crypto/esp4
   lmp-v/lmp-v

which I don't think will look as nice.  All easily changeable though.

fxlb:
Keeping FILE section header looks good for me.

I suggest:
```
    bgp-aigp-oobr-ssl                   : passed
    bgp-as-path-oobr-nossl              : skipped (HAVE_LIBCRYPTO is set)
```

mcr:
fxlb <notifications@github.com> wrote:
    > Keeping FILE section header looks good for me.

    > I suggest:
    > ```
    > bgp-aigp-oobr-ssl                   : passed
    > bgp-as-path-oobr-nossl              : skipped (HAVE_LIBCRYPTO is set)
    > ```

Done.

tests/TESTrun
```diff
@@ -107,12 +107,12 @@ sub runOneComplexTest {
...
 
     if(!$foundit) {
-        print "${name} ... skipped, no ${configset}\n";
+        printf "    %-40s: skipped (no %s)\n", $name, $configset;
         return 0;
     }
 
     if(!$unfoundit) {
-        print "${name} ... skipped, ${configunset} is set\n";
+        printf "    %-40s: skipped (no %s)\n", $name, $configunset;
```
fxlb: Is this not?
`s/skipped (no %s)/skipped (%s)/`

PR #829, opened 2020-02-21
Parse the IS-IS Instance Identifier TLV from RFC8202

fenner:
RFC8202 replaces the variable-length neighbor TLV with an Instance Identifier TLV.


fxlb:
Thank you!

PR #830, opened 2020-02-21
Dispatch the PTP ethertype to ptp_print()

fenner:
This implements "IEEE 802.3/Ethernet" mode as described in annex F of the PTP standard.  This uses the same format packets, but directly over Ethernet instead of encapsulated in UDP.


fxlb:
> This implements "IEEE 802.3/Ethernet" mode as described in annex F of the PTP standard. This uses the same format packets, but directly over Ethernet instead of encapsulated in UDP.

I suggest your comment be added in the commit message body.

print-ether.c
```diff
@@ -100,6 +100,7 @@ const struct tok ethertype_values[] = {
     { ETHERTYPE_GEONET,         "GeoNet"},
     { ETHERTYPE_CALM_FAST,      "CALM FAST"},
     { ETHERTYPE_AOE,            "AoE" },
+    { ETHERTYPE_PTP,            "PTP" },
```
fxlb: Should we print "PTPv2" ?

fxlb:
Please, rebase on top of the-tcpdump-group:master (there were some recent fixes on print-ptp.c).
Also tests/ptp_ethernet.out will require an update.

print-ether.c
fenner: IMO, no: the ptp_print() will print PTPv2.  And we don't know that PTPv3 won't use the same ethertype.

fenner:
> > This implements "IEEE 802.3/Ethernet" mode as described in annex F of the PTP standard. This uses the same format packets, but directly over Ethernet instead of encapsulated in UDP.
> 
> I suggest your comment be added in the commit message body.

ok.

print-ether.c
fenner: 16:47:07.869101 74:83:ef:01:ac:5b > 01:1b:19:00:00:00, ethertype PTP (0x88f7), length 60: ***PTPv2***, v1 compat : no, msg type : sync msg, length : 44, domain : 0, reserved1 : 0, Flags [two step], NS correction : 0, sub NS correction : 0, reserved2 : 0, clock identity : 0x7483efffff01ac16, port id : 274, seq id : 0, control : 0 (sync msg), log message interval : 0, originTimeStamp : 0 seconds, 0 nanoseconds

fxlb: Agreed.

fxlb:
Thank you!

PR #833, opened 2020-02-27
MPTCP: parse v1 ADD_ADDR option

pkrystad:
RFC6824bis-18 (MPTCP version 1) makes changes to the MPTCP
ADD_ADDR option format.

An HMAC field is added to the option and the IP version field is
removed and replaced with an Echo bit that provides a reliability
mechanism.

This change allows either v0 or v1 ADD_ADDR options
to be parsed and displayed correctly.

Signed-off-by: Peter Krystad <peter.krystad@linux.intel.com>


fxlb:
Merged, thank you!

PR #836, opened 2020-02-29
Create main.yml

radenkaclak:


guyharris:
Which CI system is this intended for?  We already have .appveyor.yml, .cirrus.yml, and .travis.yml.

guyharris:
> Which CI system is this intended for? We already have .appveyor.yml, .cirrus.yml, and .travis.yml.

[Ansible](https://www.ansible.com/use-cases/continuous-delivery)?  If so, this looks like a template or a stub; why not just check in what's intended to be the final version?

fxlb:
Account deleted, closing.

PR #838, opened 2020-02-29
move packets_captured out of arguments, and into netdissect_options

mcr:
This minor bit of cleanup moves the packets_captured variable into the ndo structure.
Unfortunately, this requires that the ndo structure be available globally for the signal handler to be able to call info().  It's not clear that we should really be doing that much work in a signal handler, but we do that already, so I did not change that.
Fundamentally, I made this change so make it easier to call pretty_print_packet() from other contexts (in patches still coming).

### DIFF
--- a/netdissect.h
+++ b/netdissect.h
@@ -221,7 +221,8 @@ struct netdissect_options {
   const char *ndo_protocol;	/* protocol */
   jmp_buf ndo_truncated;	/* jmp_buf for setjmp()/longjmp() */
   void *ndo_last_mem_p;		/* pointer to the last allocated memory chunk */
-  int ndo_packet_number;	/* print a packet number in the beginning of line */
+  int ndo_packet_number;	/* flag: print a packet number in the beginning of line */
+  int ndo_packets_captured;     /* current count of captured packets */
   int ndo_suppress_default_print; /* don't use default_print() for unknown packet types */
   int ndo_tstamp_precision;	/* requested time stamp precision */
   const char *program_name;	/* Name of the program using the library */
--- a/print.c
+++ b/print.c
@@ -397,14 +397,15 @@ get_if_printer(netdissect_options *ndo, int type)
 }
 
 void
-pretty_print_packet(netdissect_options *ndo, const struct pcap_pkthdr *h,
-		    const u_char *sp, u_int packets_captured)
+pretty_print_packet(netdissect_options *ndo,
+                    const struct pcap_pkthdr *h,
+                    const u_char       *sp)
 {
 	u_int hdrlen;
 	int invalid_header = 0;
 
 	if (ndo->ndo_packet_number)
-		ND_PRINT("%5u  ", packets_captured);
+		ND_PRINT("%5u  ", ndo->ndo_packets_captured);
 
 	/* Sanity checks on packet length / capture length */
 	if (h->caplen == 0) {
--- a/print.h
+++ b/print.h
@@ -34,9 +34,9 @@ int	has_printer(int type);
 
 if_printer_t get_if_printer(netdissect_options *ndo, int type);
 
-void	pretty_print_packet(netdissect_options *ndo,
-	    const struct pcap_pkthdr *h, const u_char *sp,
-	    u_int packets_captured);
+void    pretty_print_packet(netdissect_options *ndo,
+                            const struct pcap_pkthdr *h,
+                            const u_char       *sp);
 
 void	ndo_set_function_pointers(netdissect_options *ndo);
 
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -271,8 +271,7 @@ static void flushpcap(int);
   static void verbose_stats_dump(int sig);
 #endif /* _WIN32 */
 
-static void info(int);
-static u_int packets_captured;
+static void info(netdissect_options *ndo, int);
 
 #ifdef HAVE_PCAP_FINDALLDEVS
 static const struct tok status_flags[] = {
@@ -1439,6 +1438,9 @@ open_interface(const char *device, netdissect_options *ndo, char *ebuf)
 	return (pc);
 }
 
+/* global because signal handlers */
+netdissect_options Ndo;
+
 int
 main(int argc, char **argv)
 {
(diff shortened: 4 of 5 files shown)


netdissect.h
```diff
@@ -221,7 +221,8 @@ struct netdissect_options {
   const char *ndo_protocol;	/* protocol */
   jmp_buf ndo_truncated;	/* jmp_buf for setjmp()/longjmp() */
   void *ndo_last_mem_p;		/* pointer to the last allocated memory chunk */
-  int ndo_packet_number;	/* print a packet number in the beginning of line */
+  int ndo_packet_number;	/* flag: print a packet number in the beginning of line */
+  int ndo_packets_captured;     /* current packet number */
```
guyharris: "current packet number" could be read as "current number of packets" or "number of current packet"; in a sense, the two are the same (if tcpdump has captured N packets, the current packet is the Nth), but "current count of captured packets" might be a little clearer.

print.c
```diff
@@ -397,14 +397,15 @@ get_if_printer(netdissect_options *ndo, int type)
 }
 
 void
-pretty_print_packet(netdissect_options *ndo, const struct pcap_pkthdr *h,
-		    const u_char *sp, u_int packets_captured)
+pretty_print_packet(netdissect_options *ndo,
+                    const struct pcap_pkthdr *h,
+		    const u_char       *sp)
```
guyharris: Are these indented the same amount?

guyharris:
> This minor bit of cleanup moves the packets_captured variable into the ndo structure.
> Unfortunately, this requires that the ndo structure be available globally for the signal handler to be able to call info(). It's not clear that we should really be doing that much work in a signal handler, but we do that already, so I did not change that.
> Fundamentally, I made this change so make it easier to call pretty_print_packet() from other contexts (in patches still coming).

So this is useful even if we change the code *not* to call `info()` in a signal handler, e.g. if we use a loop with `pcap_dispatch()`, have the signal handler call `pcap_breakloop()` and set a "please print statistics" flag in Ndo, and, after `pcap_dispatch()` returns in the loop, check that flag and, if set, clear it and call `info()`?

guyharris:
> > Fundamentally, I made this change so make it easier to call pretty_print_packet() from other contexts (in patches still coming).
> 
> So this is useful even if we change the code _not_ to call `info()` in a signal handler, e.g. if we use a loop with `pcap_dispatch()`, have the signal handler call `pcap_breakloop()` and set a "please print statistics" flag, and, after `pcap_dispatch()` returns in the loop, check that flag and, if set, clear it and call `info()`?

AAAAAAAAAAAAAAAAAAAAAAAAAArgh `pcap_breakoop()` can lose the race:

`pcap_dispatch()`/`pcap_loop()`/etc. is in the middle of the read routine and has already checked the "break out of the loop" flag internal to pcap, and it's not set.
Signal comes in at time T1; the handler sets the "break out of the loop" flag, and returns.
The read routine continues, lands in the read/receive/whatever call to read more packets, and blocks.  It's *not* interrupted out of that, because the signal has already happened.

This means that, with my "print the ^T information in the capture loop" suggestion, the message might not be printed until the buffer timeout expires, which could be as much as 1 second.  This is noticeable.

I think there are tricks I can do to fix that (involving temporarily turning non-blocking mode on in `pcap_breakloop()`, in cases where it's necessary), but it's probably not appropriate to rely on it.  Perhaps only doing that with libpcap 1.10 or later, when it comes out, will do it.

fxlb:
> Unfortunately, this requires that the ndo structure be available globally
In 2015 Brooks Davis works to disentangle libnetdissect from tcpdump to allows
libnetdissect to be more easily sandboxed. (see 115a06358970e16e772da97df73a39283a9fd77e).
One of the result was to remove the global ndo.
Should have ndo structure be available globally again a regression in security?

mcr:
> > Fundamentally, I made this change so make it easier to call pretty_print_packet() from other contexts (in patches still coming).
> 
> So this is useful even if we change the code _not_ to call `info()` in a signal handler, e.g. if we use a loop with `pcap_dispatch()`, have the signal handler call `pcap_breakloop()` and set a "please print statistics" flag, and, after `pcap_dispatch()` returns in the loop, check that flag and, if set, clear it and call `info()`?

I can do this work, noted in #840. Clearly we need a pcap_dispatch()/breakloop that works better.
I suggest we do this as separate work.
I agree with the dis-intanglement work, and I am trying to take this further.

netdissect.h
mcr: Changed the comment in a new pull request.

print.c
mcr: Begon EVIL TAB.

mcr:
> > > Fundamentally, I made this change so make it easier to call pretty_print_packet() from other contexts (in patches still coming).
> > 
> > 
> > So this is useful even if we change the code _not_ to call `info()` in a signal handler, e.g. if we use a loop with `pcap_dispatch()`, have the signal handler call `pcap_breakloop()` and set a "please print statistics" flag, and, after `pcap_dispatch()` returns in the loop, check that flag and, if set, clear it and call `info()`?
> 
> AAAAAAAAAAAAAAAAAAAAAAAAAArgh `pcap_breakoop()` can lose the race:

Where did you propose that it be called?
We need a handle, and if we want to avoid global handles for pcap_t * into the signal handler, then the signal handler needs to set a single global flag, and we need to breakloop() in the application (callback) code.
Does this change the race condition?

> This means that, with my "print the ^T information in the capture loop" suggestion, the message might not be printed until the buffer timeout expires, which could be as much as 1 second. This is noticeable.

Annoying.

> I think there are tricks I can do to fix that (involving temporarily turning non-blocking mode on in `pcap_breakloop()`, in cases where it's necessary), but it's probably not appropriate to rely on it. Perhaps only doing that with libpcap 1.10 or later, when it comes out, will do it.

I'm okay with this.  Let's move this to #840 .

infrastation:
It is still not clear what the practical use case for these changes is, and why it involves changing `TESTrun`. From the pull request comments and the remaining indentation discrepancies it seems these changes are not ready for merging. Unless anyone is willing to address these issues, this pull request will be closed in 14 days and the branch will be moved to the incomplete works repository.

infrastation:
It is time. The draft branch has been moved to the-tcpdump-group/tcpdump-incomplete-works/mcr-move-packets-captured-to-ndo.

PR #844, opened 2020-04-03
add support for IS-IS Segment routing (Prefix/Node SIDs) as per RFC 8667

hannesgredler:
This is is to support for Node and Prefix SIDs in IS-IS.
Also add support for a general subTLV parser to the router-CAP TLV
Please review also the nested subTLV logic for buffer overruns.
Thinks should be relatively robust, but second ball of eyes never hurts ;

thanks,

/hannes


fxlb:
There is a problem: This PR must be on top of  the-tcpdump-group:master. Github shows 13 commits mixing with commits of PR 843 and display "This branch cannot be rebased due to conflicts".

hannesgredler:
submitted a fresh one.

PR #850, opened 2020-05-08
Set default linktype LINUX_SLL2 when capturing on the "any" device

fxlb:
[Request for comments]

I propose to set the default linktype to LINUX_SLL2 when capturing on the "any" device.
(if available in libpcap)

It gives useful informations: interface name, in/out with '-e' option.

libpcap support since 1.9.1 version.
The printing is currently not supported in release 4.9.3.

Already asked on the mailing list.
No one said "no".

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -1932,6 +1932,14 @@ main(int argc, char **argv)
 		show_remote_devices_and_exit();
 #endif
 
+#if defined(DLT_LINUX_SLL2) && defined(HAVE_PCAP_SET_DATALINK)
+/* Set default linktype DLT_LINUX_SLL2 when capturing on the "any" device */
+		if (device != NULL &&
+		    strncmp (device, "any", strlen("any")) == 0
+		    && yflag_dlt == -1)
+			yflag_dlt = DLT_LINUX_SLL2;
+#endif
+
 	switch (ndo->ndo_tflag) {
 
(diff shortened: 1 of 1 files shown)


fxlb:
cc @pevik

guyharris:
Why not just make it the default in libpcap?  Tcpdump and Wireshark both have an option to override the default - -y on the command line, and a combo box in the GUI.

fxlb:
Why not ?
Could other project not supporting LINUX_SLL2 be broken ?

guyharris:
> Could other project not supporting LINUX_SLL2 be broken ?

If those projects work like tcpdump, they will immediately fail as soon as they see a DLT_ value they don't support, and will allow one of the alternate link-layer types to be selected when capturing.

If they don't, they should be fixed to do so.  (And, yes, that includes Wireshark.  I'll see whether the "immediately fail" part for TShark - and whether it will even refuse to *offer* LINUX_SLL2 via the GUI - happen, and fix it if not.  The next step is to add LINUX_SLL2 support.)

fxlb:
My idea was to use LINUX_SLL2  by default.
With this PR or with a libpcap update, I don't mind if this does not create new problems.

pevik:
@fxlb thanks! I'm glad now the work has been completed, because not many people has been using `LINUX_SLL2` as it wasn't the default.

PR #851, opened 2020-05-09
[RFC] Print packets for unsupported link protocols in hexadecimal and ASCII

fxlb:
This avoids to get only:
    tcpdump: packet printing is not supported for link type XYZ: use -w
    
    Set ndo->ndo_packet_number.
    
    The default printing is like:
        1  15:00:02.232595 UNSUPPORTED
            0x0000:  0000 0338 0000 0000 0000 0000 0000 0000
            0x0010:  1400 0000 1600 0503 0000 0000 0000 0000
            0x0020:  0200 0000
        2  15:00:02.232622 UNSUPPORTED
            0x0000:  0000 0338 0000 0000 0000 0000 0000 0000
            0x0010:  3000 0000 1400 0200 0000 0000 0000 0000
            0x0020:  0208 0000 0000 0000 0800 0200 7f00 0001
            [...]

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1077,6 +1077,7 @@ set(NETDISSECT_SOURCE_LIST_C
     print-token.c
     print-udld.c
     print-udp.c
+    print-unsupported.c
     print-usb.c
     print-vjc.c
     print-vqp.c
--- a/Makefile.in
+++ b/Makefile.in
@@ -233,6 +233,7 @@ LIBNETDISSECT_SRC=\
 	print-token.c \
 	print-udld.c \
 	print-udp.c \
+	print-unsupported.c \
 	print-usb.c \
 	print-vjc.c \
 	print-vqp.c \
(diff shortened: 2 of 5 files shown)


guyharris:
+1 - Wireshark doesn't completely give up on packet types it doesn't understand; it doesn't *dissect* them, obviously, but it does show them, complete with the hex/ASCII dump pane, and TShark prints them, with the hex/ASCII dump.

Is there a reason to print the packet number for packets of an unknown link-layer type regardless of whether the `-#` flag was supplied or not?

fxlb:
> Is there a reason to print the packet number for packets of an unknown link-layer type regardless of whether the -# flag was supplied or not?

This allows to better identify packets without giving -#.
I can remove that if it seems confusing.

fxlb:
I think I will replace print_unknown_data() call by hex_and_ascii_print() call to print also in ASCII.

fxlb:
> I think I will replace print_unknown_data() call by hex_and_ascii_print() call to print also in ASCII.

Done.

fxlb:
Removed "print the packet number" by default.

fxlb:
Merged.

PR #854, opened 2020-05-22
Fix BGP Route Target Membership NLRI prefix decoding

### LINKED ISSUE #853: Regression in BGP SAFI 132 route target printing
In tcpdump 4.9.3, SAFI 132 NLRI are printed like:
```
          Multi-Protocol Reach NLRI (14), length: 14, Flags [O]:
            AFI: IPv4 (1), vendor specific SAFI: Route Target Routing Information (132)
            nexthop: 1.0.0.2, nh-length: 4, no SNPA
              origin AS: 22, route target 0:0 (= 0.0.0.0)
```
In master, some refactoring lead to using a print function for data in the packet against a variable allocated on the heap (which leads to incorrect truncation checks depending on memory layout), and requiring that the prefix length be 96 (the full value):
```
          Multi-Protocol Reach NLRI (14), length: 14, Flags [O]:
            AFI: IPv4 (1), vendor specific SAFI: Route Target Routing Information (132)
            nexthop: 1.0.0.2, nh-length: 4, no SNPA
            (illegal prefix length)
```

The comment at the top of the function correctly points out that `valid lengths are { 0, 32, 33, ..., 96 } bits` but later the incorrect comment `An extended community is 8 octets, so the remaining prefix length must be 64` precedes the check that results in the "(illegal prefix length)" here.

I have a patch developed by a colleague at Arista at https://github.com/fenner/tcpdump/commits/rfc4684 but have not created a pull request because I think it requires some minor cleanups.  However, I heard rustlings that a release might be coming soon, and you may not want to release with this regression.

fenner:
As described in RFC4684, prefix lengths of 0 and
32-96 are all valid.  Introduce bgp_rt_prefix_print()
to print partial route targets.

Fixes #853


fenner:
I think the only problems with this patch are variable names and whitespace usage.  I wanted to improve it before creating the pull request but don't think that's critical.

fenner:
Travis says failure - this is only because 22 arm64 builds failed because of overload in the system - said "retry later".  2 arm64 builds did succeed, so there is at least proof that it builds on arm64.

guyharris:
> Travis says failure - this is only because 22 arm64 builds failed because of overload in the system - said "retry later". 2 arm64 builds did succeed, so there is at least proof that it builds on arm64.

When all else fails, manually restart the builds.

A bunch of the builds are whiling about sudo being unable to resolve some host name (probably the host name of the VM running the build), so Travis are apparently having some issues, but those builds appear to be proceeding anyway.  I know they had, at least one point, a page to show the overall status of Travis, but it's no longer as easy to find as it used to be.

guyharris:
> I know they had, at least one point, a page to show the overall status of Travis, but it's no longer as easy to find as it used to be.

Click "Help" at the top of the page and, on the help page, click the "* Travis CI: {status}" link.

guyharris:
> Click "Help" at the top of the page and, on the help page, click the "* Travis CI: {status}" link.

There's a small break in the graphs somewhere in the early morning of 2020-05-23, so perhaps there was a glitch.

guyharris:
> Travis says failure - this is only because 22 arm64 builds failed because of overload in the system - said "retry later". 2 arm64 builds did succeed, so there is at least proof that it builds on arm64.

After restarting, Travis now reports success.  (AppVeyor has a "restart incomplete" button that will restart the whole process if some build failed due to CI platform problems.)

PR #867, opened 2020-08-04
Add support for ethernet IP protocol

ahsalam:
The SRv6 network programming defines a new IP protocol number [1].

The new protocol number is used in the Next Header field of an IPv6 header
or any extension header indicates that the payload is an Ethernet.

IANA has assigned 143 as the protocol number of ethernet [2].

The SRv6 packet has the following format:
IPv6 HDR + SRH + Ethernet HDR + IPv4/IPv6 packet

This patch allows TCPDUMP to dissect such SRv6 packets.

[1] https://tools.ietf.org/html/draft-ietf-spring-srv6-network-programming-16#section-9.1
[2] https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

Signed-off-by: Ahmed Abdelsalam <ahabdels@gmail.com>


ahsalam:
Any comments on the PR ?

fxlb:
Please, first rebase on top on master (9 new commits) and remove the FALSE parameter from ether_print() call (it was removed).

fxlb:
You also should add a test case with '-e' option like to show Ethernet addresses.
```
ipv6-srh-ipproto-ether-ev ipv6-srh-ipproto-ether.pcap ipv6-srh-ipproto-ether.out-ev -ev
```

ipproto.h
```diff
@@ -140,3 +140,6 @@ extern const char *netdb_protoname (const uint8_t);
 #ifndef IPPROTO_MOBILITY
 #define IPPROTO_MOBILITY	135
 #endif
+#ifndef IPPROTO_ETHERNET
+#define IPPROTO_ETHERNET	143
```
fxlb, line 144: Should add a comment "TEMPORARY - registered 2020-01-31, expires 2021-01-31"

print-ip-demux.c
```diff
@@ -208,6 +208,10 @@ ip_demux_print(netdissect_options *ndo,
 		pgm_print(ndo, bp, length, iph);
 		break;
 
+	case IPPROTO_ETHERNET:
+		ether_print(ndo, bp, length, ND_BYTES_AVAILABLE_AFTER(bp), NULL, NULL, FALSE);
```
fxlb, line 212: IP version can be 4 or 6 ?

ipproto.h
ahsalam: Ok. will add it in the updated PR.

print-ip-demux.c
ahsalam: No. IPPROTO_ETHERNET indicates that the traffic carried inside the SRv6 tunnel is L2 frame. So it doesn't matter if it is IPv4 or IPv6.

fxlb: Thus we can have:
Ether/IPv6/Ether/IPv[46]
or
Ether/IPv4/Ether/IPv[46] ?
or only 
Ether/IPv6/Ether/IPv[46]

fxlb: In other words, should we have?
```patch
diff --git a/ipproto.c b/ipproto.c
index 58ceeeba..a3b0714a 100644
--- a/ipproto.c
+++ b/ipproto.c
@@ -50,6 +50,7 @@ const struct tok ipproto_values[] = {
     { IPPROTO_PGM, "PGM" },
     { IPPROTO_SCTP, "SCTP" },
     { IPPROTO_MOBILITY, "Mobility" },
+    { IPPROTO_ETHERNET, "Ethernet" },
     { 0, NULL }
 };
 
diff --git a/print-ip-demux.c b/print-ip-demux.c
index 308529b8..8b53187c 100644
--- a/print-ip-demux.c
+++ b/print-ip-demux.c
@@ -209,7 +209,13 @@ again:
 		break;
 
 	case IPPROTO_ETHERNET:
-		ether_print(ndo, bp, length, ND_BYTES_AVAILABLE_AFTER(bp), NULL, NULL, FALSE);
+		if (ver == 6)
+			ether_print(ndo, bp, length, ND_BYTES_AVAILABLE_AFTER(bp), NULL, NULL);
+		else {
+			ND_PRINT("[%s requires IPv6]",
+				 tok2str(ipproto_values,"unknown",nh));
+			nd_print_invalid(ndo);
+		}
 		break;
 
 	case IPPROTO_NONE:
```

ahsalam: We can have only this case. 
**Ether/IPv6/Ether/IPv[46]**

ahsalam: > Thus we can have:
> Ether/IPv6/Ether/IPv[46]
> or
> Ether/IPv4/Ether/IPv[46] ?
> or only
> Ether/IPv6/Ether/IPv[46] ?

We can have only this case.
**Ether/IPv6/Ether/IPv[46]**

fxlb: OK, thus the above patch is needed.

ahsalam: Ok. will include it in the update PR.

ahsalam:
PR updated.

fxlb:
Thank you!

PR #874, opened 2020-09-18
Print LSP ping timestamps as NTP time, not UNIX time

### LINKED ISSUE #873: LSP Ping timestamps are NTP time, but printed using ts_print()
RFC4379 is a little confusing, since sometimes it refers to "seconds and microseconds" and sometimes to NTP time.  However, https://www.rfc-editor.org/errata/eid1714 makes it clear that it's NTP time.  print-lspping.c should be decoding the time into a `struct l_fixedpt` and using `p_ntp_time()`.

fenner:
During development of LSP ping, the embedded timestamps were apparently changed from struct timeval to NTP time, but the printer has not been updated.

Fixes #873 

(Sorry for filing an issue and then a pull request, I didn't think I was going to fix it but wanted to record the problem)


infrastation:
Thank you. As a generic remark, the reasoning behind the change would serve the purpose much better in the commit message, where it can survive regardless of github.com.

infrastation:
It looks like every CI build that runs `make check` fails, if the only reason for that is because the older LSP ping test cases need to be updated too, the solution is straightforward. However, please note that the older tests now have their printed timestamps in year 1934, is it correct? Besides that, perhaps you would want to base this change on the latest master branch and choose it as the destination.

fenner:
Probably the right thing is to add a heuristic, if the year is before 2000 then assume it's the "unix epoch and milliseconds" style.  That will last until 2070.

infrastation:
Do you mean these older captures contain packets that implement a draft that is different from the final specification? These .pcap files originate from commit 62fbfa3f in 2004. I can patch them the same way as in commit bc3bc62f if that helps.

fenner:
> Do you mean these older captures contain packets that implement a draft that is different from the final specification?

Yes, I believe the timestamp change came from late in the development process.

I have no idea whether or not there are implementations in the wild that have this behavior.  It will break in 2036 when NTP time rolls over, though, so maybe it's not a good idea.

fenner:
> It looks like every CI build that runs `make check` fails

It looks like all of the builds using CMAKE + BUILD_LIBPCAP are failing too, due to brcm tests.

infrastation:
On a second thought about the timestamp encoding, the function of a packet analyzer is to tell precisely what the bytes on the wire mean, so if something is sending packets with an incorrect (as far as the current spec defines it) encoding, the right thing to do would be just to print "this packet says the timestamp is this (1934)" and let the human apply any interpretation. Heuristics could be appropriate in the protocol implementation to aid in a transition from draft to draft, if such a problem still exists. Does it make sense?

CMAKE + BUILD_LIBPCAP failures may be related to the problem I had described on tcpdump-workers about a week ago.

fenner:
> On a second thought about the timestamp encoding, the function of a packet analyzer is to tell precisely what the bytes on the wire mean, so if something is sending packets with an incorrect (as far as the current spec defines it) encoding, the right thing to do would be just to print "this packet says the timestamp is this (1934)" and let the human apply any interpretation. Heuristics could be appropriate in the protocol implementation to aid in a transition from draft to draft, if such a problem still exists. Does it make sense?

I've come to the same conclusion and will remove the heuristic.  I'll regenerate the failing test output with the encoded timestamp - after all, it's what the packets say - and squash and repost.

infrastation:
Thank you for the update. Year 1934 in NTP encoding seems to work on Solaris 10 SPARC too, so it will not be necessary to modify the old .pcap files to make `make check` work there (as was necessary in commit bc3bc62f).

One more thing: `if ((int_part != 0) && (fraction != 0))` should use a logical "or" because in its current form it will print both 12345.0 and 0.12345 as "no timestamp".

When the commit of interest is ready, I am going to cherry-pick it into the master branch (w/o the change that removes the currently failing configurations from the matrix -- that could be "allowed to fail" for now, but the root cause must be fixed eventually).

infrastation:
I have cherry-picked the substantial commit from this pull request into the master branch verbatim, and fixed the time testing condition in a separate commit with comments (this code had this bug way before your change). Thank you for the contribution!

PR #875, opened 2020-09-20
remove redundant NT_TCHECK, let GET_ routines handle checks

mcr:
This is my first attempt at removing ND_TCHECK macros.
I notice that the output prints slightly less in the tests.
Please validate that I'm doing the right thing.


infrastation (changes requested):
1. Most Linux, FreeBSD and macOS builds fail because of the numerous `[-Wunused-label]` and `[-Wunused-parameter]` warnings (`touch .devel`).
2. If a test output changes after removing an `ND_TCHECK_LEN()`, the expected difference is that the amount of printed detail increases (because removing a long `ND_TCHECK_LEN()` before a series of short `GET_()` gives the latter a chance to decode the remainder of the packet buffer). When the opposite effect is in place, something is likely wrong.
3. Removing an `ND_TCHECK_LEN()` before a conditional fetch (such as `GET_ETHERADDR_STRING()`) is likely to introduce a conditional OOBR.
4. Removing checks of the form `if (length < something) goto trunc;` is likely wrong, because often they are poor man's `goto invalid`, which ideally should result in flagging the current structure as invalid and trying to decode the next structure, not failing the rest of the packet. But even that is usually better than nothing.
5. It helps to make one kind of change in one commit, such as, for example, replacing `memcpy()` with `GET_CPY_BYTES()`, and to explain the reasoning behind this kind of change in the commit message, please see previous commits.

fxlb (changes requested):
Output like:  
`[|802.11_radio]Beacon IBSS, PRIVACY [|802.11]`
or
`[|802.11_radio]Beacon [|802.11]`
with two truncated indications seems show there was/is a problem.

print-802_11.c
```diff
@@ -1220,13 +1216,15 @@ parse_elements(netdissect_options *ndo,
 			}
 			break;
 		case E_RATES:
-			memcpy(&rates, p + offset, 2);
+                        GET_CPY_BYTES(&challenge.text, p + offset,
```
fxlb, line 1219: Copy/paste problem?

PR #876, opened 2020-09-22
Update DHCP printer for messages and TLVs defined in RFC4388

fenner:
RFC4388 defines 4 message types and two options, this change adds the message types and changes the options from their historical value to their RFC-defined values.

This points out that the IANA registry for bootp options went through a great cleanup many years ago but many of the removed types are still present in tcpdump.


fenner:
I know Travis on master is broken, but https://travis-ci.org/github/fenner/tcpdump/builds/729289535?utm_medium=notification&utm_source=email is a successful run with the broken combinations commented out.

fxlb:
Thank you!

PR #879, opened 2020-09-28
mptcp: add support for ADD_ADDR "echo bit"

dcaratti:
print "echo" if the echo bit is present in the ADD_ADDR suboption (see
RFC8684 §3.4.1).

Signed-off-by: Davide Caratti <dcaratti@redhat.com>


infrastation:
Thank you for suggesting this improvement. If you want a simple solution for a simple problem, `if (GET_U_1(add_addr->sub_echo) & 0x01)` would do the job without the need to scroll 240 lines away to unpack a single-use macro. However, it you want to solve the problem in a systematic way, you can articulate the MBZ bits of the 4-bit field by having a simple `struct tok[]` variable and using it with `bittok2str()`.

infrastation:
Apparently, none of the existing MPTCP tests have the E bit set, do you have a sample .pcap file for this change?

dcaratti:
> Thank you for suggesting this improvement. If you want a simple solution for a simple problem, `if (GET_U_1(add_addr->sub_echo) & 0x01)` would do the job without the need to scroll 240 lines away to unpack a single-use macro. However, it you want to solve the problem in a systematic way, you can articulate the MBZ bits of the 4-bit field by having a simple `struct tok[]` variable and using it with `bittok2str()`.

thanks for reviewing! Ok, maybe the first option is the quickest one.

> Apparently, none of the existing MPTCP tests have the E bit set, do you have a sample .pcap file for this change?

the echo bit is described in the recent V1 specification, and its implementation landed in Linux for the first time a couple of weeks ago. Sure, I will add a test that uses a trace from packetdrill or kselftests.

dcaratti:
> > Thank you for suggesting this improvement. If you want a simple solution for a simple problem, `if (GET_U_1(add_addr->sub_echo) & 0x01)` would do the job without the need to scroll 240 lines away to unpack a single-use macro. However, it you want to solve the problem in a systematic way, you can articulate the MBZ bits of the 4-bit field by having a simple `struct tok[]` variable and using it with `bittok2str()`.
> 
> thanks for reviewing! Ok, maybe the first option is the quickest one.

on a second thought, it's better to check all values of these 4 bits, like you suggested: in this way, it becomes possible to:
- check also the correctness of 'ver' in v0 packets, as specified in [version 0 of the protocol] https://tools.ietf.org/html/rfc6824#section-3.4.1)
- inform users when MBZ bits are non-zero on v0 and v1 logs

> > Apparently, none of the existing MPTCP tests have the E bit set, do you have a sample .pcap file for this change?
> 
> the echo bit is described in the recent V1 specification, and its implementation landed in Linux for the first time a couple of weeks ago. Sure, I will add a test that uses a trace from packetdrill or kselftests.

... done

infrastation:
Thank you. It is unusual to validate protocol version numbers with bitmaps, but let's see. The bitmap is 16-bit, but the argument to `bittok2str()` is 8-bit. Also the MBZ bit numbers are one-off. Could you check?

dcaratti:
> Thank you. It is unusual to validate protocol version numbers with bitmaps, but let's see. The bitmap is 16-bit, but the argument to `bittok2str()` is 8-bit. Also the MBZ bit numbers are one-off. Could you check?

... because it's not a bitmap actually, and I should use `tok2str()` rather than `bittok2str()`. When dissecting an `ADD_ADDR` 
conformant to MPTCP v0, the echo bit is systematically 0 and the 'res' bit contain either  0x2 or 0x3. As a consequence, the whole 4 bits contain 0x4 or 0x6 (that is a hint about the length of the IP address in the option). On the contrary, when dissecting an `ADD_ADDR` conformant to MPTCP v1, the echo bit can be 0 or 1, while the remaining bits must be zero.

infrastation:
It would be possible to define one bit as the smallest amount of binary information that can cause a considerable confusion.

infrastation:
All looks good now, merged, thank you!

PR #880, opened 2020-09-30
Fix ssize_t for _WIN64

gvanem:
Compiling Windump for `x64` using clang-cl, gives this warning:
```c
tcpdump.c(1085,45): warning: format specifies type 'ssize_t' (aka 'long long') but the argument has type 'ssize_t'
      (aka 'int') [-Wformat]
                error("short read %s (%zd != %d)", fname, cc, (int)buf.st_size);
                                      ~~~                 ^~
                                      %zd

```
So do as MinGW does and define `ssize_t` as `__int64` for `_WIN64`.

### DIFF
--- a/netdissect-stdinc.h
+++ b/netdissect-stdinc.h
@@ -184,7 +184,11 @@
   /*
    * Windows doesn't have ssize_t; routines such as _read() return int.
    */
-  typedef int ssize_t;
+  #ifdef _WIN64
+    typedef __int64 ssize_t;
+  #else
+    typedef int ssize_t;
+  #endif  
 #endif  /* _MSC_VER */
 
 /*


infrastation:
Some macOS jobs fail in Travis CI because of the environment bootstrap failure, so the CI fault is a false negative. The change looks trivial at a glance, but because so many fine details were involved in spelling integer types right, I would not be a good enough reviewer for this change.

gvanem:
>  so the CI fault is a false negative. 

AFAICS, looking at the Travis details, that has nothing to do with this PR. Some issue with curl/homebrew:
```
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
```

infrastation:
Yes, that is correct. I have raised the failure with Travis support.

guyharris:
That function probably needs some UN\*X-vs-Windows and 32-bit-vs-64-bit cleanup.  I'll look at that.

guyharris:
> That function probably needs some UN*X-vs-Windows and 32-bit-vs-64-bit cleanup. I'll look at that.

faf8fb70af3a013e5d662b8283dec742fd6b1a77 should clear this up.

infrastation:
@gvanem, could you confirm the fix and close this issue if it is resolved?

gvanem:
> could you confirm the fix and close this issue if it is resolved?

Yes, the warning on `x64` is now gone. Closing.

PR #888, opened 2020-10-27
Clean exit from tcpdump with asan.

enh-google:
Fixes the use-after-free of `pd` caused by leaving the signal handlers
that call pcap_breakloop() active after freeing the `struct pcap_t`.

This also fixes the leak of `device`.

Caught by gwp-asan in the wild. Android internal bug 170778539.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -2239,8 +2239,12 @@ main(int argc, char **argv)
 		error("%s", pcap_geterr(pd));
 	if (dflag) {
 		bpf_dump(&fcode, dflag);
-		pcap_close(pd);
+		/* Clear pd so our signal handler won't use-after-free it. */
+		pcap_t *to_free = pd;
+		pd = NULL;
+		pcap_close(to_free);
 		free(cmdbuf);
+		free(device);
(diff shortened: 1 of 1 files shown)


tcpdump.c
```diff
@@ -2239,8 +2239,12 @@ main(int argc, char **argv)
 		error("%s", pcap_geterr(pd));
 	if (dflag) {
 		bpf_dump(&fcode, dflag);
-		pcap_close(pd);
+		/* Clear pd so our signal handler won't use-after-free it. */
+		pcap_t *to_free = pd;
+		pd = NULL;
+		pcap_close(to_free);
```
mcr, line 2245: I feel like maybe we need to block signals somewhere here. We need the transfer to to_free, and the NULL of pd to happen atomically, right?

mcr: It would be nice to somehow move that logic into pcap_close(), but that won't work for other users, I think.
So we do this twice, let's have a function in tcpdump.c near the signal handler, that does that.

mcr (changes requested):

PR #898, opened 2021-01-05
MPLS: rename the "exp" field to "tc"

fenner:
RFC5462 (https://tools.ietf.org/html/rfc5462) renamed the
MPLS "EXP" field to "Traffic Class" in 2009.  Use "tc" as an
abbreviation for this field name.


mcr (approved):

fxlb:
Thank you!

PR #905, opened 2021-02-10
BFD: add SBFD support (RFC7880 and RFC7881)

fenner:
Add support for SBFD on UDP port 7784.  SBFD is different
in that packets from the reflector will be sent with *source*
port 7784; in all other BFD mechanisms, it is only the
destination port that matters.  For SBFD print-udp.c has to
check both source and destination port.


infrastation:
(Rebased on the current master to see up-to-date CI results.)

infrastation:
Bill, would you mind if I update this to print "S-BFD" instead of "Sbfd"? Other than that, the proposed changes seem to match the spec exactly.

infrastation:
Thank you for making this contribution!

fenner:
> Bill, would you mind if I update this to print "S-BFD" instead of "Sbfd"? Other than that, the proposed changes seem to match the spec exactly.

Hi Denis, thank you very much for merging this and for this naming improvement.  I apologize for not having been able to reply quickly, and I appreciate that you made the updates.

PR #907, opened 2021-02-25
IPv6: Implemented Tunnel Payload Forwarding (TPF) option

nickludwig:
Implementation of Tunnel Payload Forwarding option, as specified in [https://tools.ietf.org/html/draft-bonica-6man-vpn-dest-opt-13](https://tools.ietf.org/html/draft-bonica-6man-vpn-dest-opt-13).


print-ip6opts.c
```diff
@@ -196,6 +196,13 @@ ip6_opt_process(netdissect_options *ndo, const u_char *bp, int len,
 		ND_PRINT(")");
 	    }
 	    break;
+	case IP6OPT_TPF:
+	    if (GET_U_1(bp + 1) != 4) {
```
fenner, line 200: You're checking the length of the first option, no matter where this option is in the list.  Other option printers use `bp + i + 1` as the location of the length field.

print-ip6opts.c
```diff
@@ -196,6 +196,13 @@ ip6_opt_process(netdissect_options *ndo, const u_char *bp, int len,
...
+		ND_PRINT("Tunnel Payload Forwarding option: invalid len");
+		goto trunc;
+	    }
+	    ND_PRINT("Tunnel Payload Forwarding (TPF) option");
```
fenner, line 204: Is it useful to print the 32-bit TPF Information?

ip6.h
```diff
@@ -157,6 +157,7 @@ struct ip6_dest {
 #define IP6OPT_TYPE_ICMP	0xC0
 
 #define IP6OPT_MUTABLE		0x20
+#define IP6OPT_TPF		0x41	/* 01 0 00001 */
```
fenner, line 160: This value has not been assigned by IANA yet (and 0x41 does not follow the registry's guideline for "distinct values in the "rest" subfield", so may not match what is eventually assigned).  I do not think it should be merged with an un-allocated value.

fenner (changes requested):

PR #911, opened 2021-03-05
.cirrus.yml: quote strings in only_if expression

edigaryev:
Hi!

We're going to make some changes to how Cirrus CI parses configurations in the next couple of weeks (there's no press-release ATM) and found some potential issues with your `.cirrus.yml` while doing backwards compatibility checks.

Conditional task execution using`only_if` script should quote strings, similarly to the [examples in the documentation](https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution), otherwise the expression engine might treat these characters as operators.

This change makes sure the transition goes smoothly.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,7 +4,7 @@ env:
 
 freebsd_task:
   name: FBSD+BM
-  only_if: $CIRRUS_BRANCH != coverity_scan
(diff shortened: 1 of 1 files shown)


fxlb:
Thank you!

PR #912, opened 2021-03-09
DNS: Sync types with IANA

phy1729:
IANA reference: https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml

I updated the comments to reference the latest RFC where they didn't already. Let me know if you'd prefer that I leave that out.

Removed the `UNSPECA` type parsing as IANA has that query type number assigned to `NID` now.


fxlb:
Rebased, commit message updated, merged, thank you!

PR #914, opened 2021-03-19
MPTCP: Print 'Length' inside mptcp options

mruprich:
Current mptcp parser prints 'mptcp <Subtype>...' but RFC 8684 in Section 3. states that the format should be 'mptcp <Length> <Subtype>...'

I added the printing of the Length field as just a plain number, it seems to me that having 'mptcp Length 8 dss...' is a little bit disruptive.

PR #915, opened 2021-03-23
MPTCP: parse MPC data_len field

pabeni:
The RFC 8684 introduced a new field in the MPC header: the data_len.
If the client has not seen yet a DSS sub-option from the server, it
must use MPC with data_lan instead of a DSS for the first packet
carrying a data payload.

This change print the data_len, if present, after the receiver_key.

The field is already there in the existent pcap test trace, we just
need to update the expected output.

Signed-off-by: Paolo Abeni <pabeni@redhat.com>


mruprich:
@infrastation Hi Denis, Paolo rebased the PR so that it corresponds to data from https://github.com/the-tcpdump-group/tcpdump/pull/914. Is there anything that Paolo or me can do to help?

mruprich:
Any update on this PR?

fxlb:
Merged, thank you!

PR #918, opened 2021-05-12
mptcp: add support for MP_TCPRST

dcaratti:
dissect the MP_TCPRST sub-option in accordance to RFC8684 §3.6.

Link: https://github.com/multipath-tcp/mptcp_net-next/issues/189
Signed-off-by: Davide Caratti <dcaratti@redhat.com>


print-mptcp.c
```diff
@@ -454,6 +480,23 @@ mp_fast_close_print(netdissect_options *ndo,
...
+{
+        const struct mp_tcprst *mpr = (const struct mp_tcprst *)opt;
+
+        if (opt_len != 4)
+                return 0;
+
+        ND_PRINT(" flags [%s]", bittok2str_nosep(mp_tcprst_flags, "none",
+                 GET_U_1(mpr->sub_b)));
+
+        ND_PRINT(" reason %s", tok2str(mp_tcprst_reason, "invalid (0x%02x)",
```
matttbe, line 495: detail: should we indicate `unknown` instead of `invalid`? The RFC specifies 7 possible reasons but it doesn't forbid extensions later.

https://www.rfc-editor.org/rfc/rfc8684.html#name-subflow-reset

print-mptcp.c
```diff
@@ -182,6 +183,31 @@ struct mp_prio {
...
 
+static const struct tok mp_tcprst_flags[] = {
+        { 0x08, "U" },
+        { 0x04, "V" },
+        { 0x02, "W" },
+        { 0x01, "T" },
+        { 0, NULL }
+};
+
+static const struct tok mp_tcprst_reason[] = {
```
matttbe, line 194: (very little) detail if you have to update this commit: just to uniform with what is above, maybe best to add `s` (`reasons`) and use `0x0X` (`0x06` instead of `0x6` like in the RFC :)

matttbe (commented):
Thanks Davide for working on that!
It looks good to me, just 2 very small details, it would be fine without them but up to you ;-)

print-mptcp.c
dcaratti: indeed, it's not `invalid ` just because it's not a known value of `mp_tcprst_flags`. Will fix that.

print-mptcp.c
dcaratti: sure, will do. Thanks for reviewing!

matttbe (approved):
thank you for this patch :-)

Looks good to me for the MPTCP part, hopefully to TCPDump maintainers as well :)

print-mptcp.c
```diff
@@ -182,6 +183,31 @@ struct mp_prio {
...
+};
+
+static const struct tok mp_tcprst_reasons[] = {
+        { 0x06, "Middlebox interference" },
+        { 0x05, "Unacceptable performance" },
+        { 0x04, "Too much outstanding data" },
+        { 0x03, "Administratively prohibited" },
+        { 0x02, "Lack of resources" },
+        { 0x01, "MPTCP-specific error" },
+        { 0x00, "Unspecified error" }
```
fxlb, line 201: Thanks for the PR.
There is a missing '{ 0, NULL }' here.
And please rebase on top of the-tcpdump-group:master.

fxlb (changes requested):

print-mptcp.c
dcaratti: > Thanks for the PR.
> There is a missing '{ 0, NULL }' here.

added, and ran again `make check` - to ensure it still passes when it parses MPTCP_TRST sub-option having `reason` equal to zero. 

> And please rebase on top of the-tcpdump-group:master.

done.
thanks for reviewing!
-- 
davide

mcr (approved):

mcr:
Will merge when Appveyor checks complete.  They are always slow.

fxlb (approved):

PR #923, opened 2021-06-27
Fix the misleading comment regarding "From DS", "To DS" Frame Control Flags

gokul-kumar-s:
In the case where the "From DS" bit is not set and the "To DS" bit is set in the Frame Control Bitmask, the comment line in the corresponding if else block was misleading that the "From DS" bit is set and the "To DS" bit is not set.


fxlb:
Thanks.
s/In the case where the the/In the case where the/

gokul-kumar-s:
> 
> 
> Thanks.
> s/In the case where the the/In the case where the/

I have updated the PR.

fxlb:
Merged, thanks.

PR #925, opened 2021-07-07
WIP (test for now) Reinstate running tcpdump -J/-L and capture

fxlb:


infrastation:
linux-s390x is not good for the packet capture test, I had to cancel [this build](https://ci.tcpdump.org/#/builders/41/builds/66) because it was stuck for 5 hours.

PR #937, opened 2021-09-03
Remove last remaining trailing whitespace

a1346054:
The CONTRIBUTING file advises us to "Avoid trailing tabs/spaces". This PR removes last remaining trailing whitespace from the source code, making it easier to setup a warning in case any new trailing whitespace were to be introduced.


fxlb:
Thank you!

PR #939, opened 2021-09-09
Decode FQDN Capability

ton31337:
Tested using FRR/GoBGP.

Signed-off-by: Donatas Abraitis <donatas.abraitis@gmail.com>


fxlb:
This capability was introduced in an expired individual I-D.
```
Type: Expired Internet-Draft (individual)
Last updated: 2016-07-21 (latest revision 2016-01-06)
```
(https://datatracker.ietf.org/doc/draft-walton-bgp-hostname-capability/)

Any reason to decode it?

ton31337:
The reason is simple, there are daemons supporting it: frr, gobgp, bird, exabgp, ... Even the number is assigned by IANA and wireshark has this as well.

infrastation:
Thank you for preparing these changes. It would help to understand who is the most interested party in maintaining this protocol extension, if there is such a party.

* It is not Cumulus Networks, as they abandoned the I-D in 2016 and ceased to exist as a standalone company in 2020.
* It is not Nvidia, as with their deep pockets and 1.5 years of owning the former Cumulus Networks business Nvidia could have the extension revived without effort, but they didn't.
* It is not any of the tcpdump maintainers, as it occurs again and again that someone contributes an implementation of a specific revision of an I-D and disappears instead of updating tcpdump code to subsequent revisions of the I-D or the final document, leaving tcpdump maintainers with the ballast.
* Is it you?

ton31337:
@infrastation hello,

>It would help to understand who is the most interested party in maintaining this protocol extension, if there is such a party.

Do you mean extension == draft or maintaining this code which is gonna be pushed for tcpdump?

infrastation:
I mean the usual ecosystem of a network protocol. So long as the extension specification and everything that implements it hinge on an IANA allocation in an IETF protocol registry, the specification likely has to be something the registry recognizes (a current independent I-D or something else, it would be best to confirm there). Out of the spec come the protocol instance implementations (which you say already exist) and the packet analyzer implementations (which you are working to get in a better shape now). So are you maintaining the implementations and leaving the spec abandoned?

ton31337:
>So are you maintaining the implementations and leaving the spec abandoned?

Right, I'm not the author of the draft.

ton31337:
Well, it's expired or not, but there are players in the market supporting it and using it intensively. I agree to reject that if capability would be random and without its own IANA number. 

For example, Bird and GoBGP just got this functionality recently. If that's useful, it doesn't care if it's expired or not. And IMO, since the IANA number is there, we should use it and adopt it as much as possible and not show as an Unknown capability.

infrastation:
This discussion seems to require some more time to achieve a good level of mutual understanding.

ton31337:
Do we have any consensus/decision here or just closing? :)

infrastation:
I would have more peace of mind closing this request now rather than talking in circles.

ton31337:
I don't understand then your point and overall attitude to open-source here.

infrastation:
My previous comment was an honest answer to your question ("Do we have any consensus/decision here or just closing? :)") in a comment you have just deleted. I recognize your right not to like the answer and not to answer all questions about this request. Please recognize project maintainers' duty not to accept changes that do not look right. This might make more sense later if you re-read the conversation. Maybe on a different occasion a different request will work out better.

PR #941, opened 2021-09-22
bgp: Decode BGP link-bandwidth extended community properly

ton31337:
First two bytes are reserved for AS.

Before:
```
	  Extended Community (16), length: 8, Flags [OT]:
	    link-BW (0x4004), Flags [non-transitive]: bandwidth: -308754489798474897524123005616128.000 Mbps
	    0x0000:  4004 fde8 47f4 2400
	  Updated routes:
	    192.168.10.0/23
```

After:
```
	  Extended Community (16), length: 8, Flags [OT]:
	    link-BW (0x4004), Flags [non-transitive]: bandwidth: 1.000 Mbps
	    0x0000:  4004 fde8 47f4 2400
	  Updated routes:
	    192.168.10.0/23
```

Signed-off-by: Donatas Abraitis <donatas.abraitis@gmail.com>


infrastation:
Thank you for identifying this problem. Would you like to add a test .pcap file as well?

ton31337:
Sure, I'll do.

ton31337:
@infrastation I added a pcap test for this case.

infrastation:
Thank you for waiting. A very similar reasoning applies to this change as in pull request #939 earlier, with the remarkable difference being that this protocol element has already been in `print-bgp.c` since 2002 and seems to predate the earliest respective internet draft by 7 years. So in this case it makes more sense to split the hair in favour of accepting the change.

That said, please remember that implementations of expired specifications are less likely to survive the code base churn due to the costs of long-term maintenance. As just one example, in this particular case commits that resulted in this little case block are 43fa6e0, c77a3f0, 7c790c6, 1a859e9, ee68aa3 and 12f570c. It has been a while since I last looked into BGP encoding, so it took me to look up draft-ietf-idr-link-bandwidth-07, draft-ietf-idr-link-bandwidth-00, RFC 7153 Section 5.2.3 and RFC4360 Section 3.1 to verify that your simple change is safe and consistent with the last revision of the I-D, which expired on 2018-09-06.

As before, if you need this protocol element to persist long-term, please find someone to revive the specification.

PR #942, opened 2021-09-26
BGP: add dissector for BGPsec capability and path

colinbs:
Add BGPsec capablility and path attribute parsing to the BGP dissector. BGPsec is defined by [RFC8205](https://datatracker.ietf.org/doc/html/rfc8205).
Here is the example output of capability and BGPsec path attribute:

Capability:
```
Option Capabilities Advertisement (2), length: 5
  BGPsec (7), length: 3
    Version 0, Direction Receive (0), AFI IPv4 (1)
```

BGPsec path attribute:
```
BGPsec Path (33), length: 203, Flags [OE]: 
  Secure Path Length: 14
    Secure Path Segment: pCount: 1, Flags: [none] (0x00), AS: 3
    Secure Path Segment: pCount: 1, Flags: [none] (0x00), AS: 2
  Signature Block: Length: 189, Algo ID: 1
    Signature Segment:
      SKI:
        0x0000:  0ca0 6bdf c294 88fa 1a7f 4a47 93f3 a940
        0x0010:  10a8 0f97
      Length: 71
      Signature:
        0x0000:  3045 0221 0087 24ba eae1 09bc a23a c0f1 
        0x0010:  794c e301 3bab 329b 7c23 5e0a a0fd a4ca 
        0x0020:  0afe 736c ef02 203b 42ff c318 7568 16bb 
        0x0030:  3e88 0796 7379 db4f a83c fd56 1821 28d2 
        0x0040:  b78a 6ec7 145e 75 
    Signature Segment:
      SKI:
        0x0000: 0a15 fa1c b92f 4d34 95fe f0ca 9180 c0a9
        0x0010: 3c4a 254f 
      Length: 71
      Signature:
        0x0000:  3045 0220 7d87 e54d f871 4400 768a 3db7 
        0x0010:  acfa 55ae be23 e200 3cdf 21b6 1424 0d1e 
        0x0020:  8831 49e3 0221 00e2 d828 edb0 0277 dc25 
        0x0030:  f4c5 aa4e 2906 74e3 aa13 6c30 f78b d415 
        0x0040:  0e55 35ef 3fc8 70
```


print-bgp.c
```diff
@@ -2496,6 +2500,63 @@ bgp_attr_print(netdissect_options *ndo,
...
+                 GET_U_1(tptr + idx + 2));
+        idx += 3;
+        sblen -= 3;
+
+        /* parse signature segments */
+        while (sblen > 0) {
+            siglen = GET_BE_U_2(tptr + idx + 20);
+            ND_PRINT("\n\t      Signature Segment:\n\t        SKI: ");
+            for (j = 0; j < 20; j++) {
+                ND_PRINT("%02x ", GET_U_1(tptr + idx + j));
```
colinbs, line 2543: This loop seems highly inefficient and ugly, but I couldn't get `hex_print` to work... I'd be happy about some guidance.

print-bgp.c
```diff
@@ -2496,6 +2500,63 @@ bgp_attr_print(netdissect_options *ndo,
...
+            }
+            idx += 22;
+            sblen -= 22;
+
+            ND_PRINT("\n\t        Length: %u\n\t        Signature:", siglen);
+            for (j = 0; j < siglen; j++) {
+                if (j % 16 == 0) {
+                    ND_PRINT("\n\t          ");
+                }
+                ND_PRINT("%02x ", GET_U_1(tptr + idx + j));
```
colinbs, line 2553: Here as well.

print-bgp.c
fxlb: Something like:
```patch
hex_print(ndo, "\n\t          ", tptr + idx, 20);
```
?

print-bgp.c
fxlb: Something like:
```patch
hex_print(ndo, "\n\t          ", tptr + idx, siglen);
```
?

print-bgp.c
colinbs: Thanks, looks like this works like a charm.

infrastation:
The changes to `config.h.in` do not look relevant.

colinbs:
> The changes to `config.h.in` do not look relevant.

TBH I have no idea where they come from. I will remove them.

print-bgp.c
```diff
@@ -2506,6 +2510,56 @@ bgp_attr_print(netdissect_options *ndo,
...
+
+        ND_PRINT("\n\t    Secure Path Length: %u", splen);
+
+        idx += 2;
+        splen -= 2;
+        /* parse secure path segments */
+        while (splen > 0) {
+            ND_PRINT("\n\t      Secure Path Segment: pCount: %u, Flags: %u, AS: %u",
+                     GET_U_1(tptr + idx),
+                     GET_U_1(tptr + idx + 1),
```
guyharris, line 2535: Should this use `bittok2str()` or `bittok2str_nosep()` to show the individual flags (or flag, in this case), rather than just dumping the flag value as a decimal number?

print-bgp.c
```diff
@@ -2555,6 +2609,23 @@ bgp_capabilities_print(netdissect_options *ndo,
...
+            if (cap_len < 3) {
+                ND_PRINT(" (too short, < 3)");
+                return;
+            }
+
+            ND_PRINT("\n\t\tVersion %u, Direction %s (%u), Flags 0x%01x, AFI %s (%u)",
+                      GET_U_1(opt + i + 2)&0xf0,
+                      (GET_U_1(opt + i + 2)&0x08) ? "Send" : "Receive",
+                      (GET_U_1(opt + i + 2)&0x08)>>3,
+                      GET_U_1(opt + i + 2)&0x07,
```
guyharris, line 2624: Should this use `bittok2str()` or `bittok2str_nosep()` to show the individual flags, rather than just dumping the flag value as a decimal number?

guyharris: Or should it just ignore those bits, as [RFC 8205 section 2.1](https://datatracker.ietf.org/doc/html/rfc8205#section-2.1) appears to show them as "Unassigned"?

print-bgp.c
colinbs: Right, `Flags` are currently ignored (and are thus always 0). I oriented myself on the example here:
https://datatracker.ietf.org/doc/html/rfc8208#appendix-A.4
There, the flags are also displayed. Whether or not this is useful, I am not sure myself.

print-bgp.c
colinbs: Right, `Flags` are currently ignored (and are thus always 0). If it is common in tcpdump to omit such fields, I'm all for it to scratch them.

print-bgp.c
colinbs: Just like with capabilities, here, only the leftmost bit of the octet is relevant (see [RFC 8205 Section 3.1](https://datatracker.ietf.org/doc/html/rfc8205#section-3.1)). To make the output consistent with the rest of the code, I would suggest to scrap the `Flags` field.

colinbs: One more question: since there is only one possible flag to be set (or not set), how to best display the value?
`Flags: [0/1]`
or
`Flags: [Set/Unset]`
or leave out the `Flags` field entirely if its value is 0?

infrastation: The specification calls the other 7 bits "unassigned" and explains they are a typical case of "must be zero", so you could spell the individual bits as `MBZ-n` (as done in `print-udld.c`, for example) and use `bittok2str()`.

colinbs: Thank you very much for the pointer. I updated the code and the example output accordingly. Let me know if is as you intended it to be.

print-bgp.c
```diff
@@ -2506,6 +2522,60 @@ bgp_attr_print(netdissect_options *ndo,
...
+            tptr += splen;
+            break;
+        }
+
+        ND_PRINT("\n\t    Secure Path Length: %u", splen);
+
+        idx += 2;
+        splen -= 2;
+        /* parse secure path segments */
+        while (splen > 0) {
```
fxlb, line 2544: Insecure loop:
If Secure Path Length == 9 (invalid file), at the end of the loop `splen` will have the following values:
```
1
65531
65525
65519
...
```
Thus the `(splen > 0)` condition is not sufficient.
Before the loop is needed:
```
        if (splen == 0 || splen % 6) {
            ND_PRINT(" [invalid total segments len %u]", splen);
            break;
        }
```
and the condition of the loop should be `(splen != 0)` to make
clearer what we want to test.

colinbs: Thanks, good catch! I will make changes accordingly.

print-bgp.c
```diff
@@ -2506,6 +2522,60 @@ bgp_attr_print(netdissect_options *ndo,
             len -= 12;
         }
         break;
+    case BGPTYPE_BGPSEC_PATH:
+    {
+        uint16_t sblen, splen, siglen;
+        u_int idx;
```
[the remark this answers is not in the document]
colinbs: Valid point, I will remove idx.

print-bgp.c
```diff
@@ -2506,6 +2522,60 @@ bgp_attr_print(netdissect_options *ndo,
...
+
+        splen = GET_BE_U_2(tptr);
+        idx = 0;
+
+        if (splen < 8) {
+            ND_PRINT(" [secure path too short]");
+            tptr += splen;
```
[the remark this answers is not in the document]
colinbs: Nope. Removed!

print-bgp.c
colinbs: Sadly, due to variable length of signature segments, the same cannot be applied there. But I added a check for minimum length.

infrastation:
Quite a number of Linux CI builds failed.  For example, on linux-mips the reason for failure was as follows:
```
configure:6044: checking for pcap_loop
configure:6044: gcc -o conftest -g -O2   conftest.c ../libpcap/libpcap.a -libverbs -lhfi1verbs-rdmav22 -lipathverbs-rdmav22 -lrxe-rdmav22 -libverbs -lpthread -lnl-route-3 -lnl-3 -lnl-genl-3 -lnl-3  -ldbus-1 -lpthread -lsystemd  -lsmi  >&
5
/usr/bin/ld: cannot find -lsystemd
collect2: error: ld returned 1 exit status
```
Based on the fact Cirrus CI tries to use FreeBSD 13.0, these changes need to be rebased on the current master branch in order to get an up to date status (the build failure might go away, or might remain, or might change to a different error).

fxlb:
Yes a rebase is needed with recent libpcap/tcpdump changes. Done.

print-bgp.c
```diff
@@ -2543,6 +2559,80 @@ bgp_attr_print(netdissect_options *ndo,
...
+                     flags,
+                     asn);
+            tptr += 6;
+            splen -= 6;
+        }
+
+        sblen = GET_BE_U_2(tptr);
+
+        /*
+         * A signature block has a minimum length of 92 bytes:
```
fxlb, line 2604: 95 bytes?

colinbs: Whoops, that's embarrassing. Fixed it.

print-bgp.c
```diff
@@ -2543,6 +2559,80 @@ bgp_attr_print(netdissect_options *ndo,
...
+        }
+
+        sblen = GET_BE_U_2(tptr);
+
+        /*
+         * A signature block has a minimum length of 95 bytes:
+         * 2 bytes length field
+         * 1 bytes algorithm id
+         * 92-94 bytes per signature segment
+         *  '-> 20 bytes SKI, 2 bytes length field, 70-72 bytes signature
```
fxlb, line 2608: To better understand how to test for length errors:
"70-72 bytes signature". Where is this specified?

colinbs: I was trying really hard to actually find a resource that states how long a ECDSA signature is. I solely base this claim on my observations of never having seen a signature of different size other than 70, 71 or 72 bytes. Not the most solid foundation to stand on, I must confess...

fxlb: >  I solely base this claim on my observations of never having seen a signature of different size other than 70, 71 or 72 bytes.

**If** the answer to the following question is correct (and **if** I understand it correctly), it would only be a matter of probability.
https://bitcoin.stackexchange.com/questions/112662/what-is-the-probability-of-an-ecdsa-signature-being-less-than-71-bytes

colinbs: Ah I see, highly interesting! I didn't new this. I guess the check for a minimum length must be set to a lower value. Should it be 64+3, according to the SE answer? Or even lower, in case a signature could potentially be even smaller in size?

fxlb: As stated in the answer "Its output contains:": It's only a partial output with the highest probabilities.

fxlb:
But it's not very important to find a minimum length for the test. I think this test (and the comment) before the loop is not useful, because it don't fix the invalid loop (underflow risk).
Example: If (Signature Block Length)==95 and (Segment Length)==71 (invalid file), After sblen -= siglen, sblen = 65535 -> invalid loop (underflow).

fxlb:
In such case, we want something like:
```
            Signature Block: Length: 95, Algo ID: 1
              Signature Segment:
                SKI: 
                  0x0000:  0a15 fa1c b92f 4d34 95fe f0ca 9180 c0a9
                  0x0010:  3c4a 254f
                Length: 71 [remaining length 70 < 71]
```

fxlb:
We could use something like the following,  using the "new" (from Dec 2021, see ef5323efe6c03aaa50eac0949bf95999b91fe05b) ND_ICHECKMSG_U() macro  :
```patch
diff --git a/print-bgp.c b/print-bgp.c
index 9f91c2c6..b684f286 100644
--- a/print-bgp.c
+++ b/print-bgp.c
@@ -2561,7 +2561,7 @@ bgp_attr_print(netdissect_options *ndo,
         break;
     case BGPTYPE_BGPSEC_PATH:
     {
-        uint16_t sblen, splen, siglen;
+        uint16_t sblen, splen;
 
         splen = GET_BE_U_2(tptr);
 
@@ -2620,13 +2620,21 @@ bgp_attr_print(netdissect_options *ndo,
         sblen -= 3;
         /* Parse signature segments */
         while (sblen > 0) {
-            siglen = GET_BE_U_2(tptr + 20);
+            uint16_t siglen;
+
             ND_PRINT("\n\t      Signature Segment:\n\t        SKI: ");
+            ND_ICHECKMSG_U("remaining length", sblen, <, 20);
             hex_print(ndo, "\n\t          ", tptr, 20);
-            tptr += 22;
-            sblen -= 22;
+            tptr += 20;
+            sblen -= 20;
+            siglen = GET_BE_U_2(tptr);
+            ND_ICHECKMSG_U("remaining length", sblen, <, 2);
+            tptr += 2;
+            sblen -= 2;
 
-            ND_PRINT("\n\t        Length: %u\n\t        Signature:", siglen);
+            ND_PRINT("\n\t        Length: %u", siglen);
+            ND_ICHECKMSG_U("remaining length", sblen, <, siglen);
+            ND_PRINT("\n\t        Signature:");
             hex_print(ndo, "\n\t          ", tptr, siglen);
             tptr += siglen;
             sblen -= siglen;
@@ -2647,6 +2655,9 @@ done:
     }
     return 1;
 
+invalid:
+    return 1;
+
 trunc:
     return 0;
 }
```

fxlb:
[pr_942_1a.patch.txt](https://github.com/the-tcpdump-group/tcpdump/files/9693110/pr_942_1a.patch.txt)

colinbs:
I see, this macro proves to be very useful. The length check prior to the loop has been removed as it really does not help all that much.
Thank you for the effort of refactoring the code!

print-bgp.c
```diff
@@ -2592,6 +2682,22 @@ bgp_capabilities_print(netdissect_options *ndo,
                tok2str(bgp_safi_values, "Unknown", GET_U_1(opt + i + 5)),
                GET_U_1(opt + i + 5));
             break;
+        case BGP_CAPCODE_BGPSEC:
+            /* Version (4 bits), Direction (1 bit), Flags (3 bits), AFI (16 bits) */
+            cap_offset = 1;
+            if (cap_len < 3) {
```
fxlb, line 2688: Suggested change:
```patch
diff --git a/print-bgp.c b/print-bgp.c
index 70082ef2..765ed7a7 100644
--- a/print-bgp.c
+++ b/print-bgp.c
@@ -2685,8 +2685,9 @@ bgp_capabilities_print(netdissect_options *ndo,
         case BGP_CAPCODE_BGPSEC:
             /* Version (4 bits), Direction (1 bit), Flags (3 bits), AFI (16 bits) */
             cap_offset = 1;
-            if (cap_len < 3) {
-                ND_PRINT(" (too short, < 3)");
+            /* The capability length [...] MUST be set to 3. */
+            if (cap_len != 3) {
+                ND_PRINT(" [%u != 3]", cap_len);
                 return;
             }
 
```

fxlb: [pr_942_2a.patch.txt](https://github.com/the-tcpdump-group/tcpdump/files/9693370/pr_942_2a.patch.txt)

print-bgp.c
```diff
@@ -2543,6 +2559,76 @@ bgp_attr_print(netdissect_options *ndo,
...
+        uint16_t sblen, splen;
+
+        splen = GET_BE_U_2(tptr);
+
+        /*
+         * A secure path has a minimum length of 8 bytes:
+         * 2 bytes length field
+         * 6 bytes per secure path segment
+         */
+        if (splen < 8) {
```
fxlb, line 2573: To harmonize with other changes these four lines could be replaced with:
`ND_ICHECKMSG_U("secure path length", splen, <, 8);`

fxlb:
Thanks for the update. I suggest to rebase these 8 commits into one clean commit before merging.
To do that: `git rebase -i @~8`
```
pick 72bd0663 BGP: add dissector for BGPsec capability and path
pick 16903cca tests: add pcap for BGPsec test
pick 3402d7a4 bgpsec: make adjustments to the flags field of the bgpsec path
pick 90918f29 tests: update flag field in bgp-bgpsec.out
pick 68cea1dc bgpsec: code improvements
pick 33a16fc2 bgpsec: fix counting error for signature block length
pick 7133eab5 bgpsec: utilize ND_ICHECKMSG_U to check signature block length
pick 628e12ff bgpsec: utilize ND_ICHECKMSG_U for secure path segments
```
Proposal: Use `r` (reword) for the first commit to remove the `Signed-off-by` line. Choose `s` (squash) to keep the commit message for the second commit (it has a `Signed-off-by` line). Choose `f` (fixup) for the other commits (or `s` if you want to keep all the steps, but not sure it is useful).
After: `git push -f`
If you have problems with that, I can do the rebase.

colinbs:
I rebased. Hope it is as you wanted.
Thank you for all the work and feedback!

fxlb:
Good. Thank you for your work and your patience.

PR #889, opened 2020-10-30
DSA: Correctly determine VID

wkz:
The 4 MSBs of the VID is stored in the lower nibble of the tag's third
byte.

Previously the priority bits where folded into the VID space, e.g. a
packet with VID=1 and priority 6 was printed as having a VID of
1537 (0x601).

PR #945, opened 2021-10-04
Fix printing VLAN ID values larger than 256 for Marvell DSA switches

vladimiroltean:
The VLAN ID in Marvell DSA tags is split into two bytes of the switch
header. The extraction logic for the upper byte is broken, it is copy
pasted from the DSA_PRI above. So any VLAN ID higher than 256 gets
truncated to only the bottom-most 8 bits.

### DIFF
--- a/print-dsa.c
+++ b/print-dsa.c
@@ -83,7 +83,7 @@
 #define DSA_RX_SNIFF(tag) TOK(tag, 1, 0x04, 2)
 #define DSA_CFI(tag) TOK(tag, 1, 0x01, 0)
 #define DSA_PRI(tag) TOK(tag, 2, 0xe0, 5)
-#define DSA_VID(tag) ((u_short)((TOK(tag, 2, 0xe0, 5) << 8) | (TOK(tag, 3, 0xff, 0))))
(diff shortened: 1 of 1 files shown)


mcr:
Could you add (or amend) a test case that shows that this is fixed (and will remain fixed in the future).
Thank you.

fxlb:
It is a duplicate of #889.

vladimiroltean:
So considering that Tobias' patch does the exact same thing, should I close this PR and you can merge his?

fxlb:
Yes, but waiting first for an update and rebase of his PR.

fxlb:
@vladimiroltean
Do you have the DSA documentation or a link to it?

vladimiroltean:
http://file2.dzsc.com/product/19/03/27/360334_113151208.pdf
Section 7, starting with page 144.

infrastation:
If that document is the only source of information, it leads me to think that there may be a greater peace of mind in losing the (E)DSA decoder altogether.

vladimiroltean:
What other source of information do you expect to see? It is a vendor-specific switch tagging protocol.

fxlb:
> Section 7, starting with page 144.

Thanks for the link.

PR #946, opened 2021-10-05
some parameters for enabling probot on the repository

mcr:
https://github.com/organizations/the-tcpdump-group/settings/installations/19853667 explains how to enable probot.
This bot can close issues which have not had any activity.
Part of the goal of using this is to make active issues more active, and to figure out which issues have not had any focus.
The text of the message and the daysUntilX are open for discussion.
A similiar setting for libpcap would be enacted.
Maybe it's all a bad idea.  Opinions?

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  activity in the past 60 days.
+  Code contributions should include test pcap files, and may need to be rebased regularly.
+  Discussion of non-trivial contributes on the tcpdump-workers@lists.tcpdump.org mailing list
+  is encouraged.
+  This issue will be closed if no further activity occurs.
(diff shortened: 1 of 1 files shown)


infrastation:
"N days without updates" alone would not be a good criteria to chase/close a stale work item. Many of these are stale for a reason (a complicated problem or an expensive solution), and some of them still contain bits of relevant knowledge waiting to become a part of the documentation or a bugfix/workaround. Issues imported from SourceForge are no longer connected to their original reporters, so only the developers could respond to such automatic chasing anyway.

Which leads me to think it would be more productive to have a weekly job to pick a random ancient issue and to e-mail a random developer asking to review it and to see if it is ready to be closed. It might take a year to filter through this aspect of the technical debt as such a background activity, but this is a 30 year old project after all.

On the note of "please rebase", for me it would be a relief if a parrot script could respond with the message when the reporter updates the pull request and the branch is more than X commits behind. Instead of Francois-Xavier or myself, that is. Just chasing the authors to rebase no matter what and not having the time to review afterwards would be rude.

infrastation (commented):
As you likely noticed, in the last several weeks I picked a few dozen easy issues and tried to chase them, with varying results. Quite a few have been closed on different grounds, some others resulted in better problem statements. 

Expectedly, one default approach does not work well for every case, and it is going to stand even more for more complicated issues, which remain in quantity. So for the time being human-driven backlog chasing makes notably more sense to me than script-driven.

That's not to say that later the structure of the backlog can become uniform enough to handle certain aspects with a script, but there is a significant technical debt to address before getting there.

infrastation:
Closing because this suggestion is not actionable in foreseeable future. The change has been archived in tcpdump-incomplete-works/mcr-enable-probot-stale.

PR #949, opened 2021-10-25
rt6: parse TLV

giulio-sido:
Parse Type Length Values (TLV) in IPv6 Routing Header as specified by
RFC 8754 [1].

[1] https://datatracker.ietf.org/doc/html/rfc8754

Signed-off-by: Giulio Sidoretti <giulio.sidoretti@uniroma2.it>


ip6.h
```diff
@@ -197,6 +197,15 @@ struct ip6_srh {
...
 
+/**
+ * Type 0 => Pad1
+ * Type 4 => PadN
+ * Type 5 => HMAC
+ */
+#define IPV6_RTHDR_TLV_TYPE_0 0
+#define IPV6_RTHDR_TLV_TYPE_4 4
+#define IPV6_RTHDR_TLV_TYPE_5 5
+
```
ahsalam, line 208: These TLVs are SRH specific so I would suggest to rename them. 
Also you could use the name of the TLV in the definition instead of just using TYPE_0, TYPE_4,... 

I would do: 


#define IPV6_SRH_TLV_PAD1    0 
#define IPV6_SRH_TLV_PADN   4
#define IPV6_SRH_TLV_HMAC   5
#define IPV6_RTHDR_TLV_TYPE_4 4
 #define IPV6_RTHDR_TLV_TYPE_5 5

print-rt6.c
```diff
@@ -33,13 +33,118 @@
 
 #include "ip6.h"
 
+int
+rt6_tlv_print(netdissect_options *ndo, const u_char *p, u_int bytes_left)
```
ahsalam, line 37: The function is related to SRH and will be called for SRH only (case IPV6_RTHDR_TYPE_4).

so i think it should be renamed to srh_tlv_print ()

ahsalam (commented):
Thanks for submitting the PR. I believe it very relevant and also on time. Many other SRH TLVs are being defined as well in IETF. This PR provides the infrastructure needed to easily support them in the future.

netdissect.h
```diff
@@ -700,6 +700,7 @@ extern void rpki_rtr_print(netdissect_options *, const u_char *, u_int);
 extern void rrcp_print(netdissect_options *, const u_char *, u_int, const struct lladdr_info *, const struct lladdr_info *);
 extern void rsvp_print(netdissect_options *, const u_char *, u_int);
 extern int rt6_print(netdissect_options *, const u_char *, const u_char *);
+extern int rt6_tlv_print(netdissect_options *, const u_char *, u_int);
```
fxlb, line 703: No need to declare `rt6_tlv_print` in  netdissect.h. It is only used in `print-rt6.c` where it should be `static`.

giulio-sido:
Thanks for your comments, I have added these minor changes

ahsalam:
@fxlb any update on this PR ? 

@giulio-sido There is a couple of SRH TLVs defined in https://datatracker.ietf.org/doc/html/draft-ietf-spring-sr-service-programming-05#section-7.2.1. They are still waiting for Type assignment by IANA. When they get assigned a type, will you be willing to send one more PR to support them ? 

Thanks!

giulio-sido:
> @giulio-sido There is a couple of SRH TLVs defined in https://datatracker.ietf.org/doc/html/draft-ietf-spring-sr-service-programming-05#section-7.2.1. They are still waiting for Type assignment by IANA. When they get assigned a type, will you be willing to send one more PR to support them ?

Sure. In the meantime, all TLVs with types that are not in the RFC 8754 are parsed as generic TLVs.

ahsalam:
> > @giulio-sido There is a couple of SRH TLVs defined in https://datatracker.ietf.org/doc/html/draft-ietf-spring-sr-service-programming-05#section-7.2.1. They are still waiting for Type assignment by IANA. When they get assigned a type, will you be willing to send one more PR to support them ?
> 
> Sure. In the meantime, all TLVs with types that are not in the RFC 8754 are parsed as generic TLVs.

Great, thanks Giulio!

ahsalam:
Hi @fxlb, could you please have a look on that PR? we have other new features that we plan to submit but they depend on that PR. thanks!

giulio-sido:
Hi, I have just rebased yesterday because the main repository had advanced and there was a little conflict to solve. Now the checks have all passed, so the PR should be ready.

fxlb:
First, some questions:
You process three IPV6_SRH_TLV_* and default.
1) Why 2 IPV6_SRH_TLV_* are cases of the switch and IPV6_SRH_TLV_PAD1 is outside the switch with a 'continue'?
2) Have you some pcap files with pad1 and padn TLVs?

fxlb:
(I have rebased your branch on top `the-tcpdump-group:master`, please get the updated branch.)

giulio-sido:
Hi @fxlb, sorry for the late reply. I'll answer your questions.

1) I have chosen to parse the IPV6_SRH_TLV_PAD1 outside the switch statement because in that case there is not a further byte to read containing the TLV length, which is found in all the other TLVs. I thought it would have been cleaner this way, but I'm open to changes in the code if you think I should implement the logic differently.

2) I have added a new pcap file containing an example with the padding TLVs with relative test.

The CI tests are still running as of now, but hopefully everything will work. Let me know if you have any more comments.

Thanks for your time!

giulio-sido:
Hi @fxlb, I have edited a file because it didn't pass one check. There were a couple trailing spaces.

print-rt6.c
```diff
@@ -33,13 +33,118 @@
...
+		{
+		case IPV6_SRH_TLV_PADN:
+			ND_PRINT(", TLV-type=PadN(%u)", tlv_type);
+			ND_PRINT(", TLV-len=%u", tlv_len);
+			p += tlv_len;
+			bytes_left -= tlv_len;
+			if (bytes_left == 0)
+			{
+				parse_next = 0;
+				break;
```
fxlb, line 73: What is the goal of this break?

print-rt6.c
```diff
@@ -33,13 +33,118 @@
...
+				hmac_byte = GET_U_1(p);
+				p += 1;
+				bytes_left -= 1;
+				if (ndo->ndo_vflag)
+					ND_PRINT("%02x", hmac_byte);
+			}
+			if (bytes_left == 0)
+			{
+				parse_next = 0;
+				break;
```
fxlb, line 109: Same question.

print-rt6.c
```diff
@@ -33,13 +33,118 @@
...
+static int
+srh_tlv_print(netdissect_options *ndo, const u_char *p, u_int bytes_left)
+{
+	u_int tlv_type, tlv_len;
+	uint8_t parse_next = 1;
+	while (parse_next)
+	{
+		tlv_type = GET_U_1(p);
+		p += 1;
+		bytes_left -= 1;
```
fxlb, line 45: Use ND_ICHECK() before decrement.

The recommended way is:
Test if we can't decrement length thus invalid, decrement length, increment pointer. like:
ND_ICHECKMSG_U("remaining length", bytes_left, <, 1);
bytes_left -= 1;
p += 1;

With, at the end of srh_tlv_print():
 
invalid:
   return -1;

Reminder for ND_ICHECK*() macros: If the checked expression is true an error message is printed and a 'goto invalid' is executed.

This should avoid incorrect writing: "(invalid TLV length ..., bytes left ...) (invalid)"

print-rt6.c
```diff
@@ -33,13 +33,118 @@
 
 #include "ip6.h"
 
+static int
+srh_tlv_print(netdissect_options *ndo, const u_char *p, u_int bytes_left)
+{
+	u_int tlv_type, tlv_len;
+	uint8_t parse_next = 1;
```
fxlb, line 40: Is parse_next useful? (parse_next == 0 when bytes_left == 0)

fxlb:
> I have chosen to parse the IPV6_SRH_TLV_PAD1 outside the switch statement because [...]

OK.

> I have added a new pcap file containing an example with the padding TLVs with relative test.

Thanks.

print-rt6.c
```diff
@@ -33,13 +33,118 @@
...
+		{
+			ND_PRINT(" (invalid TLV length %u, bytes left %u)", tlv_len, bytes_left);
+			return -1;
+		}
+
+		switch (tlv_type)
+		{
+		case IPV6_SRH_TLV_PADN:
+			ND_PRINT(", TLV-type=PadN(%u)", tlv_type);
+			ND_PRINT(", TLV-len=%u", tlv_len);
```
fxlb, line 67: From RFC:
```
Length:
    0 to 5. The length of the Padding field in bytes.
```
A test could be added.

print-rt6.c
giulio-sido: I wasn't aware of that macro, thanks for pointing it out, I am now using it in the code

print-rt6.c
giulio-sido: I have added the test for this

giulio-sido:
Thaks @fxlb for your feedback. I have updated the code and rebased on the last commit of the master. Let's wait for the CI tests to finish.

PR #962, opened 2021-11-21
bgp: Parse rfc9072

ton31337:


print-bgp.c
```diff
@@ -2673,47 +2683,56 @@ bgp_open_print(netdissect_options *ndo,
...
-               ND_PRINT("\n\t      no decoder for option %u",
-               opt_type);
-               break;
+      case BGP_OPT_AUTH:
+      default:
+        ND_PRINT("\n\t      no decoder for option %u", opt_type);
+        break;
         }
-        i += BGP_OPT_SIZE + opt_len;
+        i += opt_size + opt_len;
```
fenner, line 2735: This would be easier to review if it didn't reindent the existing code.

print-bgp.c
```diff
@@ -2673,47 +2683,56 @@ bgp_open_print(netdissect_options *ndo,
         GET_BE_U_2(bgp_open_header->bgpo_holdtime));
     ND_PRINT("ID %s", GET_IPADDR_STRING(bgp_open_header->bgpo_id));
     optslen = GET_U_1(bgp_open_header->bgpo_optlen);
-    ND_PRINT("\n\t  Optional parameters, length: %u", optslen);
+    if (optslen == BGP_OPEN_NON_EXT_OPT_LEN) {
```
fenner, line 2686: rfc9072 says in 2 places that the length doesn't have to be 255 - SHOULD when constructing and MUST ignore on receipt in section 2:
```
   The Non-Extended Optional Parameters Length field (Non-Ext OP Len.)
   SHOULD be set to 255 on transmission ...
   ...; it MUST be ignored on receipt once the use of the extended
   format is determined positively by inspection of the Non-Extended
   Optional Parameters Type (Non-Ext OP Type) field.
```

And in section 3,
```
   It is not considered an error to receive an OPEN message whose
   Extended Optional Parameters Length value is less than or equal to
   255.  It is not considered a fatal error to receive an OPEN message
   whose (non-extended) Optional Parameters Length value is not 255 and
   whose first Optional Parameter type code is 255 -- in this case, the
   encoding of this specification MUST be used for decoding the message.
```

I think the fix for this is to simply eliminate this if, and just check for `opsttype == BGP_OPEN_NON_EXT_OPT_TYPE_EXTENDED_LENGTH`.

fenner (changes requested):

print-bgp.c
ton31337: optslen can't be ignored, because you don't know in advance how to parse the message. Encoding is determined by looking in optstype, but it should be checked only if optslen is 255, otherwise you will check one byte ahead in a wrong place.

print-bgp.c
ton31337: I'll reduce this noise.

ton31337: @fenner fixed.

print-bgp.c
fenner: Can you say more about how the RFC is incorrect about how to treat the length field?  What problem do you anticipate if you ignore the Non-Extended Optional Parameters Length field, in the case where the Non-Extended Optional Parameters Type identifies an extended parameter?

ton31337: Yes, we are just about the parsing here, fixed.

print-bgp.c
```diff
@@ -80,13 +80,15 @@ struct bgp_open {
     nd_uint16_t bgpo_holdtime;
     nd_uint32_t bgpo_id;
     nd_uint8_t  bgpo_optlen;
+    nd_uint8_t bgpo_opttype;
+    nd_uint16_t bgpo_optlen_extended;
```
fenner, line 84: Two things here:
1. Please line up opttype by adding a space before it, like other nd_uint8_t's
2. Please add a comment something like /* used for extended optional parameters (RFC9072) */

print-bgp.c
```diff
@@ -80,13 +80,15 @@ struct bgp_open {
...
     /* options should follow */
 };
 #define BGP_OPEN_SIZE        29    /* unaligned */
 
 struct bgp_opt {
     nd_uint8_t bgpopt_type;
-    nd_uint8_t bgpopt_len;
+    nd_uint16_t bgpopt_len;
```
fenner, line 91: Please add a comment saying that this can be one or two bytes, depending on whether or not RFC9072 is in use.

print-bgp.c
```diff
@@ -2673,22 +2682,32 @@ bgp_open_print(netdissect_options *ndo,
...
-        if (length < BGP_OPT_SIZE + i)
+        ND_TCHECK_LEN(opt + i, opt_size);
+        if (length < opt_size + i)
             goto trunc;
         bgpopt = (const struct bgp_opt *)(opt + i);
         opt_type = GET_U_1(bgpopt->bgpopt_type);
-        opt_len = GET_U_1(bgpopt->bgpopt_len);
-        if (BGP_OPT_SIZE + i + opt_len > optslen) {
+        opt_len = extended_opt_params ? GET_BE_U_2(bgpopt->bgpopt_len)
+                                    : GET_U_1(bgpopt->bgpopt_len);
```
fenner, line 2709: Very minor: can you line up the `:` with the `?`?

fenner (changes requested):
I looked at the pcap, and it looks correct, and the test output looks good.  I have only very small requests about comments and spacing - I know they seem like small issues but I honestly believe that things like this make it easier to maintain the code long term.

print-bgp.c
ton31337: Fixed, thanks.

print-bgp.c
ton31337: Fixed, thanks.

print-bgp.c
ton31337: Fixed, thanks.

fenner (approved):
Thanks for the updates!

ton31337:
@fxlb @infrastation just rebased to the latest master (was a conflict with TESTLIST file). Can be merged?

ton31337:
@fxlb @infrastation just rebased to the latest master (was a conflict with TESTLIST file). Can be merged?

fxlb:
The code and the test should be in one commit.

ton31337:
Gonna fix in few minutes.

ton31337:
> The code and the test should be in one commit.

Fixed.

fxlb:
Thank you!

fxlb:
Please, always put code and test in the same commit.

ton31337:
Thanks, noted.

Issue #116, opened 2013-04-16
Timestamp isn't getting printed for some packets


guyharris:
Converted from [SourceForge issue 2802813](http://sourceforge.net/support/tracker.php?aid=2802813), submitted by alessandro83

Hello.
I'm using Tcpdump on some remote machine to gain some information about the traffic. I'm using a script that automatical analyze the tcpdump's output and parse it using the timestamp.
Sometime(very rare) I've noticed that tcpdump doesn't write the timestamp, and when this happen my script fail invalidating my test. Someone can explain me while this happen?Is it a bug?It happens if two timestamp are the same?There is a way to avoid it?
thank you

PR #968, opened 2021-12-14
Add option to parse and print every Nth packet

nats:
New option `--sampling-rate=NTH` will parse and print every NTH packet, with
all other packets producing no output. The default behaviour is
`--sampling-rate=1`, which prints every packet.

The sampling rate feature is useful for real-time inspection of an interface
with a high packet rate, or initial inspection of large capture files.


fxlb:
Thanks for your proposal. The print_usage() function need an update.

fxlb:
About the option name:
The `rate` for `--sampling-rate 5` (by example) is `0.2`. Thus we could name the option just `--sampling`, `--sampling 5` understood sampling every 5 packets.

infrastation:
A couple points about the user interface of the proposed changes:
* Since sampling is not a new concept (sFlow had it at least 15 years ago it, at least in some implementations), it might be relevant to apply it at capture-time as well. Even if that isn't implemented right away (as perhaps the best place to do that would be as close to the OS kernel as possible), it would be easier to add it later if the option name stated which sampling it controls, for example `--print-sampling` for the proposed feature with a room for `--capture-sampling` in future.
* Also, since print-time sampling implies `-S`, it would be better to enable it automatically instead of expecting the user to remember to do it every time.

guyharris:
> A couple points about the user interface of the proposed changes:
> 
> * Since sampling is not a new concept (sFlow had it at least 15 years ago it, at least in some implementations), it might be relevant to apply it at capture-time as well. Even if that isn't implemented right away (as perhaps the best place to do that would be as close to the OS kernel as possible),

...in which case there should be a libpcap API to enable sampling, so that the implementation is hidden from applications doing sampling (as the implementation might be platform-dependent, and applications shouldn't have to know the platform details).

WinPcap and Npcap have a slightly weird API to specify sampling - [`pcap_setsampling()`](https://www.winpcap.org/docs/docs_412/html/group__wpcapfunc.html#ga84c7d55f5b0e7d545b38d6b8e7bde005) returns a pointer to [a `struct pcap_samp`](https://www.winpcap.org/docs/docs_412/html/structpcap__samp.html) for a `pcap_t`, and the caller fills in that structure with the desired parameters.  "Sampling parameters **cannot** be changed when a capture is active. These parameters must be applied **before** starting the capture. If they are applied when the capture is in progress, the new settings are ignored." - but, with `pcap_open_live()`, you can't do that, and I think they added that API before we had `pcap_create()` and `pcap_activate()`.

The structure has a member that specifies the sampling method; they offer:

* `PCAP_SAMP_NOSAMP`, which is "every packet, no sampling";
* `PCAP_SAMP_1_EVERY_N`, which is "every Nth packet";
* `PCAP_SAMP_FIRST_AFTER_N_MS`, which is "1 packet every N milliseconds".

The other member specifies N.

As they note, this affects *both* live captures *and* reading a capture file.

We could add this to libpcap-on-UN\*X, and add it to the other live-capture pcap-XXX.c files (with a helper routine so there's less code duplication).

nats:
I have renamed to `--print-sampling`, added it to `print_usage()`, enabled `-S`, and revised testcase and man page to match.

I have also enabled `--print`,  since `tcpdump -w test.cap --print-sampling 100` not producing any output feels counter-intuitive given the flag+option names are now so similar, and `tcpdump -w test.cap --print --print-sampling 100` feels redundant.

guyharris:
Why not just `--sample-packets`?  Presumably, it should work regardless of whether you're doing a live capture or reading from a file *and* should work regardless of whether you're printing dissected packets or writing them to a file.

nats:
>  or writing them to a file

This PR only affects printing.  Because writing to file is not affected, you can do: `timeout 5m tcpdump -w test.cap --print-samples 100` , to get an overview of what is being captured without flooding your terminal and consuming a lot of processor time. All packets are written to the file and you can process it however you like afterwards.

In that sense, `--print-sampling` is almost a superset of `--print`; if it were possible to give `--print` an optional argument without breaking back-compat, I would suggest going that route.  However from what I can see it cannot be optional,  as the meaning of `tcpdump --print <filter>` would change.

I think your suggestion of capture-side sampling within libpcap would be great, but that appears to be a more involved change.

guyharris:
OK, so the idea is that this *isn't* intended for writing sampled packets to a file.

That also means that capture-side/read-side sampling won't work, as that eliminates packets before the capturing program sees them, so it can't write those packets to a file.

infrastation:
Are you saying it would be better to have one sampling control knob (whatever the name, but without the "print" in it), which would make tcpdump skip the packets at **read** time, so the effect would be in both the printing branch and the writing branch? And if libpcap later gets an API call to sample packets, then tcpdump would offload the request to libpcap, which possibly one day would offload that to the OS kernel?

guyharris:
> Are you saying it would be better to have one sampling control knob (whatever the name, but without the "print" in it), which would make tcpdump skip the packets at read time, so the effect would be in both the printing branch and the writing branch?

That might be useful for some, but it appears *not* to be what @nats wants here, so it doesn't appear as if it would be completely better.

infrastation:
As another note, `PCAP_SAMP_1_EVERY_N` allows fine-grained control only for large enough values of N (1/2 = 50%, 1/3 = 33%, 1/4 = 25%, 1/5 = 20%...). At the same time, "N percent" does not allow to go below 0.01, if there is such a need (which might matter at 100Gb/s). Perhaps something like "N out of M" (or a float) could satisfy both extremes.

infrastation:
Then it looks like the read/capture time sampling would be a completely different thing, possibly not related to the print time sampling in any way except the potential command line options confusion when (and if) both things come into existence.

nats:
To be clear: the reason I want sampling (wherever it occurs within the packet pipeline) is to reduce processing time.  My personal usecase is I sample IPv4 packets in real-time on an interface , and in my testing I found that printing (parsing) is easily the single biggest consumer of CPU time.

I use a command line on linux like `tcpdump -n --snapshot-length=34 --print-sampling=100 ip` with stdout piped into another program - lets call it `stats` - for processing. Previously I did the packet sampling within `stats` by discarding most of tcpdump's output, and `tcpdump` would use about 85% of 1 core. The addition of `--print-sampling=100` has dropped `tcpdump` to just 3% of 1 core.  

Sampling at the print level was a trivial change and a big improvement for my usage of tcpdump. Pushing the sampling up into libpcap is a lot more work will have minimal further improvement to processing time, so I didn't pursue that.

--

If the thinking is `--print-sampling` is going to be confusing if/when a capture-side sampling option appears in the future, I can certainly understand that perspective and am happy to close this PR.

infrastation:
> If the thinking is `--print-sampling` is going to be confusing if/when a capture-side sampling option appears in the future, I can certainly understand that perspective and am happy to close this PR.

No, that should not be a problem ("let's capture 2% of what has arrived to the interface and print 5% of what was captured" could be a valid use case). This is an interesting new feature, let's think about it instead of closing.

nats:
The other approach I'm happy to put some work into is to capture-side sampling just within tcpdump's callback functions (`print_packet`, `dump_packet`, `dump_packet_with_trunc`) either in addition to or instead of print-side sampling.

It wouldn't make any real improvement to processing time, but would at least open the door to a future libpcap-based implementation without needing to change command-line options.

guyharris:
At this point if you have a need for print-side-only sampling, and only for that, we should just merge this branch and wait for somebody to express a need for capture-side sampling and do that in libpcap, with changes to tcpdump to use the libpcap APIs for it.

nats:
I dont specifically need print-side sampling (though as noted I do think there is merit in this PR's behaviour for `tcpdump -w test.cap --print-sampling 100`), it was  the simplest approach because there's no single `callback` function and I didnt want to duplicate the logic across the three functions. That said, if you are happy for a small refactor moving `pcap_handler callback` into a static global with a single callback for `pcap_loop` to call its pretty equivalent; something like...

```
static pcap_handler packet_callback;

static void
process_packet(u_char *user, const struct pcap_pkthdr *h, const u_char *sp)
{
    /* implement sampling here */
    ++packets_captured;
    ++infodelay;

    /* invoke print_packet/dump_packet/dump_packet_with_trunc as per existing behaviour */
   pcap_callback(user, h,k sp);
}
``` ?

**Comment by guyharris** [comment]
> I dont specifically need print-side sampling (though as noted I do think there is merit in this PR's behaviour for `tcpdump -w test.cap --print-sampling 100`)

So let's just merge this and, if anybody says they want capture-side sampling, add that in libpcap and add a separate command-line flag to tcpdump to use it.

**Comment by infrastation** [comment]
@nats, please rebase on the current master branch, update the man page to say that `--print-sampling` enables `--print`, squash the commits into one and make the resulting commit message up to date. If you need help doing that, shout.

**Comment by infrastation** [comment]
Also the timestamp in the first `.TH` of the man page needs to be updated to indicate a substantial change.

**Comment by nats** [comment]
OK, I will have a look

**Comment by infrastation** [comment]
Thank you!

**Comment by infrastation** [comment]
For reference, OpenBSD libpcap implements sampling by use of the `sample` keyword instead of extending the C API.
```
openbsd-amd64# tcpdump -ni lo0 -d 'sample 1000'
(000) ld       #random
(001) jgt      #0x418937        jt 2	jf 3
(002) ret      #0
(003) ret      #116
```
A disadvantage of this approach is the dependency on the random number BPF extension. An advantage is that sampling can be applied differently to different parts of the expression if required.

guyharris:
> A disadvantage of this approach is the dependency on the random number BPF extension.

Perhaps it's time for an API to inquire, for a given `pcap_t`, which BPF extensions are supported. (`pcap_t`, so that it can be per-device *and* can be supported for savefiles - we can add extensions to libpcap's BPF implementation so that you could use this for extracting a sample of packets from a capture file using tcpdump).

That way, we could provide better error messages, *before* trying to set a filter, for attempts to use a feature such as that on a platform that doesn't support it, rather than trying to infer the cause of an "invalid filter" error from some OS's kernel-level BPF implementation.

infrastation:
Yes, it is a bit more complicated when there is more than one OS to support.

PR #969, opened 2021-12-17
Add --sampling option for capture-side sampling

nats:
- New option `--sampling NTH` for sampling captured packets.
- Received packets are counted via new global `packets_received`.
- Captured packets are discarded in new function `capture_packet` as required,
  and passed to existing `dump_packet`/`dump_packet_with_trunc`/`print_packet`.
- The `-S` flag is enabled by default, since relative TCP sequence numbers do
  not make a lot of sense when sampling.

CLI long option has been given provisional name of `--sampling` pending
initial review. `print_usage()`, and man page will be updated after desired
option name is confirmed.

The filtering is done entirely within tcpdump, but should allow for switching to a libpcap implementation under the same CLI option in the future.

If this is an approach you are happy to go with for now, let me know what long and/or short option name you want  and I will finish up `print_usage()` and the man page.


nats:
The automatic `-S` flag under capture is somewhat nuanced; since if you `tcpdump -w test.cap --sampling 100` and later `tcpdump -r test.cap`  then relative TCP sequence numbers will be using when printing it.

As such it may be preferable to not enable it automatically, and instead just suggest using `-S` in the man page when printing  sampled captures.

infrastation:
This is a duplicate of #968, closing.

PR #974, opened 2022-01-24
bgp: Adopt to rfc9003 (Extended Shutdown Message)

ton31337:


infrastation:
I have followed the spec and the code and this change looks correct. Thank you.

PR #978, opened 2022-03-02
Add suggested fix for CVE-2018-19519

adventureloop:
There is a buffer read in print-hncp.c error due to a lack of initialisation of the buffer.
Fix my initialising the buffer with a null byte as suggested in the write up of the 
discoverer of the problem:

"Initialize buf[0] to '\0' in 'print_prefix'.
buf[0] = '\0';
(Just as print-cnfp.c and print-bgp.c do.)"

https://github.com/zyingp/temp/blob/master/tcpdump.md

### DIFF
--- a/print-hncp.c
+++ b/print-hncp.c
@@ -206,6 +206,7 @@ print_prefix(netdissect_options *ndo, const u_char *prefix, u_int max_length)
 {
     int plenbytes;
     char buf[sizeof("xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx::/128")];
+    buf[0] = '\0';
 
     if (GET_U_1(prefix) >= 96 && max_length >= IPV4_MAPPED_HEADING_LEN + 1 &&
         is_ipv4_mapped_address(prefix + 1)) {


fxlb:
See issue #763.

guyharris:
["The problem"](https://github.com/zyingp/temp/blob/master/tcpdump.md) no longer exists in the main branch or the 4.99 branch, so it's already fixed.

The test pcap file never makes it past libpcap's tests, so it can't be used to test unless the packet headers are fixed.  (As I've said on a number of occasions before: if you want to test tcpdump, rather than libpcap, *DO NOT* fuzz the packet headers in a way that modifies the captured length, or that modifies the on-the-network length so that it becomes < the captured length.)

guyharris:
The bug is present in 4.9.x releases prior to 4.9.3; it's fixed in 4.9.3, and is fixed in all releases after 4.9.3, so it's fixed in the current release and earlier 4.99.x releases.

It's also present in 4.8.1, but there's no 4.8.x release in which it's fixed; anybody using 4.8.x will need to back port the fix or upgrade to 4.9.3 or later.

adventureloop:
Thanks for the response, I found this fix in a downstream vendors tree as a part of the process of upstreaming changes. I'm sorry this seems to be a continuing fight against people trying to fuzz tcpdump.

guyharris:
The only problem with people fuzzing tcpdump is when they fuzz the pcap file record headers in such a way that causes the capture to get an error at the libpcap layer, meaning that tcpdump doesn't try to dissect the packet and no bugs are revealed by the fuzzing - all the fuzzing does, in the packet with the fuzzed record header, is indicate that libpcap is doing its job correctly.

It's probably best if fuzzers limit their fuzzing to the packet contents; at least for pcap files, the vast majority of bugs are likely to be in tcpdump, not libpcap, so whatever libpcap bugs aren't caught by fuzzing only the packet contents are probably more than made up for by tcpdump bugs caught by the fuzzing because the record headers *weren't* fuzzed and thus *didn't* cause tcpdump to give up on the capture file.

PR #980, opened 2022-03-03
OpenSSL_add_all_algorithms() is deprecated since OpenSSL 1.1.0, use it accordingly

### LINKED ISSUE #977: OPENSSL_API_COMPAT is not defined by OpenSSL 3.0
With OpenSSL 3.0, it's really up to the application to define `OPENSSL_API_COMPAT`,  and when compiling tcpdump,  `OPENSSL_API_COMPAT` is not defined, thus the check in `esp_init()` is incorrect:

```
#if !defined(OPENSSL_API_COMPAT) || OPENSSL_API_COMPAT < 0x10100000L
        OpenSSL_add_all_algorithms();
#endif
```

https://github.com/the-tcpdump-group/tcpdump/blob/9ff9f78db3cd17d3761fdc441ad58f1ab4917507/print-esp.c#L683

When compiled against OpenSSL 3.0.1 built with `--api=1.1.1` and  `no-deprecated` configuration options, `OpenSSL_add_all_algorithms()` is not defined, and build fail:

> 
> ```
> ./print-esp.c: In function 'esp_init':
> ./print-esp.c:682:2: warning: implicit declaration of function 'OpenSSL_add_all_algorithms' [-Wimplicit-function-declaration]
>   OpenSSL_add_all_algorithms();
>   ^
> libnetdissect.a(print-esp.o): In function `esp_init':
> .../print-esp.c:682: undefined reference to `OpenSSL_add_all_algorithms'
> collect2: error: ld returned 1 exit status
> ```

ydroneaud:
OPENSSL_API_COMPAT is to be set by the module including OpenSSL header to request a certain API level compatibility, and starting with OpenSSL 3, it's no more defined by OpenSSL header to the default API level.

OPENSSL_VERSION_NUMBER must be used instead.

Fixes #977

### DIFF
--- a/print-esp.c
+++ b/print-esp.c
@@ -674,12 +674,7 @@ static void esp_print_decode_onesecret(netdissect_options *ndo, char *line,
 DIAG_OFF_DEPRECATION
 static void esp_init(netdissect_options *ndo _U_)
 {
-	/*
-	 * 0.9.6 doesn't appear to define OPENSSL_API_COMPAT, so
-	 * we check whether it's undefined or it's less than the
-	 * value for 1.1.0.
-	 */
-#if !defined(OPENSSL_API_COMPAT) || OPENSSL_API_COMPAT < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 	OpenSSL_add_all_algorithms();
 #endif
 	EVP_add_cipher_alias(SN_des_ede3_cbc, "3des");


DimitriPapadopoulos:
As a side note, `EVP_add_cipher_alias` should not be used, see https://github.com/the-tcpdump-group/tcpdump/commit/d44658b2e47ad1a3724a632f0d65a81140654c15.

PR #986, opened 2022-03-31
Babel v4viav6

jech:
This adds support for displaying routes announced according to https://datatracker.ietf.org/doc/html/draft-ietf-babel-v4viav6, which is currently in the RFC Editor's queue.


jech:
@mcr, @fenner Could you please take a look?  This is being deployed in production.

mcr (approved):
Seems fine to me, thank for the additional test case.  
Do you have coverage of all the code paths?

jech:
No, the provided trace only contains Update TLVs with the new encoding.  The new encoding is also allowed in Request TLVs, but they are NOT RECOMMENDED by the RFC, and are not generted by any implementation in existence: I'd need to specifically craft a packet with them.

jech:
Ping?

mcr (approved):

PR #988, opened 2022-04-04
Add support for BFD and Hard Reset notifications, and GR N flag

fenner:
Add support for the Hard Reset notifications and the corresponding
N flag in Graceful Restart capability advertisement, per RFC8538.

Add the BFD cease notification subcode assigned by IANA.
This value is an early assignment for an in-progress standard.


fxlb:
Thank you for preparing these changes.
I suggest to have 2 PR:
1) for the RFC:
Add support for the Hard Reset notifications and the corresponding
N flag in Graceful Restart capability advertisement, per RFC8538.

2) for the draft:
Add the BFD cease notification subcode assigned by IANA.
This value is an early assignment for an in-progress standard.

fenner:
As requested, opened #1002 and #1003 for the two different parts.

PR #989, opened 2022-04-05
Add pcap files to exercise corner cases for BGP shutdown message printer

fenner:
3 new tests to make sure to exercise the BGP shutdown message printer:
1. A packet with more data than the length provides for
2. A packet that has a shutdown message with zero data
3. A packet whose shutdown message length is longer than the remaining data in the notification

I noticed these cases when porting my Hard Reset code to master, and wanted to make sure to have test cases so that the Hard Reset changes don't break these corner cases.


fxlb:
Thanks.


>  1. A packet with more data than the length provides for
>  2. A packet that has a shutdown message with zero data
>  3. A packet whose shutdown message length is longer than the remaining data in the notification

I suggest adding these informations in the commit message with a `git commit --amend` and `git push -f`.

fxlb:
Merged, thanks!

PR #990, opened 2022-04-13
TLS: Initial support for printing TLS PDUs

ryandoyle:
Supports printing plain text parts of TLS. Basic record layer printing is
supported as well as more complex printing of handshake and alert protocols,
specifically ClientHello and ServerHello - these will print interesting parts
of TLS extensions like ALPN, SNI and the underlying TLS version negotiated.

As we don't keep state and TLS is stateful, this printer does assume/
is limited by this, specifically:
 * TLS version is printed as per record layer version. In the case of TLS 1.3,
   the record layer is set to TLS 1.2, and the supported_versions extension is
   used to negotiate the actual version used.
 * Alert and Handshake protocol messages can be plain text or encrypted,
   depending on if a crypto context is established. This printer doesn't know
   this, so we try to detect it and print "(likely encrypted)".


infrastation:
Thank you for preparing these changes. Apparently, `uint` does not belong to the Windows C compiler vernacular. The type needs to be either `u_int` or `unsigned`.

ryandoyle:
Cheers @infrastation, fixed :+1:

its0x08 (approved):

PR #1001, opened 2022-08-15
NTP: Print extension fields

mlichvar:
Print types, known names and length of extension fields in NTPv4 packets
following RFC 7822. With -vvv dump also their bodies in hex.

Names of Autokey extension fields are not included as the only known implementation is using different types that what is in the IANA table and it conflicts with NTS.


mlichvar:
Rebased on current head.

fxlb:
Thanks for preparing this PR.
First remark: `tests/ntp-time.pcap` should not be updated (regression tests). Add a new pcap for this PR.

mlichvar:
The new tests are now in a separate pcap.

print-ntp.c
```diff
@@ -516,3 +530,63 @@ p_poll(netdissect_options *ndo,
 		ND_PRINT(" (1/%us)", 1U << -poll_interval);
 }
 
+/* Prints an NTPv4 extension field */
+static void
+p_ntp_ef(netdissect_options *ndo, u_int type, u_int length, const u_char *ef_body)
+{
+	switch (type) {
```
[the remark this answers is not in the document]
fxlb: Why only 5 types?
(There are more:
https://www.iana.org/assignments/ntp-parameters/ntp-parameters.xhtml#ntp-parameters-3)

print-ntp.c
```diff
@@ -516,3 +530,63 @@ p_poll(netdissect_options *ndo,
...
+p_ext_fields(netdissect_options *ndo, const u_char *cp, u_int length)
+{
+	const struct ntp_extension_field *ef;
+	u_int ef_type, ef_len, efs_len;
+	int first_ef;
+
+	first_ef = 1;
+	efs_len = 0;
+
+	while (length - efs_len > 24) {
```
fxlb, line 574: Why `24`? A comment would be useful to explain.

print-ntp.c
```diff
@@ -516,3 +530,63 @@ p_poll(netdissect_options *ndo,
...
+		ef = (const struct ntp_extension_field *)(cp + efs_len);
+		ef_type = GET_BE_U_2(ef->type);
+		ef_len = GET_BE_U_2(ef->length);
+
+		if (efs_len + ef_len > length || ef_len < 4 || ef_len % 4 != 0)
```
fxlb, line 579: Need `nd_print_invalid(ndo);` before break?

tests/ntp-time-ef--vv.out
```diff
@@ -0,0 +1,28 @@
+    1  13:23:30.254335 IP (tos 0x0, ttl 64, id 62645, offset 0, flags [DF], proto UDP (17), length 360)
```
fxlb, line 1: diff `tests/ntp-time-ef--v.out` `tests/ntp-time-ef--vv.out`: only "[udp sum ok]" difference. Any need to keep the `vv` test?

fxlb (changes requested):
Thanks for the update.
Some comments.

print-ntp.c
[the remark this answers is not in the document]
mlichvar: > This switch could be replaced by a `struct tok ntp_ef_types`and a `tok2str()` call.

Ok, I did that.

> Why only 5 types? (There are more: https://www.iana.org/assignments/ntp-parameters/ntp-parameters.xhtml#ntp-parameters-3)

The other values in the table are not used by anything. The only known autokey implementation is using different values. This will be fixed by https://datatracker.ietf.org/doc/draft-ietf-ntp-update-registries/, but it's not finished yet. The tcpdump code can be updated later.

print-ntp.c
mlichvar: I added a comment explaining the constant and changed the condition to `>= 28` to make it more obvious.

print-ntp.c
mlichvar: Good catch. Thanks

tests/ntp-time-ef--vv.out
mlichvar: No need to keep the test. It was just for consistency with other ntp tests, but it seems they are already removed.

mlichvar:
Thanks for the review. I think I addressed all issues and rebased it on top of current head.

fxlb (approved):

fxlb:
Thanks!

PR #1002, opened 2022-08-20
Add support for Hard Reset notification and GR N flag

fenner:
Add support for the Hard Reset notifications and the corresponding
N flag in Graceful Restart capability advertisement, per RFC8538.

(This is the RFC8538 part of pull request #988 )


fxlb:
Update done. Merged.

PR #1003, opened 2022-08-20
Add support for early allocated code point for the BGP Cease notification BFD-down Subcode

fenner:
Add the BFD cease notification subcode assigned by IANA.
This value is an early assignment for draft-ietf-idr-bfd-subcode .
Juniper and Arista have or will shortly shipped implementations
using this assigned code point; see
https://trac.ietf.org/trac/idr/wiki/draft-ietf-idr-bfd-subcode%20implementations

This pull request includes the hard-reset code that it requires from
the now-separate pull request #1002 (which should be merged first).
The actual code changes for BFD are just those in 1f091dc

(This is the draft-ietf-idr-bfd-subcode part of pull request #988 )


fenner:
By the way, if there are any concerns that code point 10 will be reassigned to another cease reason: per the procedures in https://github.com/the-tcpdump-group/tcpdump/pull/1003 if a code point allocated using the early allocation procedure expires, it gets marked as deprecated and is not assigned to anything else, unless the working group goes through the procedure to re-mark it as free.  So given that there are shipping implementations and the procedures make it unlikely to be reallocated to something else, I believe that it's safe to ship this code before the allocation is made permanent.

infrastation:
Thank you. Force-pushed to make the BFD change the first one, this way it can be merged cleanly before the more difficult one. I am going to merge this pull request after the CI runs down (it does not seem to like pull requests changing faster than it processes them).

fenner:
On Mon, Aug 22, 2022 at 5:33 PM Denis Ovsienko ***@***.***>
wrote:

> Thank you. Force-pushed to make the BFD change the first one, this way it
> can be merged cleanly before the more difficult one. I am going to merge
> this pull request after the CI runs down (it does not seem to like pull
> requests changing faster than it processes them).
>

Thank you, Denis!

  Bill

PR #1011, opened 2022-10-05
PTP: fix printing of the correction fields

3chas3:
The nanosecond part of the correction field is 48 bits and the sub-nanosecond part is 16 bits.


3chas3:
OK, you win, switching to the format string macros.

print-ptp.c
```diff
@@ -367,14 +370,15 @@ ptp_print_2(netdissect_options *ndo, const u_char *bp, u_int length)
     /* flags */
     len -= 2; bp += 2; flags = GET_BE_U_2(bp); ND_PRINT(", Flags [%s]", bittok2str(ptp_flag_values, "none", flags));
 
-    /* correction NS */
-    len -= 2; bp += 2; ns_corr = GET_BE_U_4(bp); ND_PRINT(", NS correction : %u", ns_corr);
+    /* correction NS (48 bits) */
+    len -= 2; bp += 2; ns_corr = GET_BE_U_4(bp);
```
fenner, line 374: Please use GET_BE_U_6 instead of GET_BE_U_4 and GET_BE_U_2 and shifting.  E.g.,

```
     /* correction NS */
-    len -= 2; bp += 2; ns_corr = GET_BE_U_4(bp); ND_PRINT(", NS correction : %u", ns_corr);
+    len -= 2; bp += 2; ns_corr = GET_BE_U_6(bp); ND_PRINT(", NS correction : %"PRIu64, ns_corr);
 
     /* correction sub NS */
-    len -= 4; bp += 4; sns_corr = GET_BE_U_4(bp); ND_PRINT(", sub NS correction : %u", sns_corr);
+    len -= 6; bp += 6; sns_corr = GET_BE_U_2(bp); ND_PRINT(", sub NS correction : %u", sns_corr);
```

fenner (changes requested):
Can you include a pcap file that includes some values in these fields?  Please see step 6 of https://github.com/the-tcpdump-group/tcpdump/blob/master/CONTRIBUTING.md regarding adding tests.

3chas3:
OK. I don't have any traces with sub-nanosecond correction fields set, just the nanosecond correction field.

fenner (approved):
Thank you for adding the test!

Regarding merge-readiness:
- This change correctly reflects the spec; the previous implementation was just incorrect
- It uses macros (GET_...) and types (uint64_t) and PRIu64 correctly
- The test would have failed with the previous implementation and passes with this implementation

I think this is ready.

fxlb:
@3chas3 Thank you for the fix.
@fenner Thank you for the review.

Issue #686, opened 2018-06-20
Color support: WIP with code+screenshots


ryandoyle:
Hi there , I sent this through to the tcpdump-workers mailing list a bit over a week ago but I never received it on the mailing list myself so it may not have gone through. Are Github issues the preferred method of collaboration or are the mailing lists still the way to go?

Anyhow, this is the email I sent. I'm keen to open up a discussion!

---
Hi there

I have started to implement color/colour support in tcpdump and wanted to get some feedback before I continue. It was mentioned in http://seclists.org/tcpdump/2012/q2/25 a few years back but can't find mentions since. Is this still something that is wanted in tcpdump? 

COLOR IMPLEMENTATION
=================
I have used ANSI escape codes for coloring and have used the 4bit pallet, having "bright/bold" as well.

APPLYING COLORS
============
I have created two helper macros like the ND_PRINT(...) macro. These are ND_PRINT_CATEGORY(category, ...) which will print the ANSI escape code, printf and then reset the terminal back to its normal color.

Also, there is a ND_PAINT_CATEGORY_START(category) and ND_PAINT_CATEGORY_END macros that set the ANSI escape code so existing code can be "painted" in a color without modifying all the ND_PRINT usages to ND_PRINT_CATEGORY. I'm still unsure if this is a good idea.

COLORS
======
I've tried to not tie the idea of color to the printing and instead defined different categories. At the moment I have:

- CATEGORY_TIMESTAMP (green bold): This helps to visually break up between packets. For me, the bright green is quite prominent so I can easily see any line breaks that might happen when printing.

- CATEGORY_PROTO_NAME, (green): Protocol names. EG: IP, IP6, ARP, STP

- CATEGORY_ENDPOINT, (cyan): Any info that is printed that denotes the source and destination of the packet. It could be IP/port pairs, or plain IP addresses or MAC addresses in the case of Ethernet.

- CATEGORY_PROTO_ENCAP_PRIMARY, (magenta bold): Used for important data in any protocols that encapsulate other protocols. EG: In the case of ICMP, it would be IP. In the case of HTTP, it would be IP _and_ TCP. 

- CATEGORY_PROTO_ENCAP_SECONDARY, (magenta): Same as above but used for things like label names.

[...]

PR #1016, opened 2022-11-01
Add colorful output support

alyanser:
Add `--color` arg that makes tcpdump print stuff using colors. The end goal is to make the output more readable.

As of now, only the ability to change colors is added, the colors themselves aren't changed.

<b>Before:</b>
```
ND_PRINT("foo %d, 123");
```
<b>After:</b>
```
ND_PRINT(C_RED, "foo %d, 123"); // prints in red
```

`C_RESET` macro in each `ND_PRINT(...)` ought to be changed to the desired colors' macro defined in `colors.h`. There are roughly 6k invocations of `ND_PRINT` so it wasn't in my scope alone to change the colors. If the idea sounds good, then I'd love to contribute further to provide colorful coverage over all outputs.

I do realize that tcpdump uses other ways than `ND_PRINT` to output stuff but gotta start from somewhere atleast :|

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,5 @@ tcpdump.sln
 .failed
 instrument_functions_global.devel
 instrument_functions_off.devel
+.cache
+compile_commands.json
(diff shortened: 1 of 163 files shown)


infrastation:
Thank you for preparing these changes. See also #686.

PR #1018, opened 2022-11-08
Always warning on bad ip checksum

junka:
IP checksum zero is a value that could be delivered. For modern nic, tso or tx checksum offloading is usually available. However, if some flag got written wrong, the checksum would be set to zero and delivered out. Warning should be necessary.

Signed-off-by: junka <wan.junjie@foxmail.com>


tcpdump.c
```diff
@@ -126,6 +126,7 @@ The Regents of the University of California.  All rights reserved.\n";
 #include <grp.h>
 #endif /* _WIN32 */
 
+#include "depktcap.h"
```
guyharris, line 129: "depktcap.h" is not a system header file on all platforms on which tcpdump runs, so presumably it's a new header file for the tcpdump project that wasn't checked into this pull request.  Please add it to the pull request.

fxlb:
Please avoid merge like "Merge branch 'the-tcpdump-group:master' into master", do a rebase instead.

tcpdump.c
junka: > "depktcap.h" is not a system header file on all platforms on which tcpdump runs, so presumably it's a new header file for the tcpdump project that wasn't checked into this pull request. Please add it to the pull request.

Apologize. I already forgot this MR and use the cloned repo as a submodule for other project. I'll close the MR now.

fxlb:
A PR should not be prepared in the `master `branch.

PR #1019, opened 2022-11-17
bgp: Add support for BGP Role capabilty and OTC attribute

ton31337:
Signed-off-by: Donatas Abraitis <donatas@opensourcerouting.org>


print-bgp.c
```diff
@@ -2654,6 +2673,16 @@ bgp_attr_print(netdissect_options *ndo,
         }
         break;
     }
+    case BGPTYPE_OTC:
+    {
+        if (len < 4) {
+            ND_PRINT("invalid len");
+            break;
+        }
+        ND_TCHECK_LEN(pptr, len);
```
fenner, line 2682: You don't need the TCHECK since you're using GET_ to get the U4 below.

print-bgp.c
```diff
@@ -2766,6 +2795,15 @@ bgp_capabilities_print(netdissect_options *ndo,
                 cap_offset += 4;
             }
             break;
+        case BGP_CAPCODE_ROLE:
+            if (cap_len < 1) {
+                ND_PRINT(" (too short, < 1)");
+                return;
+            }
+            ND_PRINT("\n\t\t Role name %s",
```
fenner, line 2803: Comparing to other capabilities, there's no space between the `\n\t\t` and the description.  Note the additional indentation of the "Role name" vs. "AFI" in the test output:
```
	    Option Capabilities Advertisement (2), length: 3
	      Role Capability (9), length: 1
		 Role name Customer
	    Option Capabilities Advertisement (2), length: 6
	      Multiple Paths (69), length: 4
		AFI IPv4 (1), SAFI Unicast (1), Send/Receive: Receive
```

print-bgp.c
```diff
@@ -2766,6 +2795,15 @@ bgp_capabilities_print(netdissect_options *ndo,
...
+                     tok2str(bgp_role_values, "Unassigned",
```
fenner, line 2804: Do you want to use something like "Unassigned (%d)" so that we know what the numeric value was?

fenner (changes requested):
This looks like a good implementation of the new code points in RFC9234.  I have only a few small changes to suggest but otherwise think these changes are ready to merge.

print-bgp.c
ton31337: Thanks, fixed.

print-bgp.c
ton31337: Thanks, fixed.

print-bgp.c
ton31337: Thanks, fixed.

fxlb:
@ton31337 Thank you for the change.
@fenner Thank you for the review.

PR #1021, opened 2022-12-09
PTP: use the proper values for the control field

fenner:
The PTP printer incorrectly re-used the message type tokens for the control field.  Some of the values are the same, but some are different, requiring a separate token list for correct output.

The extensive test output changes may cause one to wonder, "why make a change like `delay req msg` -> `Delay_Req`?  The answer is that the latter is what's in the table in the spec; it's incorrect to say "msg" for the control field, so there was going to be a change here no matter what.  It's reasonable to have a discussion about what exact strings to put in the tok list, but it's also reasonable to just accept the values that are in the spec :-)


fxlb:
The documentation is behind a paywall.
Is there really an "Other" message?
```
print-ptp.c:184:#define M_OTHER                 0x5
print-ptp.c:197:    { M_OTHER, "Other"},
```

fenner:
> > Is there really an "Other" message?
> 
> As of IEEE 1588-2008, no; message types 4-7 are reserved. There is, however, an "All others" control field value.

These are not message types; they're values for the control field (`IEEE1588-2008: 13.3.2.10 controlField (UInteger8)`)

This is the table in IEEE1588-2008 section 13.3.2.10
Screen Shot 2022-12-11 at 8 23 51 PM

print-ptp.c
```diff
@@ -203,6 +204,24 @@ static const struct tok ptp_msg_type[] = {
...
+#define C_SYNC              0x0
+#define C_DELAY_REQ         0x1
+#define C_FOLLOW_UP         0x2
+#define C_DELAY_RESP        0x3
+#define C_MANAGEMENT        0x4
+#define C_OTHER             0x5
+
+static const struct tok ptp_control_field[] = {
+    { C_SYNC ,"Sync"},
+    { C_DELAY_REQ ,"Delay_Req"},
```
fxlb, line 217: Style: Replace the `' ,'` with `', '`.

guyharris:
> Is it also the case in IEEE 1588-2008?

Yes:

> **13.3.2.10 controlField (UInteger8)**
> 
> The value of controlField depends on the message type defined in the messageType field (see 13.3.2.2) and shall have the value specified in Table 23. The use of this field by the receiver is deprecated.
> 
> NOTE—This field is provided for compatibility with hardware designed to conform to version 1 of this standard.

fxlb:
Thus, we could add:
```patch
diff --git a/print-ptp.c b/print-ptp.c
index 65be8cdc..ba56799c 100644
--- a/print-ptp.c
+++ b/print-ptp.c
@@ -205,6 +205,11 @@ static const struct tok ptp_msg_type[] = {
 };
 
 /* Values from IEEE1588-2008: 13.3.2.10 controlField (UInteger8) */
+/*
+ * The use of this field by the receiver is deprecated.
+ * NOTE-This field is provided for compatibility with hardware designed
+ * to conform to version 1 of this standard.
+ */
 #define C_SYNC              0x0
 #define C_DELAY_REQ         0x1
 #define C_FOLLOW_UP         0x2

```

fxlb:
And remove these two lines:
```
print-ptp.c:184:#define M_OTHER                 0x5
print-ptp.c:197:    { M_OTHER, "Other"},
```

print-ptp.c
fenner: Oops, sorry for missing this - thank you!

fenner: Looks like this was done intentionally, to match the style higher in the file defining `ptp_msg_type[]`.  Would you like me to change these too?

fxlb:
> The spec does specify that 0x5 has a meaning of "Other", and 0x6 and up are reserved. This would mean that a packet with a control field of 0x5 would render as "Reserved"; is that preferable?

I mean remove `M_OTHER`  and keep `C_OTHER`.

fxlb:
Ok.
Now we have `tok2str(ptp_control_field, "Reserved", control)` and `tok2str(ptp_msg_type, "none", msg_type)`.
Should the "none" be replaced by "Reserved" (or "Unknown")?

fenner:
> Ok. Now we have `tok2str(ptp_control_field, "Reserved", control)` and `tok2str(ptp_msg_type, "none", msg_type)`. Should the "none" be replaced by "Reserved" (or "Unknown")?

Table 19 lists message types 4-7 and E-F as "Reserved", so I will change the msg_type call to say `"Reserved"`.

fxlb:
> Table 19 lists message types 4-7 and E-F as "Reserved", so I will change the msg_type call to say `"Reserved"`.

Ok.

fxlb:
Thank you!

fxlb:
New commit for the missing part.

PR #1025, opened 2022-12-22
Group AFNUM needs to be extracted before moving bp pointer

mruprich:
When parsing PGM NCF packet type, the group AFNUM is extracted too late to be correct. In the switch in print-pgm.c under PGM_NCF option, when all the pointers are set, the pointers point to following:
```
struct pgm_header {                         <------ pgm points to the main PGM header
    nd_uint16_t pgm_sport;
    nd_uint16_t pgm_dport;
    nd_uint8_t  pgm_type;
    nd_uint8_t  pgm_options;
    nd_uint16_t pgm_sum;
    nd_byte     pgm_gsid[6];
    nd_uint16_t pgm_length;
};

struct pgm_nak {                     <----- nak points to (pgm + 1) that means it skips the main header and points here
    nd_uint32_t pgmn_seq;
    nd_uint16_t pgmn_source_afi;
    nd_uint16_t pgmn_reserved;
    /* ... uint8_t      pgmn_source[0]; */ <---- bp points to (nak + 1) so right after the first reserved part of the NAK header
    /* ... uint16_t     pgmn_group_afi */
    /* ... uint16_t     pgmn_reserved2; */
    /* ... uint8_t      pgmn_group[0]; */
    /* ... options */   
}
```
And then when the header is parsed, the pgmn_source is parsed, bp moves to pgmn_group_afi which is the value for the next switch but bp is again moved two bytes ahead before the group AFI can be extracted ( bp += (2 * sizeof(uint16_t)); ). 'switch (GET_BE_U_2(bp)) {' will be taking the value from pgmn_reserved2.

### DIFF
--- a/print-pgm.c
+++ b/print-pgm.c
@@ -329,6 +329,7 @@ pgm_print(netdissect_options *ndo,
 	case PGM_NCF: {
 	    const struct pgm_nak *nak;
 	    char source_buf[INET6_ADDRSTRLEN], group_buf[INET6_ADDRSTRLEN];
+	    unsigned afnum;
 
 	    nak = (const struct pgm_nak *)(pgm + 1);
 	    ND_TCHECK_SIZE(nak);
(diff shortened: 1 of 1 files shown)


mruprich:
Hi, did anyone have time to take a look at this? Thanks.

mruprich:
Hi, any update with this PR?

mruprich:
Hi again, is there anything wrong with this PR? Is something missing? Can I do something to get someone to take a look? Do you need some more info from me?

fxlb:
Thanks for the proposal. Before a review you should
1) Use "git rebase" to rebase your work on top on the master branch.
2) Add a pcap file with a PGM NCF packet to show the effect of the change.
(See: CONTRIBUTING.md, item 6)

PR #955, opened 2021-11-04
Parse lower 4 bits in version field as hw info for arista ethertype

BrynXu:
This change is based on PR668, quote its comment first to understand this change easier:

"Arista applied for, and received the assignment of, a vendor-specific EtherType Protocol Identifier in May of 2016. Details below:
Ethertype number is: D28B
Issue date is: May 12, 2016

Arista Subtype 0x01 is Timestamp L2 Header
The timestamp L2 header consist of the following fields:
Arista EtherType (0xD28B)
Two-byte protocol subtype of 0x1
Two-byte protocol version: 0x10 for 64-bit timestamp and 0x20 for 48-bit timestamp"

Recently, Arista divides the two-byte protocol version field further into two fields:
One-byte protocol version: 0x1 for 64-bit timestamp and 0x2 for 48-bit timestamp
One-byte protocol hw version: 0x0 for not-specified, 0x1 for J2

With this new definition, the current dissector will fail to print timestamp correctly, this PR fixes the problem.

The output of tcpdump are like those:

13:36:40.091512 ethertype Arista Vendor Specific Protocol (0xd28b), length 110: SubType: 0x1, Version: 0x0010, Timestamp TAI(64-bit): 2019-05-29 20:36:39, 944724424 ns, IP 10.136.1.32 > 10.2.23.24:  ip-proto-63 62
13:36:40.579036 ethertype Arista Vendor Specific Protocol (0xd28b), length 110: SubType: 0x1, Version: 0x0011, Timestamp TAI(64-bit): 2019-05-29 20:36:40, 432245804 ns, IP 10.136.1.32 > 10.2.23.24:  ip-proto-63 58

13:37:41.551225 ethertype Arista Vendor Specific Protocol (0xd28b), length 108: SubType: 0x1, Version: 0x0020, Timestamp TAI(48-bit): Seconds 60821, Nanoseconds 404038772, IP 10.136.1.32 > 10.2.23.24:  ip-proto-63 62
13:37:42.040995 ethertype Arista Vendor Specific Protocol (0xd28b), length 108: SubType: 0x1, Version: 0x0021, Timestamp TAI(48-bit): Seconds 60821, Nanoseconds 893796872, IP 10.136.1.32 > 10.2.23.24:  ip-proto-63 58

PR #1027, opened 2023-01-04
arista: update test pcap with hwinfo values

fenner:
In the original https://github.com/the-tcpdump-group/tcpdump/pull/955 pull request, the sample pcap was updated with varying types in the hwinfo field.  This was missed in the integration, so this pull request includes the updated pcap and the patches to the tests reflecting the newly varying data.


fxlb:
It's unclear for me what the problem is.
Note that with current tshark, packet number 2 gives, without this change:
```
Frame 2: 110 bytes on wire (880 bits), 110 bytes captured (880 bits)
    [...]
    [Protocols in frame: eth:ethertype:avsp:ethertype:vlan:ethertype:ip:data]
Ethernet II, Src: CIMSYS_33:44:55 (00:11:22:33:44:55), Dst: 00:00:00_00:00:01 (00:00:00:00:00:01)
    [...]
Arista Vendor Specific Protocol
    Sub Type: timestamp (1)
    Version: Version 1 (0x0010)
    Timestamp (TAI)
        Seconds: 1559162200
        Nanoseconds: 432245804
    Type: 802.1Q Virtual LAN (0x8100)
802.1Q Virtual LAN, PRI: 0, DEI: 0, ID: 100
    000. .... .... .... = Priority: Best Effort (default) (0)
    ...0 .... .... .... = DEI: Ineligible
    .... 0000 0110 0100 = ID: 100
    Type: IPv4 (0x0800)
Internet Protocol Version 4, Src: 10.136.1.32, Dst: 10.2.23.24
    0100 .... = Version: 4
    .... 0101 = Header Length: 20 bytes (5)
    Differentiated Services Field: 0x00 (DSCP: CS0, ECN: Not-ECT)
        0000 00.. = Differentiated Services Codepoint: Default (0)
        .... ..00 = Explicit Congestion Notification: Not ECN-Capable Transport (0)
    Total Length: 78
    Identification: 0x0000 (0)
    000. .... = Flags: 0x0
        0... .... = Reserved bit: Not set
        .0.. .... = Don't fragment: Not set
        ..0. .... = More fragments: Not set
    ...0 0000 0000 0000 = Fragment Offset: 0
    Time to Live: 64
    Protocol: any local network (63)
    Header Checksum: 0x4db0 [validation disabled]
    [Header checksum status: Unverified]
    Source Address: 10.136.1.32
    Destination Address: 10.2.23.24
Data (58 bytes)

0000  00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f   ................
0010  10 11 12 13 14 15 16 17 18 19 1a 1b 1c 1d 1e 1f   ................
0020  20 21 22 23 24 25 26 27 28 29 2a 2b 2c 2d 2e 2f    !"#$%&'()*+,-./
0030  30 31 32 33 34 35 36 37 38 39                     0123456789
    Data: 000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20212223â~@¦
    [Length: 58]
```
With this change:
```
Frame 2: 110 bytes on wire (880 bits), 110 bytes captured (880 bits)
    [...]
    [Protocols in frame: eth:ethertype:avsp]
Ethernet II, Src: CIMSYS_33:44:55 (00:11:22:33:44:55), Dst: 00:00:00_00:00:01 (00:00:00:00:00:01)
    [...]
Arista Vendor Specific Protocol
    Sub Type: timestamp (1)
    Version: Unknown (0x0011)
    [Expert Info (Warning/Sequence): Unknown timestamp version: 0x11]
        [Unknown timestamp version: 0x11]
        [Severity level: Warning]
        [Group: Sequence]
```

fenner:
> It's unclear for me what the problem is.

tcpdump parses the hwinfo field, but, the original pcap has no values other than zero in the hwinfo field.  This pcap includes values of 1 in the hwinfo field.  See the changes in a807516f .  The pull request that change refers to included this updated pcap file, in order to have tests for the new hwinfo functionality.

tshark does not have similar changes, so does not parse the updated file properly.

fxlb:
In PR #955, there is:
"Recently, Arista divides the two-byte protocol version field further into two fields"
Does the current code support the old and new protocol?

fenner:
> Does the current code support the old and new protocol?

Yes, the value 0 in the hwinfo field is "old", and non-zero is "new".  The change here is to add non-zero values to the pcap.

fxlb:
Thus it's perhaps better to keep arista_ether.pcap renamed to arista_ether_old.pcap (and the .out files) and add the new arista_ether.pcap with the corresponding .out files.

infrastation:
It is difficult to recall now why I did not use the updated `arista_ether.pcap` file from pull request #955, but the proposed change indeed uses that file from that pull request to replace the existing `arista_ether.pcap` file. This exercises an additional code point and the new file includes both old and new packet formats, so it looks very reasonable.

fenner:
Since the modified `arista_ether.pcap` contains both values 0 and 1, I do not understand the proposal to keep the old one that just has zero values.  In my opinion, the version that covers both 0 and 1 code points is the only version that is needed.  (This means that I do not understand fxlb's suggestion and I agree with infrastation that my pull request is reasonable ;-) )

fxlb:
OK, I understand better now.

infrastation:
Thank you!

fxlb:
Should the commit be cherry-picked in the 4.99 branch?

infrastation:
Possibly, although the effect is not user-visible and there is no urgency to it.

Issue #996, opened 2022-07-08
Haiku build is broken in various ways


infrastation:
Tested using the current master branches and the most up to date Haiku R1/beta3:
```
Haiku shredder 1 hrev55181+66 Jul  6 2022 08:03: x86_64 x86_64 Haiku
```

Compiling tcpdump with local libpcap does not work due to the-tcpdump-group/libpcap#1114.

Compiling tcpdump with system libpcap does not work because GCC linker does not seem to use Haiku directories for shared libraries and Clang seems to use the GCC linker.  The system libpcap includes the following files (in Haiku `/boot` is the boot volume and roughly an equivalent of `/` in Unix-like systems; there is no `/usr`):
```
~/gitrepos/tcpdump> find /boot/system/ -name '*pcap*'
/boot/system/package-links/libpcap-1.8.1-4
/boot/system/package-links/tcpdump-4.99.1-1/lib~libpcap
/boot/system/packages/libpcap-1.8.1-4-x86_64.hpkg
/boot/system/lib/libpcap.so.1.8.1
/boot/system/lib/libpcap.so.1
/boot/system/documentation/man/man5/pcap-savefile.5
/boot/system/documentation/man/man1/pcap-config.1
/boot/system/documentation/man/man7/pcap-tstamp.7
/boot/system/documentation/man/man7/pcap-linktype.7
/boot/system/documentation/man/man7/pcap-filter.7
/boot/system/documentation/man/man3/pcap_tstamp_type_val_to_name.3pcap
/boot/system/documentation/man/man3/pcap_tstamp_type_val_to_description.3pcap
/boot/system/documentation/man/man3/pcap_tstamp_type_name_to_val.3pcap
/boot/system/documentation/man/man3/pcap_strerror.3pcap
/boot/system/documentation/man/man3/pcap_statustostr.3pcap
/boot/system/documentation/man/man3/pcap_stats.3pcap
/boot/system/documentation/man/man3/pcap_snapshot.3pcap
/boot/system/documentation/man/man3/pcap_setnonblock.3pcap
/boot/system/documentation/man/man3/pcap_setfilter.3pcap
/boot/system/documentation/man/man3/pcap_setdirection.3pcap
/boot/system/documentation/man/man3/pcap_set_tstamp_type.3pcap
/boot/system/documentation/man/man3/pcap_set_tstamp_precision.3pcap
/boot/system/documentation/man/man3/pcap_set_timeout.3pcap
[...]

PR #1028, opened 2023-01-14
autoconf: search for gethostbyaddr() in libnetwork

davidkaroly:
This is needed for building on Haiku.

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -1166,7 +1166,7 @@ AC_DEFUN(AC_LBL_LIBRARY_NET, [
     # libraries (i.e. libc):
     # Some OSes (eg. Solaris) place it in libnsl
     # Some strange OSes (SINIX) have it in libsocket:
-    AC_SEARCH_LIBS(gethostbyaddr, nsl socket resolv)
+    AC_SEARCH_LIBS(gethostbyaddr, network nsl socket resolv)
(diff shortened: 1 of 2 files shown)


guyharris:
I don't see any calls to `socket()` in the tcpdump source; is the problem that some library *called* by tcpdump calls `socket()`?

infrastation:
It would help to know the exact steps to reproduce the problem, please see `CONTRIBUTING.md`. See also #996, maybe you would like to update it with more detail.

davidkaroly:
> I don't see any calls to `socket()` in the tcpdump source; is the problem that some library _called_ by tcpdump calls `socket()`?

It turns out that `-lnetwork` is needed for `getservent()`

Should we do something like this? in `AC_LBL_LIBRARY_NET`:

```
AC_SEARCH_LIBS(getservent, network)
```

guyharris:
Is `-lnetwork` needed for `gethostbyaddr()`?

davidkaroly:
> Is `-lnetwork` needed for `gethostbyaddr()`?

yes that one as well

davidkaroly:
Sorry about the back and forth, I just noticed the latest commit on master.

We can simply add `network` to the search libs list for `gethostbyaddr` - I guess this is the cleanest way of adding `-lnetwork`

PR #1030, opened 2023-01-22
build_common: fix missing ldd on Haiku

davidkaroly:
Haiku does not have an `ldd` script. This change allows us to get at least some successful cases from the build matrix by using `objdump` instead of `ldd`. Should help with #996.

### DIFF
--- a/build_common.sh
+++ b/build_common.sh
@@ -248,6 +248,9 @@ print_so_deps() {
     Darwin-*)
         run_after_echo otool -L "${1:?}"
         ;;
+    Haiku-*)
+        run_after_echo objdump -p "${1:?}"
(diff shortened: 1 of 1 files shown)


build_common.sh
```diff
@@ -248,6 +248,9 @@ print_so_deps() {
     Darwin-*)
         run_after_echo otool -L "${1:?}"
         ;;
+    Haiku*)
```
infrastation, line 251: `Haiku-*` is supposed to work, does it?

davidkaroly: You're right we can use `Haiku-*` - I'll update.

infrastation:
Thank you.

PR #1031, opened 2023-01-30
require suitable printf()

infrastation:
Do not merge yet, but feel free to criticise.  If these changes still hold in a couple weeks, I will rebase and merge.

At first let's see if this passes all CI (especially Appveyor).


infrastation:
This discusses `printf()`, but the actual test is for `snprintf()`, so the messages and the comments need a bit of clarification.

infrastation:
This now incorporates all the fixups I intended to make.

infrastation:
For reference, here are examples of the output:
* Autoconf on Solaris 9
  ```
  checking for vsnprintf... yes
  checking for snprintf... yes
  checking whether snprintf is suitable... no
  configure: error: The snprintf(3) implementation in this libc is not suitable,
  tcpdump would not work correctly even if it managed to compile.
  ```
* Autoconf on Linux
  ```
  checking for vsnprintf... yes
  checking for snprintf... yes
  checking whether snprintf is suitable... yes
  ```
* CMake on Linux
  ```
  -- Looking for vsnprintf
  -- Looking for vsnprintf - found
  -- Looking for snprintf
  -- Looking for snprintf - found
  -- Performing Test SUITABLE_SNPRINTF
  -- Performing Test SUITABLE_SNPRINTF - Success
  ``

fxlb:
Should we cherry-pick it in the 4.99 branch?

infrastation:
Yes, as discussed on the list.

PR #1032, opened 2023-01-31
Have a common routine for converting dates and times to strings.

guyharris:
Have a routine that takes a buffer, a strftime format, and a struct tm * as arguments, and:

* checks whether the struct tm * is null and, if so, returns a string indicating that the date and time couldn't be converted;

* otherwise, passes it to strftime(), along with the buffer and the format argument and, if strftime() returns 0, meaning the string didn't fit into the buffer and thus that the buffer's contents are undefined, returns a string indicating that the date and time didn't fit into the buffer;

* otherwise, returns a pointer to the buffer.

Call this routine instead of directly calling strftime() in printers; that prevents printing a buffer with undefined data if the buffer isn't big enough for the string.

Also, when generating file names using an strftime format, check the return value of strftime() to make sure the buffer didn't overflow.

### DIFF
--- a/ntp.c
+++ b/ntp.c
@@ -54,8 +54,8 @@ p_ntp_time(netdissect_options *ndo,
 	if (i) {
 	    int64_t seconds_64bit = (int64_t)i - JAN_1970;
 	    time_t seconds;
-	    struct tm *tm;
 	    char time_buf[128];
+	    const char *time_string;
 
 	    seconds = (time_t)seconds_64bit;
 	    if (seconds != seconds_64bit) {
@@ -63,21 +63,12 @@ p_ntp_time(netdissect_options *ndo,
 		 * It doesn't fit into a time_t, so we can't hand it
 		 * to gmtime.
 		 */
-		ND_PRINT(" (unrepresentable)");
+		time_string = "[Time is too large to fit into a time_t]";
 	    } else {
-		tm = gmtime(&seconds);
-		if (tm == NULL) {
-		    /*
-		     * gmtime() can't handle it.
-		     * (Yes, that might happen with some version of
-		     * Microsoft's C library.)
-		     */
-		    ND_PRINT(" (unrepresentable)");
-		} else {
-		    /* use ISO 8601 (RFC3339) format */
-		    strftime(time_buf, sizeof (time_buf), "%Y-%m-%dT%H:%M:%SZ", tm);
-		    ND_PRINT(" (%s)", time_buf);
-		}
+		/* use ISO 8601 (RFC3339) format */
+		time_string = nd_format_time(time_buf, sizeof (time_buf),
+		  "%Y-%m-%dT%H:%M:%SZ", gmtime(&seconds));
 	    }
+	    ND_PRINT(" (%s)", time_string);
 	}
 }
--- a/print-ahcp.c
+++ b/print-ahcp.c
@@ -102,18 +102,14 @@ ahcp_time_print(netdissect_options *ndo,
                 const u_char *cp, uint8_t len)
 {
 	time_t t;
-	struct tm *tm;
-	char buf[BUFSIZE];
+	char buf[sizeof("-yyyyyyyyyy-mm-dd hh:mm:ss UTC")];
 
 	if (len != 4)
 		goto invalid;
 	t = GET_BE_U_4(cp);
-	if (NULL == (tm = gmtime(&t)))
-		ND_PRINT(": gmtime() error");
-	else if (0 == strftime(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S", tm))
-		ND_PRINT(": strftime() error");
-	else
-		ND_PRINT(": %s UTC", buf);
+	ND_PRINT(": %s",
+	    nd_format_time(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S UTC",
+	      gmtime(&t)));
 	return;
 
 invalid:
--- a/print-arista.c
+++ b/print-arista.c
@@ -10,7 +10,6 @@
 
 #include "netdissect.h"
 #include "extract.h"
-#include "addrtoname.h"
 
 /*
 
@@ -93,17 +92,13 @@ arista_print_date_hms_time(netdissect_options *ndo, uint32_t seconds,
 		uint32_t nanoseconds)
 {
 	time_t ts;
-	struct tm *tm;
-	char buf[BUFSIZE];
+	char buf[sizeof("-yyyyyyyyyy-mm-dd hh:mm:ss")];
 
 	ts = seconds + (nanoseconds / 1000000000);
 	nanoseconds %= 1000000000;
-	if (NULL == (tm = gmtime(&ts)))
-		ND_PRINT("gmtime() error");
-	else if (0 == strftime(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S", tm))
-		ND_PRINT("strftime() error");
-	else
-		ND_PRINT("%s.%09u", buf, nanoseconds);
+	ND_PRINT("%s.%09u",
+	    nd_format_time(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S",
+	       gmtime(&ts)), nanoseconds);
 }
 
 int
--- a/print-rx.c
+++ b/print-rx.c
@@ -794,12 +794,12 @@ rx_cache_find(netdissect_options *ndo, const struct rx_header *rxh,
 			ND_PRINT(" %" PRIu64, _i); \
 		}
 
-#define DATEOUT() { time_t _t; struct tm *tm; char str[256]; \
+#define DATEOUT() { time_t _t; char str[256]; \
 			_t = (time_t) GET_BE_S_4(bp); \
 			bp += sizeof(int32_t); \
-			tm = localtime(&_t); \
-			strftime(str, 256, "%Y/%m/%d %H:%M:%S", tm); \
-			ND_PRINT(" %s", str); \
+			ND_PRINT(" %s", \
+			    nd_format_time(str, sizeof(str), \
+			      "%Y/%m/%d %H:%M:%S", localtime(&_t))); \
 		}
 
 #define STOREATTROUT() { uint32_t mask, _i; \
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -833,6 +833,8 @@ MakeFilename(char *buffer, char *orig_name, int cnt, int max_chars)
         char *filename = malloc(PATH_MAX + 1);
         if (filename == NULL)
             error("%s: malloc", __func__);
+        if (strlen(orig_name) == 0)
+            error("an empty string is not a valid file name");
 
         /* Process with strftime if Gflag is set. */
         if (Gflag != 0) {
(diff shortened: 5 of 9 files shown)


infrastation:
(The failure on netbsd-mips64 was due to a compiler segfault, rebuilding now.)

util-print.c
```diff
@@ -422,6 +422,60 @@ signed_relts_print(netdissect_options *ndo,
...
+nd_format_time(char *buf, size_t bufsize, const char *format,
+         const struct tm *timeptr)
+{
+	if (timeptr != NULL) {
+		if (strftime(buf, bufsize, format, timeptr) != 0)
+			return (buf);
+		else
+			return ("[Does not fit in buffer]");
+	} else
+		return ("[Unrepresentable]");
```
infrastation, line 442: I was going to make a point that it might be more consistent to mimic `printf("%s", NULL)`, which prints a `(null)` string in similar cases, but apparently today on Linux is just segfaults like in old times, so the argument collapsed.

Also, if I was writing this code, I would flip `if` branches around to have the shortest hand first, and then would lose `else` after `return`, which would make the code easier to follow. But for such a small function it is a non-issue.

tcpdump.c
```diff
@@ -844,9 +844,24 @@ MakeFilename(char *buffer, char *orig_name, int cnt, int max_chars)
...
+           * strftime() if it's non-empty, in which case the return
+           * value will only be 0 if the formatted date doesn't fit
+           * in the buffer.
            */
-          strftime(filename, PATH_MAX, orig_name, local_tm);
+          if (strlen(orig_name) != 0) {
+	    if (0 == strftime(filename, PATH_MAX, orig_name, local_tm))
+	      error("%s: strftime", __func__);
+	  } else {
+	    strcpy(filename, orig_name);
```
infrastation, line 863: This needs to be `strncpy()`, just like below.

print-arista.c
```diff
@@ -93,17 +93,13 @@ arista_print_date_hms_time(netdissect_options *ndo, uint32_t seconds,
...
 	nanoseconds %= 1000000000;
-	if (NULL == (tm = gmtime(&ts)))
-		ND_PRINT("gmtime() error");
-	else if (0 == strftime(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S", tm))
-		ND_PRINT("strftime() error");
-	else
-		ND_PRINT("%s.%09u", buf, nanoseconds);
+	ND_PRINT("%s",
+	    nd_format_time_with_nanoseconds(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S",
+	       gmtime(&ts), nanoseconds));
```
infrastation, line 102: The same can be done simpler:
```c
	ND_PRINT("%s.%09u",
	    nd_format_time(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S", gmtime(&ts)),
	    nanoseconds);
```
This way `nd_format_time_with_nanoseconds()` would not have to detect ns values greater than 10**9, would not have to squeeze the fractional part into the buffer and basically would not have reasons to exist.

guyharris: ...and you'd run the risk of getting output such as

    [Does not fit in buffer].013657019

rather than just

    [Does not fit in buffer]

print-rx.c
```diff
@@ -794,12 +794,12 @@ rx_cache_find(netdissect_options *ndo, const struct rx_header *rxh,
...
 
-#define DATEOUT() { time_t _t; struct tm *tm; char str[256]; \
+#define DATEOUT() { time_t _t; char str[256]; \
 			_t = (time_t) GET_BE_S_4(bp); \
 			bp += sizeof(int32_t); \
-			tm = localtime(&_t); \
-			strftime(str, 256, "%Y/%m/%d %H:%M:%S", tm); \
-			ND_PRINT(" %s", str); \
+			ND_PRINT(" %s", \
+			    nd_format_time(str, 256, "%Y/%m/%d %H:%M:%S", \
```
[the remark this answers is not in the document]
guyharris: Yes, we might as well also clean up the calls to `strftime()` when changing them to calls to `nd_format_time()`.  Fixed.

guyharris:
> Also the commit message has two typos ("::" and "inticating").

Fixed.

print-ahcp.c
```diff
@@ -102,18 +102,14 @@ ahcp_time_print(netdissect_options *ndo,
                 const u_char *cp, uint8_t len)
 {
 	time_t t;
-	struct tm *tm;
 	char buf[BUFSIZE];
```
fxlb, line 105: BUFSIZE is defined in addrtoname.h. It should not be used for this usage.

print-arista.c
```diff
@@ -93,17 +93,13 @@ arista_print_date_hms_time(netdissect_options *ndo, uint32_t seconds,
 		uint32_t nanoseconds)
 {
 	time_t ts;
-	struct tm *tm;
 	char buf[BUFSIZE];
```
fxlb, line 96: Same.

tcpdump.c
guyharris: Fixed.

print-arista.c
```diff
@@ -93,17 +93,13 @@ arista_print_date_hms_time(netdissect_options *ndo, uint32_t seconds,
 		uint32_t nanoseconds)
 {
 	time_t ts;
-	struct tm *tm;
-	char buf[BUFSIZE];
+	char buf[128];
```
fxlb, line 96: OK. Thus `#include "addrtoname.h"` can be removed.

guyharris: Removed.  I also changed `128` to `sizeof(<example string>)`.

print-ahcp.c
guyharris: Fixed.

print-arista.c
guyharris: Fixed.

tcpdump.c
```diff
@@ -844,9 +844,24 @@ MakeFilename(char *buffer, char *orig_name, int cnt, int max_chars)
...
+           * so we must check for a buffer overflow.
+           *
+           * So we check for an empty orig_name, and only call
+           * strftime() if it's non-empty, in which case the return
+           * value will only be 0 if the formatted date doesn't fit
+           * in the buffer.
            */
-          strftime(filename, PATH_MAX, orig_name, local_tm);
+          if (strlen(orig_name) != 0) {
+	    if (0 == strftime(filename, PATH_MAX, orig_name, local_tm))
```
fxlb, line 860: Style: we have `xxx != 0`, we should have `yyy == 0`, not `0 == yyy`

guyharris: Done.

print-arista.c
infrastation: ...which would say "the fractional part is definitely this many nanoseconds, but the integer part is not a even a valid date, so you would need to inspect the packet closer to understand what this timestamp means".

util-print.c
```diff
@@ -422,6 +422,60 @@ signed_relts_print(netdissect_options *ndo,
...
+		 * doesn't fit in the buffer.  Otherwise, append the
+		 * nanoseconds.
+		 */
+#define NANOSECONDS_LEN	(9+1)	/* .000000000 */
+		if (ftimelen != 0 && bufsize > ftimelen + NANOSECONDS_LEN) {
+			snprintf(buf + ftimelen, bufsize - ftimelen,
+			    ".%09u", nanoseconds);
+			return (buf);
+		} else
+			return ("does not fit in buffer");
```
fxlb, line 474: "[Does not fit in buffer]" like above.

util-print.c
```diff
@@ -422,6 +422,60 @@ signed_relts_print(netdissect_options *ndo,
...
+	} else
+		return ("unrepresentable");
```
fxlb, line 476: "[Unrepresentable]" like above.

ntp.c
```diff
@@ -65,19 +64,10 @@ p_ntp_time(netdissect_options *ndo,
 		 */
 		ND_PRINT(" (unrepresentable)");
```
fxlb, line 65: Also " [Unrepresentable]"

infrastation:
Getting a pull request merged in this project can be remarkably difficult.

print-arista.c
guyharris: > "integer part is not a even a valid date"

Given the current code, that would only be the case if the integer part were a negative number with an absolute value so great that the year was before 0 CE, and I'm not sure that `[Does not fit in buffer].013657019` would clearly indicate that. `-1-01-31 00:23:59..013657019` would probably indicate that better; I've adjusted the sample string to allow negative years, as `tm_year` is the only field that doesn't have a strict limit on values.  That means that, at least for the one place that uses `nd_format_time_with_nanoseconds()`, "Does not fit in buffer" means "there's a bug in the code" (or in `localtime()` or `gmtime()`), not "the integer part isn't a valid date".

The one place where the `struct tm` isn't generated from a `time_t` is the DOS-style date processing in `interpret_dos_date()` in smbutil.c, but that keeps all fields other than `tm_year` in the range 0-99, so, again, bogus dates and times will show up as 2023-01-32 23:66:77 or something such as that, not as "Does not fit in buffer".

For `[Unrepresentable].013657019`, that is, in practice, an indication that 1) this is on Windows and Microsoft have decided not to support converting pre-UNIX Epoch times in `localtime()` and `gmtime()` and 2) the time is pre-UNIX Epoch.  That's an indication of a misfeature of MSVC; on a UN\*X, it'd show up as 1957-01-12 16:42:86 or something such as that, which is a valid-but-surprising date (but I know of at least one organization that used file system date fields to record events that long predated computers, so it's not inconceivable that they'd show up in SMB packets, given that they *did* show up in SMB packets).

util-print.c
guyharris: Fixed.

util-print.c
guyharris: Fixed.

ntp.c
guyharris: Fixed.

guyharris:
> Getting a pull request merged in this project can be remarkably difficult.

"Difficult" as in "a lot of people point out problems with the code", or "difficult" as in "the mechanisms provided by Git and GitHub make it overly complicated"? The former is arguably a feature; the latter is helped, on my side, with some scripts wherein, after I've set up my forked repository, I can do:

- `git reviewbranch <branch name>` to create a branch for the changes;
- `git doreview` after committing those changes or committing updates with `git commit --amend`;
- `git reviewdone` when the change has been committed to the main repository.

They were originally developed when Wireshark moved to GitLab, but it turned out the basic commands they do work the same with GitHub.

print-arista.c
infrastation: Precise time processing is a difficult subject, and it would be fine to let the code give up and fail safe on a strange value. But if you see it differently, make your try.

guyharris: OK, I've gotten rid of `nd_format_time_with_nanoseconds()` and also changed a bunch of error strings to indicate the underlying problem making it impossible to print the date/time in some reasonable form.

infrastation: That makes more sense to me, please note that the removed function still appears in `netdissect.h` and in the commit message. Speaking of which, the message also contains a "the the".

infrastation:
That was my attempt of a philosophical remark about software criticism. These changes certainly deserve another review round after a good sleep.

print-arista.c
guyharris: Fixed.

infrastation: Thank you.

fxlb:
Message on  CI netbsd-mips64host:
```
/home/buildbot-worker/netbsd-mips64/tmpfs/tcpdump-netbsd-mips64/build/print-radius.c: In function ‘print_attr_string’:
/home/buildbot-worker/netbsd-mips64/tmpfs/tcpdump-netbsd-mips64/build/print-radius.c:831:7: internal compiler error: Segmentation fault
  831 |       nd_print_trunc(ndo);
      |       ^~~~~~~~~~~~~~
no stack trace because unwind library not available
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://www.NetBSD.org/support/send-pr.html> for instructions.
*** [CMakeFiles/netdissect.dir/print-radius.c.o] Error code 1
```

infrastation:
GCC segfaults on that host from time to time, I have restarted that build. I do not have a better solution for this problem now.

fxlb:
"no stack trace because unwind library not available"
Perhaps the installation of this library could give some indication.

tcpdump.c
```diff
@@ -833,6 +833,8 @@ MakeFilename(char *buffer, char *orig_name, int cnt, int max_chars)
         char *filename = malloc(PATH_MAX + 1);
         if (filename == NULL)
             error("%s: malloc", __func__);
+        if (strlen(orig_name) != 0)
```
infrastation, line 836: Did you mean `== 0`?

guyharris: Yes.  Thanks. Fixed.

infrastation:
If there is anything else that prevents these changes from going into the master branch, I cannot see it.

fxlb:
(Rebased on master.)

fxlb:
Fix patch line 211: space before tab in indent.

fxlb:
It seems good.

infrastation:
Thanks to everyone.

fxlb:
Thank you!

guyharris:
This should probably be backported to 4.99.

fxlb:
> This should probably be backported to 4.99.

Agreed.

PR #1036, opened 2023-02-22
Detect OS IPv6 support using AF_INET6 only.

infrastation:
Let's see if this passes CI cleanly. A couple additional notes:
* There's one call to `cmake_pop_check_state()` missing from `CMakeLists.txt`, I guess it should be just before "Check for various functions in libpcap/WinPcap/Npcap.", but this discrepancy is old and seems to be irrelevant to this commit.
* The pre-existing `check_type_size("struct in6_addr" HAVE_STRUCT_IN6_ADDR)` line remains because without it my CMake fails to configure as shown below without an apparent reason. I tried to debug it, but both with and without the line the value of `CMAKE_REQUIRED_LIBRARIES` seems to be the same, so I cannot explain this failure, but it reproduces steadily.
```
-- Looking for sys/types.h
CMake Error (dev) at /usr/share/cmake-3.18/Modules/CheckIncludeFile.cmake:80 (message):
  Policy CMP0075 is not set: Include file check macros honor
  CMAKE_REQUIRED_LIBRARIES.  Run "cmake --help-policy CMP0075" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  CMAKE_REQUIRED_LIBRARIES is set to:

    /usr/lib/x86_64-linux-gnu/libpcap.so

  For compatibility with CMake 3.11 and below this check is ignoring it.
Call Stack (most recent call first):
  /usr/share/cmake-3.18/Modules/CheckTypeSize.cmake:230 (check_include_file)
  CMakeLists.txt:764 (check_type_size)
This error is for project developers. Use -Wno-error=dev to suppress it.

-- Configuring incomplete, errors occurred!
```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -48,6 +48,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       At build time require a proof of suitable snprintf(3) implementation in
         libc (and document Solaris 9 as unsupported because of that).
       Autoconf: Remove detection of early IPv6 stacks.
+      Detect OS IPv6 support using AF_INET6 only.
 
 DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
   Summary for 4.99.4 tcpdump release (so far!)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -633,22 +633,10 @@ cmake_pop_check_state()
 #
 
 #
-# Check for IPv6 support.
-# We just check for AF_INET6 and struct in6_addr.
+# FIXME: This check does not influence the build logic, but without it CMake
+# 3.18.4 fails trying to make the next check_type_size() check later on.
 #
-cmake_push_check_state()
-if(WIN32)
-    set(CMAKE_EXTRA_INCLUDE_FILES sys/types.h ws2tcpip.h)
-    check_symbol_exists(AF_INET6 "sys/types.h;ws2tcpip.h" HAVE_AF_INET6)
-else(WIN32)
-    set(CMAKE_EXTRA_INCLUDE_FILES sys/types.h sys/socket.h netinet/in.h)
-    check_symbol_exists(AF_INET6 "sys/types.h;sys/socket.h;netinet/in.h" HAVE_AF_INET6)
-endif(WIN32)
 check_type_size("struct in6_addr" HAVE_STRUCT_IN6_ADDR)
-cmake_pop_check_state()
-if(HAVE_AF_INET6 AND HAVE_STRUCT_IN6_ADDR)
-    set(HAVE_OS_IPV6_SUPPORT TRUE)
-endif(HAVE_AF_INET6 AND HAVE_STRUCT_IN6_ADDR)
 
 ######################################
 # External dependencies
--- a/addrtoname.c
+++ b/addrtoname.c
@@ -148,14 +148,15 @@ win32_gethostbyaddr(const char *addr, int len, int type)
 	static struct hostent host;
 	static char hostbuf[NI_MAXHOST];
 	char hname[NI_MAXHOST];
-	struct sockaddr_in6 addr6;
 
 	host.h_name = hostbuf;
 	switch (type) {
 	case AF_INET:
 		return gethostbyaddr(addr, len, type);
 		break;
-	case AF_INET6:
+#ifdef AF_INET6
+	case AF_INET6: {
+		struct sockaddr_in6 addr6;
 		memset(&addr6, 0, sizeof(addr6));
 		addr6.sin6_family = AF_INET6;
 		memcpy(&addr6.sin6_addr, addr, len);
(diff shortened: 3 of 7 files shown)


infrastation:
This time with a more useful commit message. This revision is going to be merged anytime soon.

guyharris:
> There's one call to cmake_pop_check_state() missing from CMakeLists.txt, I guess it should be just before "Check for various functions in libpcap/WinPcap/Npcap."

Or after all the libpcap function checks; I've added an additional one in #1037.

I want to make more use of pkg-config in tcpdump configuration, to make it easier to support static builds (that would also involve other work, such as trying to *detect* that the user is doing a static build).  I'll look at cleaning up the check state stuff as part of that.

infrastation:
Thank you.

guyharris:
> The pre-existing `check_type_size("struct in6_addr" HAVE_STRUCT_IN6_ADDR)` line remains because without it my CMake fails to configure as shown below without an apparent reason. I tried to debug it, but both with and without the line the value of `CMAKE_REQUIRED_LIBRARIES` seems to be the same, so I cannot explain this failure, but it reproduces steadily.

There is an explanation, but it's not an obvious one; I found it with the classic debugging technique of "inserting printouts".

Apparently, `check_type_size()` checks (at least on UN*Xes) for the existence of sys/types.h, and the results of that check are remembered, so that subsequent checks for sys/types.h aren't done.

If you do the `check_type_size("struct in6_addr" HAVE_STRUCT_IN6_ADDR)` at that point, by the time it gets to the `check_type_size(pcap_if_t PCAP_IF_T)` the check for sys/types.h has already been done, and isn't done again.  The first check for sys/types.h is done with `CMAKE_REQUIRED_LIBRARIES` not set, so no warning is produced.

If you *don't* do the `check_type_size("struct in6_addr" HAVE_STRUCT_IN6_ADDR)` at that point, by the time it gets to the `check_type_size(pcap_if_t PCAP_IF_T)` the check for sys/types.h has *not* been done, and has to be done. *That* check is done with `CMAKE_REQUIRED_LIBRARIES` set to include the pcap library, so the warning is produced.

guyharris:
[Policy CMP0075](https://cmake.org/cmake/help/latest/policy/CMP0075.html) was introduced in CMake 3.12, and causes the include file check macros to link the check executable to the libraries listed in the `CMAKE_REQUIRED_LIBRARIES` variable - those macros don't just compile a test C program, they link it.

CMake policies that change CMake behavior seem to be tricky to deal with:

- If you go with the `OLD` behavior, which is usually the default if the policy isn't specified and if that behavior hasn't been removed, your CMake files will behave, at least for this policy, in the same way in CMake releases prior to the new policy and in CMake releases with the new policy, but something undesirable might happen in the future in a release in which the old behavior is removed.
- If you go with the `NEW` behavior,  your CMake files will behave, at least for this policy, differently in CMake releases prior to the new policy and in CMake releases with the new policy, which might be undesirable.

My inclination, at least for now, is to go with the `OLD` policy.  If a future release removes the policy, presumably we'd have to go with the `NEW` policy and, if that doesn't work in releases where the new policy is missing, either work around it if possile or require 3.12 or later.

infrastation:
Oh. That's why before the failure the last line was `Looking for sys/types.h`, not `Check size of pcap_if_t`, now it makes sense. It may be possible to avoid the CMake policies complication space using at least one of the following:
* The check for `pcap_if_t` could move before the changes to `CMAKE_REQUIRED_LIBRARIES` because the check for `pcap_if_t` does not depend on availability of `pcap_findalldevs()` and its result is not an input to any other configuration logic, `tcpdump.c` relates the two outputs at the compile time.
* The check for `<sys/types.h>` could be done separately beforehand instead of the current check for `in6_addr` if it populates the same cache and this implicit CMake behaviour is reliable.
* The current phony check for `in6_addr` could remain with a better comment under the same condition as above.

guyharris:
> The check for `pcap_if_t` could move before the changes to `CMAKE_REQUIRED_LIBRARIES` because the check for `pcap_if_t` does not depend on availability of `pcap_findalldevs()`

...and if you don't have `pcap_if_t` defined in pcap.h, you can't *use* `pcap_findalldevs()`, because it takes a pointer to a pointer to a `pcap_if_t` as an argument.  That check is there because, as the comments in configure.ac and CMakeLists.txt note, Apple picked up a new version of libpcap, probably to fix a bug, but didn't update the pcap headers in their SDK because, as I remember, they don't do that for dot-dot releases, rendering `pcap_findalldevs()` unusable.

So I'd vote for doing the `pcap_if_t` check first (with a comment noting the CMake reason why we do that).

But the policy stuff might be a good idea anyway, just to ensure that we don't have to debug weird problems with header file searches that work with some versions of CMake and don't work with others - if the `NEW` behavior isn't necessary, we might as well turn it off, and if it ever ends up being necessary, that means we need to make the minimum CMake version a version that has the policy and turn it on.

PR #1038, opened 2023-02-24
Untangle detection of pcap_findalldevs().

infrastation:
This is the solution I meant in the follow-up discussion in pull request #1036.


infrastation:
Alternatively, with the same two separate checks this
```c
#ifdef HAVE_PCAP_FINDALLDEVS
#ifndef HAVE_PCAP_IF_T
#undef HAVE_PCAP_FINDALLDEVS
#endif
#endif
```
could be this:
```c
#ifndef HAVE_PCAP_IF_T
struct pcap_if {
        struct pcap_if *next;
        char *name;             /* name to hand to "pcap_open_live()" */
        char *description;      /* textual description of interface, or NULL */
        struct pcap_addr *addresses;
        bpf_u_int32 flags;      /* PCAP_IF_ interface flags */
};
typedef struct pcap_if pcap_if_t;
#define HAVE_PCAP_IF_T 1
#endif
```

guyharris:
I vote for "disregard the incomplete `pcap_findalldevs()`", as it's only some old version of macOS (old enough that it was, I think, still called "Mac OS X", not even "OS X") - I *probably* have a VM with that version on it, and I can see whether  we've gotten to the point yet where it's to old to support compiling the current tcpdump.

infrastation:
Meanwhile do you see any reason not to make the change in the commit?

guyharris:
> Meanwhile do you see any reason not to make the change in the commit?

No, I don't.

infrastation:
Thank you for identifying the root cause.

PR #1039, opened 2023-03-05
IKEv2: add missing notification types

mspncp:
The notification type entries were crafted from the CSV files provided by IANA for the [error types] resp. [status types].

[error types]:  https://www.iana.org/assignments/ikev2-parameters/ikev2-parameters-14.csv
[status types]: https://www.iana.org/assignments/ikev2-parameters/ikev2-parameters-16.csv

### DIFF
--- a/print-isakmp.c
+++ b/print-isakmp.c
@@ -385,32 +385,88 @@ struct ikev2_n {
 };
 
 enum ikev2_n_type {
-	IV2_NOTIFY_UNSUPPORTED_CRITICAL_PAYLOAD            = 1,
-	IV2_NOTIFY_INVALID_IKE_SPI                         = 4,
-	IV2_NOTIFY_INVALID_MAJOR_VERSION                   = 5,
-	IV2_NOTIFY_INVALID_SYNTAX                          = 7,
-	IV2_NOTIFY_INVALID_MESSAGE_ID                      = 9,
-	IV2_NOTIFY_INVALID_SPI                             =11,
-	IV2_NOTIFY_NO_PROPOSAL_CHOSEN                      =14,
-	IV2_NOTIFY_INVALID_KE_PAYLOAD                      =17,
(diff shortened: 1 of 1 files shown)


print-isakmp.c
```diff
@@ -2386,7 +2442,7 @@ ikev2_n_print(netdissect_options *ndo, u_char tpay _U_,
 	ND_TCHECK_SIZE(p);
 	ikev2_pay_print(ndo, NPSTR(ISAKMP_NPTYPE_N), GET_U_1(p->h.critical));
 
-	showspi = 1;
+	showspi=0;
```
mspncp, line 2445: I changed the initial value of `showspi` in order to make the cases of the switch-statement less noisy. For the existing cases, the value of `showspi` should be unchanged, unless I made a mistake. For the new cases, It's not clear to me according to which criterion the SPI should be printed or not.

Wouldn't it make more sense to always print the SPI if it is present? I.e., to change the line
```
if (showspi && spi_size) {
```
below to

```
if (spi_size) {
```

Maybe you can enlighten me @mcr?   You introduced it in 650e708afb8c6b20d49dfec794a238eac9c70736.

mspncp: I removed the spaces around the assignment because the two assignments below don't have them either. Alternatively, I could add surrounding spaces to all three assignments.

print-isakmp.c
```diff
@@ -2427,7 +2481,7 @@ ikev2_n_print(netdissect_options *ndo, u_char tpay _U_,
 		break;
 
 	case IV2_NOTIFY_NO_PROPOSAL_CHOSEN:
-		notify_name = "no_protocol_chosen";
+		notify_name = "no_proposal_chosen";
```
mspncp, line 2484: This typo here bugged my and got me started ;-)

mspncp:
ping?

mcr (approved):

PR #1045, opened 2023-04-18
Fix for DHCPv6 Duid type 2 values

eamonjd:
DHCPv6 client-id DUID type 2 support. Previously tcpdump chopped DUID type 2 values after 8 octets, e.g. (client-ID vid 0000757148534831) where it should be (client-ID vid 000075714853483134343235313438).

From RFC8415, client-id is defined as follows:

21.2.  Client Identifier Option

...

  |        OPTION_CLIENTID        |          option-len           |
  .                                                               .
  .                              DUID                             .
  .                        (variable length)                      .
  .                                                               .

There are a number of DUIDs defined in section 11 and type 2 is defined as follows:

11.3.  DUID Assigned by Vendor Based on Enterprise Number (DUID-EN)

The vendor assigns this form of DUID to the device.  This DUID
   consists of the 4-octet vendor's registered Private Enterprise Number
   as maintained by IANA [IANA-PEN] followed by a unique identifier
   assigned by the vendor.  The following diagram summarizes the
   structure of a DUID-EN:

      |         DUID-Type (2)         |       enterprise-number       |
      |   enterprise-number (contd)   |                               |
      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+                               |
      .                           identifier                          .
      .                       (variable length)                       .
      .                                                               .


fenner (approved):
For those looking for the reference, this option is defined in https://datatracker.ietf.org/doc/html/rfc8415#section-11.3 .

Regarding the code, I assert both changes are correct:
- comparing `optlen >= 2 + 4` ensures that the 2 bytes of "DUID type 2" and the 4 bytes of "enterprise number" are present; the spec says that the enterprise number is "followed by a unique identifier" but it is reasonable for tcpdump in its role as a packet dumper to be able to dump a slightly-malformed packet that only has the enterprise number.
- Looping until reaching optlen prints all of the bytes of the variable-length field.

fxlb:
Thank you for preparing this PR and the review.
Why not print separately the `enterprise-number` and the `identifier`?

fenner:
> Why not print separately the `enterprise-number` and the `identifier`?

I think this is for backwards compatibility: if anyone is depending on the output that worked with the previous code, that output remains the same - this patch just allows printing other valid instances of this option using the already-defined format.

If you decide it's better to break backwards compatibility for the benefit of more precise parsing, we can update the patch.

fenner:
What do you think about this output:
`(client-ID enterprise 30065 DUID-EN 0x4853483134343235313438)`
I picked `enterprise 30065` to match the sflow parser's output, and the string `DUID-EN` based on https://datatracker.ietf.org/doc/html/rfc8415#section-11.3

fenner:
> `(client-ID enterprise 30065 DUID-EN 0x4853483134343235313438)`

I guess other examples end up being `(client-ID  hwaddr/time type %u time %u  4853483134343235313438)` or `(client-ID hwaddr type %u 4853483134343235313438)` - I added the `0x` to the type-2 because of the example in the doc, but maybe neither the string `DUID-EN` nor `0x` belong in the output for consistency with the existing type-1 and type-3 output.

fxlb:
> maybe neither the string `DUID-EN` nor `0x` belong in the output for consistency with the existing type-1 and type-3 output.

Yes, thus `enterprise 30065 4853483134343235313438` should be consistent with other types.

fxlb:
Thank you!

PR #1047, opened 2023-05-07
Update ND_BYTES_BETWEEN() and ND_BYTES_AVAILABLE_AFTER() macros for better accuracy

fxlb:
1)
Update ND_BYTES_BETWEEN() macro for better accuracy

Update the macro that computes how many bytes are present, starting
at the first argument and running up to (but not including) the second
argument, and returns that as a u_int (cutting it to 32 bits on LP64
and LLP64 platforms).

This reverses, for reasons of readability, the order of the arguments
which was based on old SMB's PTR_DIFF().

With this change the number of bytes "between" given by the macro is 0
when the first argument is greater than or equal to the second argument.

This is a follow-up to f9c2c905b118b69a0b102549c1b25cca871947b5.

2)
Update ND_BYTES_AVAILABLE_AFTER() macro for better accuracy

With this change the number of bytes available in the captured data
given by the macro is 0 when the argument is greater than or equal
to ndo_snapend or less than ndo_packetp (e.g. pointer underflow).
 
i.e.:
If p is like p3, the macro gives 0.
If p is like p1, the macro gives 0.
```
|------------------|-----------------------|-----------------|
     p1  <   ndo_packetp  <=  p2  <  ndo_snapend   <=   p3 
 ```
Update the ascii_print(), hex_and_ascii_print_with_offset() and 
hex_print_with_offset() functions accordingly.

### DIFF
--- a/netdissect.h
+++ b/netdissect.h
@@ -391,7 +391,7 @@ nd_trunc_longjmp(netdissect_options *ndo)
  * Number of bytes remaining in the captured data, starting at the
  * byte pointed to by the argument.
  */
-#define ND_BYTES_AVAILABLE_AFTER(p) ND_BYTES_BETWEEN((p), ndo->ndo_snapend)
+#define ND_BYTES_AVAILABLE_AFTER(p) ((const u_char *)(p) < ndo->ndo_packetp ? 0 : ND_BYTES_BETWEEN((p), ndo->ndo_snapend))
 
 /*
  * Check (expression_1 operator expression_2) for invalid packet with
--- a/netdissect.h
+++ b/netdissect.h
@@ -385,13 +385,13 @@ nd_trunc_longjmp(netdissect_options *ndo)
 /*
  * Number of bytes between two pointers.
  */
-#define ND_BYTES_BETWEEN(p1, p2) ((u_int)(((const uint8_t *)(p1)) - (const uint8_t *)(p2)))
+#define ND_BYTES_BETWEEN(p1, p2) ((const u_char *)(p1) >= (const u_char *)(p2) ? 0 : ((u_int)(((const u_char *)(p2)) - (const u_char *)(p1))))
 
 /*
  * Number of bytes remaining in the captured data, starting at the
  * byte pointed to by the argument.
  */
-#define ND_BYTES_AVAILABLE_AFTER(p) ND_BYTES_BETWEEN(ndo->ndo_snapend, (p))
+#define ND_BYTES_AVAILABLE_AFTER(p) ND_BYTES_BETWEEN((p), ndo->ndo_snapend)
 
 /*
  * Check (expression_1 operator expression_2) for invalid packet with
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: netdissect.h [hunk 1]
```diff
@@ -391,7 +391,7 @@ nd_trunc_longjmp(netdissect_options *ndo)
  * Number of bytes remaining in the captured data, starting at the
  * byte pointed to by the argument.
  */
-#define ND_BYTES_AVAILABLE_AFTER(p) ND_BYTES_BETWEEN(ndo->ndo_snapend, (p))
+#define ND_BYTES_AVAILABLE_AFTER(p) ((const u_char *)(p) >= ndo->ndo_snapend || (const u_char *)(p) < ndo->ndo_packetp ? 0 : (u_int)(ndo->ndo_snapend - (const u_char *)(p)))
```
guyharris, line 394: Or maybe

    #define ND_BYTES_AVAILABLE_AFTER(p) ((const u_char *)(p) >= ndo->ndo_snapend || (const u_char *)(p) < ndo->ndo_packetp ? 0 : ND_BYTES_BETWEEN(ndo->ndo_snapend, (p)))

fxlb: Or perhaps update first `ND_BYTES_BETWEEN`.
In f9c2c905b118b69a0b102549c1b25cca871947b5, message we have:
```
Add an ND_BYTES_BETWEEN() macro that computes how many bytes are 
present, starting at the second argument and running up to (but not 
including) the first argument, and returns that as a u_int [...]
```
If p1 >= p2: OK.
If p1 < p2: negative -> big value via the u_int cast. Should be zero (no bytes present).

fxlb:
Someting like:
`#define ND_BYTES_BETWEEN(p1, p2) ((const u_char *)(p2) >= (const u_char *)(p1) ? 0 : (u_int)((const u_char *)(p1) - (const u_char *)(p2)))`
or cast with `uint8_t *`if this really make a difference.

netdissect.h
fxlb: Someting like:
`#define ND_BYTES_BETWEEN(p1, p2) ((const u_char *)(p2) >= (const u_char *)(p1) ? 0 : (u_int)((const u_char *)(p1) - (const u_char *)(p2)))`
or cast with `uint8_t *`if this really make a difference.

fxlb: I think the purpose was 1.

guyharris: Yes, the purpose was 1.  The goal was to have a macro to find out how many bytes there were from p2 to p1 (so p1 >= p2) and return the result as a `u_int`, with an explicit cast to squelch any "that's a 64-bit quantity and you're throwing away the upper 32 bits" warnings that you get (either from Apple's compilers, which can be told to warn about 64->32 narrowings, or from Visual C/++, which can be told to warn about *all* narrowings).

fxlb: Ok. I have reverse, in the first commit, the order of the arguments because it seems to me more natural to go from p1 to p2 (no need to keep the order from the old SMB's PTR_DIFF()).

PR #1051, opened 2023-05-19
Fix spelling of PTP type SIGNALING

cappe987:
The IEEE 1588 (PTP) standard uses the spelling SIGNALING, with one L.

I understand that one L vs two L can vary depending on where in the world you are, but the standard uses one L so I think it's better to use that.


infrastation:
Thank you.

PR #1053, opened 2023-05-31
Prefer unsigned over signed where possible

AtariDreams:


infrastation:
As elsewhere, this includes changes that have nothing to do with type signedness.

PR #1056, opened 2023-06-04
Prefer calloc over malloc

AtariDreams:
This makes the code cleaner anyway and avoids branching.


infrastation:
This change is correct.

PR #1061, opened 2023-07-10
ptp: Fix output to be more up-to-date with the standards

cappe987:
Edit: I decided to add another commit since they both fix PTP output issues.

- Fix PTP version parsing to separate major and minor version now that minor version is being used
- Print whole transportSpecific field since it no longer means only "v1 compat"


fxlb:
From: https://sagroups.ieee.org/1588/wp-content/uploads/sites/144/2020/04/IEEE-1588-2019-changes.pdf

Version number in the common header
```
A minorVersionPTP field has been created in the common PTP message header, from a previously
reserved field. Implementations compatible to the 2019 edition shall indicate a versionPTP field value of
2 and minorVersionPTP field value of 1, indicating that this is PTP version 2.1.
```
Perhaps this "official" change entry should be added as a comment.

fxlb:
About the second commit:
Wireshark remove `transport specific` stuff in ed2955cf4c3659be41cb848a6051ae036183ef58...

cappe987:
> From https://sagroups.ieee.org/1588/wp-content/uploads/sites/144/2020/04/IEEE-1588-2019-changes.pdf:
> 
> Version number in the common header
> 
> ```
> A minorVersionPTP field has been created in the common PTP message header, from a previously
> reserved field. Implementations compatible to the 2019 edition shall indicate a versionPTP field value of
> 2 and minorVersionPTP field value of 1, indicating that this is PTP version 2.1.
> ```
> 
> Perhaps this "official" change entry should be added as a comment.

Added to the commit. Nice document you found that lists all the changes.

> About the second commit: Wireshark remove `transport specific` stuff in ed2955cf4c3659be41cb848a6051ae036183ef58...

Ah, I should have checked what Wireshark uses. 802.1AS-2020 renames the transportSpecific field as majorSdoId, hence the change in that Wireshark commit. So I did the same.

fxlb:
Thanks for the update. Do you have a pcap(ng) file with a (or some)  PTP version 2.1 packet(s)?

cappe987:
> Thanks for the update. Do you have a pcap(ng) file with a (or some) PTP version 2.1 packet(s)?

Added a test with PTP v2.1 packets

fxlb:
Thank you!

PR #1072, opened 2023-08-12
QUIC: Fix a pointer overflow with 32-bit executable

fxlb:
When decoding an IPv6 Jumbo frame, the lenght of the payload may be huge (e.g. 201311154). This gives a huge length to udp_print() and then to quic_print() (e.g. 201311130). With 32-bit executable, addding the length to the pointer may overflow, like:

print-quic.c:277:26: runtime error: pointer index expression with base 0xf42032c0 overflowed to 0x001ff65a

Use ND_BYTES_AVAILABLE_AFTER() to assign the 'end' pointer.

quic_print(): Remove the parameter 'len' no longer used.

Add a test file.

Update the output of a test accordingly (packet 10 not truncated).


fxlb:
Waiting for CI...

fxlb:
Test file added.

fxlb:
Merged.

PR #1076, opened 2023-08-16
LDP: outstanding work to display additional TLVs and Message Types

hannesgredler:
This is to complete support for all the yet unsupported messages/TLVs as per the ldp-common-session.pcap testfile.
testfile output has been updated.

### DIFF
--- a/print-ldp.c
+++ b/print-ldp.c
@@ -554,8 +554,7 @@ ldp_tlv_print(netdissect_options *ndo,
 	    ND_PRINT(", causing Message ID: 0x%08x", ui);
 	ui = GET_BE_U_2(tptr);
 	if (ui)
-	    ND_PRINT(", Message ID: %s", tok2str(ldp_msg_values, "Unknown", ui));
-
+	    ND_PRINT(", Message Type: %s", tok2str(ldp_msg_values, "Unknown", ui));
 	break;
 
     case LDP_TLV_FT_SESSION:
--- a/print-ldp.c
+++ b/print-ldp.c
@@ -142,6 +142,7 @@ static const struct tok ldp_msg_values[] = {
 #define	LDP_TLV_ATM_SESSION_PARM     0x0501
 #define	LDP_TLV_FR_SESSION_PARM      0x0502
 #define LDP_TLV_FT_SESSION	     0x0503
+#define LDP_TLV_TYPED_WC_FEC_CAP     0x050b /* rfc 5918 */
 #define	LDP_TLV_LABEL_REQUEST_MSG_ID 0x0600
 #define LDP_TLV_MTU                  0x0601 /* rfc 3988 */
 #define LDP_TLV_DUAL_STACK_CAP       0x0701 /* rfc 7552 */
@@ -166,6 +167,7 @@ static const struct tok ldp_tlv_values[] = {
     { LDP_TLV_ATM_SESSION_PARM,      "ATM Session Parameters" },
     { LDP_TLV_FR_SESSION_PARM,       "Frame-Relay Session Parameters" },
     { LDP_TLV_FT_SESSION,            "Fault-Tolerant Session Parameters" },
+    { LDP_TLV_TYPED_WC_FEC_CAP,      "Typed Wildcard FEC Capability" },
     { LDP_TLV_LABEL_REQUEST_MSG_ID,  "Label Request Message ID" },
     { LDP_TLV_MTU,                   "MTU" },
     { LDP_TLV_DUAL_STACK_CAP,        "Dual-Stack Capability" },
@@ -557,6 +559,11 @@ ldp_tlv_print(netdissect_options *ndo,
 	ui = GET_BE_U_4(tptr);
 	if (ui)
 	    ND_PRINT(", Recovery Time: %ums", ui);
+	break;
+
+    case LDP_TLV_TYPED_WC_FEC_CAP:
+	TLV_TCHECK(1);
+	ND_PRINT("\n\t      %s", GET_U_1(tptr)&0x80 ? "Support" : "No Support");
 	break;
 
     case LDP_TLV_MTU:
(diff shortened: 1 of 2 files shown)


print-ldp.c
```diff
@@ -517,6 +529,15 @@ ldp_tlv_print(netdissect_options *ndo,
 	ND_PRINT("\n\t      MTU: %u", GET_BE_U_2(tptr));
 	break;
 
+    case LDP_TLV_DUAL_STACK_CAP:
+        TLV_TCHECK(4);
+	transport_pref = GET_U_1(tptr);
```
infrastation, line 593: Indentation is off.

print-ldp.c
```diff
@@ -491,17 +522,21 @@ ldp_tlv_print(netdissect_options *ndo,
...
+		 tok2str(ldp_status_code_values, "Unknown", ui&0x3fffffff),
+		 ui&0x80000000 ? "Fatal error" : "Advisory Notification",
+		 ui&0x40000000 ? "do" : "don't");
 	ui = GET_BE_U_4(tptr);
 	tptr+=4;
 	if (ui)
 	    ND_PRINT(", causing Message ID: 0x%08x", ui);
+	ui = GET_BE_U_2(tptr);
+	if (ui)
+	    ND_PRINT(", Message ID: %s", tok2str(ldp_msg_values, "Unknown", ui));
```
infrastation, line 557: Section 3.4.6 of RFC 5036 calls this 16-bit field "Message Type".

print-ldp.c
```diff
@@ -372,6 +372,23 @@ ldp_tlv_print(netdissect_options *ndo,
         }
 	break;
 
+    case LDP_TLV_HOP_COUNT:
+        TLV_TCHECK(1);
+	ND_PRINT("\n\t      Hop Count: %u", GET_U_1(tptr));
```
infrastation, line 377: Indentation is off in the new code.

print-ldp.c
hannesgredler: Is it ? I am still using emacs default value plus 'c-basic-offset 4' since I have authored the file 21 years ago.
Is there some indent config file or editor settings that you do recommend ?

print-ldp.c
hannesgredler: my guess is that ui just means unsigned integer and is used as a general purpose variable for extracting.
ui is being used for extracting status code and message_id previously - shall i change that as well ?

print-ldp.c
hannesgredler: There seems to be a wild mix of tabs and whitespaces in the current code. - Just check sed -i 's/^        /\t/' print-ldp.c; git diff
What are the new house rules ? Tab or whitespace or just whitespace for indentation ?

print-ldp.c
infrastation: The variable name is not a problem. The label is wrong:
```
   |U|F| Status (0x0300)           |      Length                   |
   |                     Status Code                               |
   |                     Message ID                                |
   |      Message Type             |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
What the new code prints is "Message Type", and "Message ID" was printed just before.

print-ldp.c
infrastation: It looks off in GitGub diffs, and loading the file into an editor that can display whitespace (I use Mousepad for that) confirms it:
Screenshot_2023-08-18_08-26-32
The current approach is "try to keep code style consistent per file", and this file seems to assume a tab character for every 8 spaces.

print-ldp.c
fxlb: > Tab or whitespace or just whitespace for indentation ?

Currently there are:
```
$ git grep '^   ' '*.c' |wc -l
46703        (TABS)
$ git grep '^        ' '*.c' |wc -l
14657        (8 SPACES)
$ git grep '^    [^ ]' '*.c' |wc -l
7738         (4 SPACES)
```
For new sources, "tab" should be used.

The rule in CONTRIBUTING.md is "Use the style of the modified sources."
```
For print-ldp.c:
$ git grep '^   ' print-ldp.c |wc -l
118        (TABS)
$ git grep '^        ' print-ldp.c |wc -l
200        (8 SPACES)
$ git grep '^    [^ ]' print-ldp.c |wc -l
136        (4 SPACES)
```
For this one there is no obvious style. We have to re-format the indentation and therefore choose between tabs, 8 spaces and 4 spaces.

infrastation: Yes, it is quite inconsistent as it is now. If possible, try to keep it consistent within the same case block at least.

fxlb: This should be done before this PR as a separate commit.

tabs ?

infrastation: I could redo it nicely using tabs that are one level of indentation rather than n spaces, but since this problem isn't new, I would also be happy to do it much later, and for now just not make it worse than it is.

fxlb: Or perhaps '4 spaces' is less change. i.e:
1) s/^\t\t/<16 spaces>/
2) s/^\t/<8 spaces>/

print-ldp.c
hannesgredler: got it - nice catch;

print-ldp.c
hannesgredler: IMO keeping the git blame history is much more important than a uniform style. - It would help if we leave editor indentation configs at the end of the file.

```/*
 * Local Variables:
 * c-basic-offset: 4
 * indent-tabs-mode: t
 * End:
 */```

hannesgredler: I have fixed the patch and replaced leading blocks of 8 whitespaces with a tab.

hannesgredler:
Have incorporated review comment. Any more you want to get changed ?

fxlb:
Fix 2 indentations, 6 to 5 commits, rebased, merged. Thanks!

PR #1085, opened 2023-09-21
IP: Enable TSO (TCP Segmentation Offload) support

fxlb:
It's a follow-up to c8623960f050cb81c12b31107070021f27f14b18.

The support is now enabled by default. No more "#ifdef GUESS_TSO".

Output example:
  length 2016 [was 0, presumed TSO]

Add a test file with presumed TSO.
(From http://cloudshark.org/captures/25e40f73bc1c, found via https://osqa-ask.wireshark.org/questions/16279/
why-are-the-bytes-00-00-but-wireshark-shows-an-ip-total-length-of-2016/)

Use ND_ICHECKMSG_U() to test an invalid total length < header length.

Output example:
  [total length 19 < 20] (invalid)

Add a test file for this case.

Move ND_TCHECK_SIZE(ip) after the new tests.

Add a const qualifier for the ip_print() parameter 'length'.

Update the man page.


fxlb:
First: CI checks.

PR #1086, opened 2023-09-30
RADIUS: Modernize packet parsing

fxlb:
Add a sanity check: The maximum length is 4096.
Print the protocol name with nd_print_protocol_caps() before any check. Enable ND_LONGJMP_FROM_TCHECK and remove the 'trunc' labels. Report invalid packets as invalid. Use ND_ICHECK*() for length check. Remove some redundant ND_TCHECK_*().
Fix a length check for ARAP-Challenge-Response attribute (s/</!=/). Add a missing comma between 'id' and 'length' in non-verbose output. print_vendor_attr(): Factorize some code; print length including type and length as in most cases.
Use GET_CPY_BYTES() instead of memcpy().
Add a const qualifier for the radius_print() parameter 'length'.

The test file radius_rfc5447_invalid_length.pcap gives now an error: [length 57 > 56] (invalid); follow-up to f64a4a5f49. Update RADIUS length (from 44006 to 263) in radius_attr_asan.pcap to pass the two new tests (length <= 4096 and length <= original length). Update some test outputs accordingly.

Moreover:
Remove some extra blank lines.


fxlb:
Testing with CI...

PR #1094, opened 2023-10-28
Remove local bpf_dump(), supplied by libpcap since version 0.6

fxlb:


infrastation:
Apparently, OpenBSD libpcap does not implement `bpf_dump()`.

fxlb:
It seems that OpenBSD libpcap don't supply bpf_dump()...

fxlb:
Keep running the tests to see if it's the only OS, before closing.

infrastation:
Well, maybe then the best thing to do would be to add a comment somewhere to say "the only known reason to keep this check/code is that OpenBSD libpcap does not implement bpf_dump()" or something like that.

fxlb:
Yes, done.

PR #1096, opened 2023-11-04
Ospf: Update LS-Ack printing to not run off the end of the packet

fenner:
The LS-Ack loop printed `lshdr`s until it got truncation.  Before truncation was changed to longjmp, this worked - the truncation return value meant we had gone past the end of the data.  However, with the change to longjmp handling, this results in LS-Ack always printing `[|ospf]` as if the packet was truncated.

This change updates the "while" loop condition to check the `dataend` pointer, avoiding the problem of looking past the end of the packet, and adds a test packet.


fxlb:
Thank you!

PR #1097, opened 2023-11-05
NFS: Fix a MemorySanitizer error

fxlb:
In parsenfsfh.c, Parse_fh(), switch (fhtype), case FHT_SUNOS5, the Fsid_dev.Minor can be 257.

Thus using 257 as a flag value ("bogus") in case FHT_UNKNOWN when Opaque_Handle[] is initialized is incorrect. This value is tested in nfs_printfh() to print or not Opaque_Handle[]. This can result in a case of use-of-uninitialized-value.

To avoid this, use UINT_MAX as flag values for Fsid_dev.Minor and Fsid_dev.Major and test the two variables in nfs_printfh().

The error was:
==9391==WARNING: MemorySanitizer: use-of-uninitialized-value [...]
MemorySanitizer: use-of-uninitialized-value util-print.c:91:2 in fn_print_str

Add two XXX comments with questions.

follow-up to commit b5353b62c2c58d3ac7c8c84b2df041f5c1602b6b.

### DIFF
--- a/parsenfsfh.c
+++ b/parsenfsfh.c
@@ -48,6 +48,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <limits.h>
 
 #include "netdissect-ctype.h"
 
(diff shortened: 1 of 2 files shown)


fxlb:
CI checks...

PR #1099, opened 2023-11-09
Use symmetrical quotation characters in error messages

fxlb:
Symmetrical quoting looks better.

Thus s/`/'/

See '5.10 Quote Characters' in GNU Coding Standards:

https://www.gnu.org/prep/standards/standards.html#Quote-Characters

[skip ci]

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -80,6 +80,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
     User interface:
       Print the supported time stamp types (-J) to stdout instead of stderr.
       Print the list of data link types (-L) to stdout instead of stderr.
+      Use symmetrical quotation characters in error messages.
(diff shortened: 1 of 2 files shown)


infrastation:
This software certainly is not in the GNU orbit, so the coding standards apply only as much as the case make sense in the code base. That said, this change may go better with a change log entry because somebody somewhere may be using a very specific regexp matching exactly one or another error message.

fxlb:
> This software certainly is not in the GNU orbit, 

Yes, but we can use some good ideas from these coding standards.

> so the coding standards apply only as much as the case makes sense in the code base.

The errors messages looks better with this change.

> That said, this change may go better with a change log entry because somebody somewhere may be using a very specific regexp matching exactly one or another error message.

Done

PR #1088, opened 2023-10-05
OSPF: add support for segment routing extensions as per rfc8665 and r…

hannesgredler:
This is the rebased version of the previous ospf-sr merge request.

PR #1101, opened 2023-11-10
OSPF: add additional support for segment routing extensions

fenner:
This pull request builds upon @hannesgredler's #1088 to display additional TLVs and their sub-TLVs from RFC8665:
- SR-Algorithm TLV
- SR Local Block TLV and its sub-TLVs
- SRMS Preference TLV

It also uses the SR-Algorithm names when printing the Prefix-SID Sub-TLV (added in #1088).

This pull request is a draft; it is currently based on Hannes' changes and will be rebased on master when that PR is accepted.  The git commit messages will be updated, and changes will be squashed as appropriate.


fenner:
Sorry, I thought github would be able to recognize that the pull request was based on top of another one and only show the diffs.

My changes can be seen at https://github.com/fenner/tcpdump/compare/hannes-ospf-sr...fenner:tcpdump:arista-sr-additions?expand=1 or we can just ignore this PR until the predecessor merges.

print-ospf.c
```diff
@@ -803,15 +803,16 @@ ospf_print_ep_lsa_extd_prefix_tlv(netdissect_options *ndo, const uint8_t *tptr,
...
-			 GET_BE_U_4(tptr+4), mt_id, algo,
-			 bittok2str(ep_range_tlv_prefix_sid_subtlv_flag_values, "none", flags));
+		ND_PRINT("\n\t\t  Index: %u", GET_BE_U_4(tptr+4));
 	    } else {
 		ND_PRINT("\n\t\tBogus subTLV length %u", subtlv_length);
+		break;
 	    }
+	    ND_PRINT( ", MT-ID: %u, Algorithm: %s (%u), Flags [%s]",
+		      mt_id, tok2str(lsa_opaque_ri_tlv_sr_algos, "Unknown", algo), algo,
+		 bittok2str(ep_range_tlv_prefix_sid_subtlv_flag_values, "none", flags));
```
fenner, line 815: @hannesgredler just wanted to check that you're ok with this change to your prefix printing - using the list of sr algorithms to print `Shortest-Path First (0)` instead of `0`.

hannesgredler: Yes, that's fine.

hannesgredler:
fine with me.

On Mon, Nov 13, 2023 at 05:14:23AM -0800, Bill Fenner wrote:
|    @fenner commented on this pull request.
| 
|      ----------------------------------------------------------------------
| 
|    In print-ospf.c:
| 
|  >           if (subtlv_length == 7) {
|  -               ND_PRINT("\n\t\t  Label: %u, MT-ID: %u, Algorithm: %u",
|  -                        GET_BE_U_3(tptr+4), mt_id, algo);
|  +               ND_PRINT("\n\t\t  Label: %u", GET_BE_U_3(tptr+4));
|              } else if (subtlv_length == 8) {
|  -               ND_PRINT("\n\t\t  Index: %u, MT-ID: %u, Algorithm: %u, Flags [%s]",
|  -                        GET_BE_U_4(tptr+4), mt_id, algo,
|  -                        bittok2str(ep_range_tlv_prefix_sid_subtlv_flag_values, "none", flags));
|  +               ND_PRINT("\n\t\t  Index: %u", GET_BE_U_4(tptr+4));
|              } else {
|                  ND_PRINT("\n\t\tBogus subTLV length %u", subtlv_length);
|  +               break;
|              }
|  +           ND_PRINT( ", MT-ID: %u, Algorithm: %s (%u), Flags [%s]",
|  +                     mt_id, tok2str(lsa_opaque_ri_tlv_sr_algos, "Unknown", algo), algo,
|  +                bittok2str(ep_range_tlv_prefix_sid_subtlv_flag_values, "none", flags));
| 
|    @hannesgredler just wanted to check that you're ok with this change to
|    your prefix printing - using the list of sr algorithms to print
|    Shortest-Path First (0) instead of 0.
| 
|    —
|    Reply to this email directly, view it on GitHub, or unsubscribe.
|    You are receiving this because you were mentioned.Message ID:
|    ***@***.***>

fenner:
> Does it replace the PR #1049?

Yes, it builds upon 6883ca65bed0b591fb611990ee8353182055b89d to include everything that was in the earlier PR.  I've closed the earlier one, thanks for pointing out that oversight.

tests/ospf-sr-ri-sid-v.out
hannesgredler, line 1: Could you trim down the output and the pcap file to only the frame that shows the relevant TLVs using a binary editor (.e.g bvi). There is a ton of hellos and other stuff here that will create false positive once somebody changes e.g. an unrelated part of the parser (OSPF Hello).

print-ospf.c
```diff
@@ -1099,6 +1164,15 @@ ospf_print_lsa(netdissect_options *ndo,
                         nd_printjnp(ndo, tptr, tlv_length);
                         break;
 
+                    case LS_OPAQUE_RI_TLV_SR_ALGO:
+                        num_tlv = tlv_length;
+                        while (num_tlv != 0) {
```
hannesgredler, line 1169: I'd rather compare here for a min-length of the algo (> 1)

tests/ospf-sr-ri-tlv-v.out
hannesgredler, line 1: Can you also add .pcap examples of the SRGB local block and the SRMS Preference TLV. Looks like those are missing from the UT.

tests/ospf-sr-ri-sid-v.out
fxlb: +1.

If the need is to extract some packets without modifying them, this should work:
(`pcappick` is in the Debian package `pcaputils`)
```shell
pcap_in=tests/ospf-sr-ri-sid.pcap
pcap_out=new.pcap

packets='3 5 6 10'

: > file_list.txt
for p in $packets; do
        pcappick -c $p -r $pcap_in -w /tmp/pkt-$p.pcap
        echo "/tmp/pkt-$p.pcap" >> file_list.txt
done
./tcpdump -V file_list.txt -w $pcap_out
```

fxlb: The packet list is just an example...

print-ospf.c
```diff
@@ -708,6 +719,59 @@ ospf_print_ri_lsa_sid_label_range_tlv(netdissect_options *ndo, const uint8_t *tp
     return 0;
 }
 
+static int
+ospf_print_ri_lsa_sid_local_block_tlv(netdissect_options *ndo, const uint8_t *tptr,
```
fenner, line 723: It turns out, when stepping back one step, that the [SID/Label Range TLV](https://datatracker.ietf.org/doc/html/rfc8665#section-3.2) and the [SR Local Block TLV](https://datatracker.ietf.org/doc/html/rfc8665#section-3.3) have exactly the same definition:
```
   |              Type             |             Length            |
   |                    Range Size                 |   Reserved    |
   |                        Sub-TLVs (variable)                    |
   +-                                                             -+
   |                                                               |
   +                                                               +
```
and each section says *separately* that only the single SID/Label Sub-TLV in section 2.1 is allowed.

So right now, the implementation of `ospf_print_ri_lsa_sid_local_block_tlv()` is identical to that of `ospf_print_ri_lsa_sid_label_range_tlv()`.

Here's the question: do we unify them to one "print subtlvs of local-block or label-range" function, or do we keep them two identical functions (to start with) with the assumption that it's possible that a sub-TLV will be defined only for local-block, or only for label-range?

If we assume they're identical, then the whole local-block printer could just be an additional "case:" label that uses the label-range printer.  But there's nothing in the spec that requires that they be identical.

Any thoughts? @hannesgredler

hannesgredler: You are right - makes certainly sense to unify. As you have noted encoding is same on both formats.

ospf.h
```diff
@@ -110,9 +111,12 @@
 #define LS_OPAQUE_GRACE_TLV_REASON_SW_UPGRADE  2 /* rfc3623 */
 #define LS_OPAQUE_GRACE_TLV_REASON_CP_SWITCH   3 /* rfc3623 */
 
-#define LS_OPAQUE_RI_TLV_CAP             1 /* draft-ietf-ospf-cap-03 */
-#define LS_OPAQUE_RI_TLV_HOSTNAME        7 /* rfc5642 */
-#define LS_OPAQUE_RI_TLV_SID_LABEL_RANGE 9 /* rfc8665 */
+#define LS_OPAQUE_RI_TLV_CAP               1 /* draft-ietf-ospf-cap-03 */
```
fxlb, line 114: rfc7770 ?

hannesgredler: Yes -  draft-ietf-ospf-cap-03 became rfc4970 which later got obsoleted by rfc7770
Here is the history -> https://datatracker.ietf.org/doc/search?name=+draft-ietf-ospf-cap-03&sort=&rfcs=on&activedrafts=on&by=group&group=

fenner: Yup!

print-ospf.c
fenner: I've unified the case statements in the latest set of changes, thanks.

fxlb:
Thank you!

PR #1104, opened 2023-12-01
Have a #define to squelch -Wunused-result warnings.

guyharris:
In one place, we really *do* legitimately ignore the return value of pcap_set_datalink(). We cast its return value to void, which appears to squelch warnings from Clang, but not from at least some versions of GCC, so provide DIAG_OFF_WARN_UNUSED_RESULT/DIAG_ON_WARN_UNUSED_RESULT and use them to squelch the warning.

### DIFF
--- a/diag-control.h
+++ b/diag-control.h
@@ -132,6 +132,22 @@
   #define DIAG_ON_CAST_QUAL \
     DIAG_DO_PRAGMA(GCC diagnostic pop)
 
+  #if ND_IS_AT_LEAST_GNUC_VERSION(4,5)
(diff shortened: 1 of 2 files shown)


diag-control.h
```diff
@@ -158,7 +174,7 @@
    * So please remember to use this very carefully.
    */
   #define ND_UNREACHABLE __builtin_unreachable();
-#endif
+ #endif
```
infrastation, line 178: ?

guyharris: Fixed.

PR #1107, opened 2023-12-02
Remove some useless "ipproto.h" includes

fxlb:


fxlb:
CI test for now.

guyharris:
Almost all of the print-XXX.c files that include ipproto.h use an `IPPROTO_` value; the exception is print-isakmp.c, which uses `netdb_protoname()`, which is declared in ipproto.h.

PR #1111, opened 2023-12-08
Configure check for MD5_Init instead of DES_cbc_encrypt

justsmth:
With this configure-script patch, tcpdump successfully builds and (with [a few other small changes](https://github.com/aws/aws-lc/pull/1351)) tests pass using AWS-LC as the libcrypto.
   * Related PR for AWS-LC is here: [#1351](https://github.com/aws/aws-lc/pull/1351)
   * AWS-LC doesn't have a `DES_cbc_encrypt` function, but it supports the ciphers required for ESP in the tests.
   * `MD5_Init` appears to be supported by other libcrypto implementations (e.g., [OpenSSL](https://www.openssl.org/docs/man1.1.1/man3/MD5_Init.html) and [LibreSSL](https://man.openbsd.org/MD5_Init)).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -868,8 +868,8 @@ if test "$want_libcrypto" != "no"; then
 	#
 	AC_CHECK_HEADER(openssl/crypto.h,
 	[
-		AC_CHECK_LIB(crypto, DES_cbc_encrypt)
-		if test "$ac_cv_lib_crypto_DES_cbc_encrypt" = "yes"; then
+		AC_CHECK_LIB(crypto, AES_cbc_encrypt)
+		if test "$ac_cv_lib_crypto_AES_cbc_encrypt" = "yes"; then
 			AC_CHECK_HEADERS(openssl/evp.h)
 			#
 			# OK, then:
--- a/configure.ac
+++ b/configure.ac
@@ -868,8 +868,8 @@ if test "$want_libcrypto" != "no"; then
 	#
 	AC_CHECK_HEADER(openssl/crypto.h,
 	[
-		AC_CHECK_LIB(crypto, AES_cbc_encrypt)
-		if test "$ac_cv_lib_crypto_AES_cbc_encrypt" = "yes"; then
+		AC_CHECK_LIB(crypto, MD5_Init)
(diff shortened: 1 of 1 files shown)


infrastation:
Thank you for bringing this up. The CMake leg of the build process tests for `EVP_CIPHER_CTX_new()` and `EVP_DecryptInit_ex()`, which are what `print-esp.c` uses. Perhaps it would make more sense to test for the same in the Autoconf leg.

Have you tried whether CMake tcpdump build works correctly with AWS-LC? (The simplest way to test for that should be running `CMAKE=yes CRYPTO=yes ./build.sh` and checking for `Found CRYPTO` in the output.)

justsmth:
> Thank you for bringing this up. The CMake leg of the build process tests for `EVP_CIPHER_CTX_new()` and `EVP_DecryptInit_ex()`, which are what `print-esp.c` uses. Perhaps it would make more sense to test for the same in the Autoconf leg.

On [line 894](https://github.com/the-tcpdump-group/tcpdump/pull/1111/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R894) (right after this) it does a check for `EVP_CIPHER_CTX_new` and `EVP_DecryptInit_ex`. So I think the `DES_cbc_encrypt` (or `AES_cbc_encrypt`) check can be removed. I'll update this PR.

> Have you tried whether CMake tcpdump build works correctly with AWS-LC? (The simplest way to test for that should be running `CMAKE=yes CRYPTO=yes ./build.sh` and checking for `Found CRYPTO` in the output.)

Yeah, I was able to build/test tcpdump against AWS-LC with CMake. I first built/installed AWS-LC into a directory.  Then set `CMAKE_PREFIX_PATH` to that directory when setting up the tcpdump build:
```
cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=/home/justsmth/tcpdump-testing/aws-lc-install/ -S /home/justsmth/tcpdump-testing/tcpdump -B /home/justsmth/tcpdump-testing/tcpdump-build
```

justsmth:
Resolving this in favor of a more recent/better proposal in #1174

PR #1112, opened 2023-12-12
Cirrus CI: drop FreeBSD 12.4 and add FreeBSD 14.0

emaste:
FreeBSD 12.4 is EOL at the end of 2023, and FreeBSD 14.0 was released recently.


infrastation:
Thank you!

PR #1114, opened 2023-12-19
Add --lengths option to print capture and wire packet lengths

fxlb:
The lengths will be printed after the packet number, if any.
    
    Examples
    1) With -#n
        1  caplen 80 len 98 14:41:53.503612 IP 192.168.1.11.43966 > [...]
    2) With -n
    caplen 80 len 98 14:43:38.185603 IP 192.168.1.11.43966 > [...]


fxlb:
We also could use `--lengths` as option name: shorter and just as clear.

fxlb:
Updated to use `--lengths`.

infrastation:
It may be more user-friendly to print `declared %u captured %u` or `declared-length %u captured-length %u` or some such. Also the man page should explain the option and both fields in more detail.

fxlb:
> It may be more user-friendly to print `declared %u captured %u` or `declared-length %u captured-length %u` or some such.

I had thought of similar texts at first, but then I thought it was necessary to be consistent with sanity checks messages dating from 2018, like `[Invalid header: len(34) < caplen(35)]`.

> Also the man page should explain the option and both fields in more detail.

Yes, I should have convert the PR to draft. It's the first step. Need man and print_usage() updates.

PR #1115, opened 2024-01-05
Parse ICMP Extended Echo (PROBE) (RFC8335)

fenner:
This adds parsing of ICMP extended echo request/reply, printing ID and seq without -v, and additional information with -v.

I've made two changes here; the first one is *just* moving the existing parsing of extended icmp objects to a separate function, and the second one adds the extended echo functionality, to make reviewing easier.  I can squash them before merge if that's preferable.


print-icmp.c
```diff
@@ -59,12 +59,19 @@ struct icmp {
 			nd_uint16_t icd_id;
 			nd_uint16_t icd_seq;
 		} ih_idseq;
+		struct ih_idseqx {	/* RFC8550 */
```
fenner, line 62: This is supposed to be RFC8335, my memory misfired here.

fenner (commented):
My ongoing work conflicted with changes from master, and I didn't rebase the first commit completely properly, so unfortunately the second commit has a couple of fixups of the rebase as well as my changes.  I don't think this is too confusing, if you have a question of "why is this change in this commit" that's why.  I can separating them out if really needed.

fxlb:
To make review easier, please put the first commit ("Move multipart extension object printing to its own function") in its own pull request. This step should be straightforward.

fenner:
Fuzzing found an unchecked call to `in_cksum()`, so I've made this into a draft so that it's not merge eligible until we decide what to do.

Should `in_cksum()` itself check for enough bytes, or, should all of its callers check before calling it?

fenner:
As discussed, I've added a length check around the call to `in_cksum()`, and added a pcap that contains a sample truncated packet.

fenner:
[I've updated the test outputs to match the timestamping change]

fxlb:
[The build errors don't seem to be related to this PR. More tests needed].

fxlb:
`netbsd-mips64` CI is now OK.

fxlb:
Thanks!

fenner:
> Thanks!

Thank *you* too!

PR #1119, opened 2024-01-16
ICMP: refactor RFC4884 object printing

fenner:
For a future change, we refactor RFC4884 object printing into its own function.  This change just moves and reindents the lines in question.

### DIFF
--- a/print-icmp.c
+++ b/print-icmp.c
@@ -353,6 +353,142 @@ icmp_tstamp_print(u_int tstamp)
     return buf;
 }
 
+static int
+print_icmp_multipart_ext_object(netdissect_options *ndo, const uint8_t *obj_tptr)
+{
+	u_int obj_tlen, obj_class_num, obj_ctype;
+	const struct icmp_multipart_ext_object_header_t *icmp_multipart_ext_object_header;
+
+	icmp_multipart_ext_object_header = (const struct icmp_multipart_ext_object_header_t *)obj_tptr;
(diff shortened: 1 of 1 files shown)


fxlb:
Fix indent.

print-icmp.c
```diff
@@ -782,139 +918,12 @@ icmp_print(netdissect_options *ndo, const u_char *bp, u_int plen,
...
-                  }
-
-                default:
-                    print_unknown_data(ndo, obj_tptr, "\n\t    ", obj_tlen);
-                    break;
-                }
-                if (hlen < obj_tlen)
+                int obj_tlen = print_icmp_multipart_ext_object(ndo, obj_tptr);
+                if (obj_tlen < 0) {
+		    nd_print_trunc(ndo);
```
fxlb, line 923: Because ND_LONGJMP_FROM_TCHECK is defined in this printer, all truncated cases should longjmp(). Thus we should use nd_trunc_longjmp(ndo) here.

fxlb: But are 'return -1' truncated cases or just invalid cases?

fenner: Thanks for catching this. I will re-review tomorrow.

fxlb: > It's possible that there's an improvement that would continue the loop in the case of an unknown address family, since we know the length of the overall object, but, that can be a future change.

OK.

fxlb:
Fix some indentations and spaces, merged. Thanks.

PR #1126, opened 2024-02-01
Add printing for HSR/PRP protocols

cappe987:
HSR and PRP are redundancy protocols that duplicates frames and then eliminates the duplicates to ensure only one copy arrives at the destination. The protocols are interoperable and can be converted between. They are defined in the standard IEC62439-3.

Though, PRP is a bit odd so I'm not sure it can even be merged. It uses a trailer at the end of the packet to store the information. So parsing the packet means when need to search from the back. The trailer also holds the packet size to help verify that it is an actual PRP packet, but there's no guarantee. It could match on data in a non-PRP packet that just happens to coincide with the protocol specification. I didn't see any good way to handle this, but if you have ideas I could give it a try. If not possible I can remove the commit.

HSR tags work similar to VLAN tags, and are always placed after any VLAN tags.

### DIFF
--- a/netdissect.h
+++ b/netdissect.h
@@ -223,6 +223,7 @@ struct netdissect_options {
   int ndo_Aflag;		/* print packet only in ASCII observing TAB,
 				 * LF, CR and SPACE as graphical chars
 				 */
+  int ndo_Pflag;		/* print PRP trailer */
   int ndo_Hflag;		/* dissect 802.11s draft mesh standard */
   const char *ndo_protocol;	/* protocol */
   jmp_buf ndo_early_end;	/* jmp_buf for setjmp()/longjmp() */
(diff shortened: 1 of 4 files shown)


infrastation:
Does `ether proto 0x892f` as the filter match the packets you expect?

cappe987:
> Does `ether proto 0x892f` as the filter match the packets you expect?

Yes that matches all HSR tagged packets. I meant if I want to filter on HSR tagged ARP packets I can't do that. I just realized you can't do `ether proto 0x8100` for VLAN, I assume the filtering is transparent to VLAN tags (unless specified with `vlan X`). So should HSR tags behave the same perhaps?

cappe987:
Yes. Take your time. I haven't started any libpcap work.

PR #1129, opened 2024-02-07
TESTrun: Fix typos in the Windows code path

fxlb:
The typos were "\\" instead of "\\\\" (second 'fc' command).

Moreover:
Use more File::Spec->canonpath() with 'newdir' and 'diffdir' variables.

### DIFF
--- a/tests/TESTrun
+++ b/tests/TESTrun
@@ -225,7 +225,9 @@ sub runtest {
         #
         if ($^O eq 'MSWin32') {
             my $winoutput = File::Spec->canonpath($output);
-            $r = system "fc /lb1000 /t /1 $winoutput tests\\NEW\\$outputbase >tests\\DIFF\\$outputbase.diff";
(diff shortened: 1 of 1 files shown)


infrastation:
FWIW, this works on Solaris 10, which has Perl 5.8.4 from 2004.

guyharris:
> FWIW, this works on Solaris 10, which has Perl 5.8.4 from 2004.

Presumably meaning "the Perl parser doesn't have a problem with those changes" - that code should never be used at run time on a UN\*X.

Given that, and given that all checks have passed, is there any reason not to merge this?

fxlb:
> Given that, and given that all checks have passed, is there any reason not to merge this?

I don't see any, I will merge it.

infrastation:
I meant, this change in the Perl code will not require a fix in the non-Windows space such as the one in commit 85a2789, which was my only concern.

PR #1130, opened 2024-02-07
[DRAFT] Just try doing checks on Windows to view any error messages...

fxlb:
[skip ci]


fxlb:
The tcpdump.exe don't run:
```
Release\tcpdump.exe --version
Command exited with code -1073741515
```
(Same when using "Debug".)

infrastation:
-1073741515 == 0xffffffffc0000135, both values yield search results discussing location of Windows libraries and setting of `PATH`.

fxlb:
Thanks, and missing DLLs also.

fxlb:
Investigation with `Dependencies` (https://github.com/lucasg/Dependencies)
SDK=WpdPack --> wpcap.dll (NOT_FOUND). But no unattended installation possible with WinPcap_4_1_3.exe.
SDK=npcap-sdk-1.12 --> wpcap.dll (NOT_FOUND). It seems that npcap-1.79.exe in silent mode needs an Npcap OEM version.

guyharris:
> -1073741515 == 0xffffffffc0000135, both values yield search results discussing location of Windows libraries and setting of `PATH`.

And as a 32-bit value, it's 0xc0000135, which is, according to [the list of NT status values](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55), `STATUS_DLL_NOT_FOUND`.

fxlb:
> so that may be OK for tcpdump tests.

Any way to build without the capture function ? (Because currently `tcpdump.exe` need `wpcap.dll`)

guyharris:
> Any way to build without the capture function ?

"Build" as in "build libpcap" or "build" as in "build tcpdump"?

Libpcap *can* be built without capture support - if using CMake, passing `-DPCAP_TYPE=none` to CMake should be sufficient.

Then you'd want to build tcpdump to link with the libpcap that you built.

> (Because currently tcpdump.exe needs wpcap.dll)

Tcpdump's going to need some libpcap library, even if you're only building it to run the tests.

fxlb:
Next test: extract `wpcap.dll` from installer, copy it in \Windows\system32 and see...

infrastation:
**If** (please check that first) the licence does not forbid it, you could extract the required files once on a Windows PC, put them [here](https://ci.tcpdump.org/public/) and let the CI script download them from there into the correct place of the CI filesystem instead of the installers.

fxlb:
> extract `wpcap.dll` from installer, copy it in \Windows\system32 and see...

Copying is not enough. Installation required.

fxlb:
> Libpcap _can_ be built without capture support - if using CMake, passing `-DPCAP_TYPE=none` to CMake should be sufficient.

`-DPCAP_TYPE=null` ?

> Tcpdump's going to need some libpcap library, even if you're only building it to run the tests.

Yes, sure.

PR #1131, opened 2024-02-09
ICMPv6: print RFC8335 PROBE extended echo/reply messages

fenner:
Similar to #1115 for ICMP, this adds handling of ICMPv6 Extended Echo.

This spec (RFC8335) appears to be the first time where ICMP and ICMPv6 can reuse a non-trivial amount code.  Therefore, this change refactors the ICMP handling into a function, and calls that function from the ICMPv6 handling.

This pattern can be reused in the future, if someone implements RFC4664/RFC5387 handling for ICMPv6, since these extensions are defined identically for ICMP and ICMPv6.

I'm willing to rebase this and the ICMP pull request to be a single pull request; I am keeping this separate for now pending feedback on this style of code reuse; if there's a long discussion to be had on that topic, then the IPv4 code can at least be merged by itself.


fenner:
Is there something different about the `linux-amd64` CI system?  It can not find "icmp.h" which is part of the same change that added the #include of icmp.h.  I'd like to try to fix it if I've done something wrong with my patch, but since it works on all the other systems I wonder what could be wrong.

fxlb:
> Is there something different about the `linux-amd64` CI system?

Yes,  it run `make releasecheck` to build from a release archive to see if the archive contains all the needed files.
This command shows that `icmp.h` is missing in `EXTRA_DIST` in `Makefile.in`.

fxlb:
No it is missing in `HDR`.

fenner:
[I've updated the test outputs to match the timestamping change]

fenner:
I've updated the v6 version to resolve conflicts with the v4 version after the merge.

fxlb:
Thank you!

fenner:
That's great, thank you very much!

PR #1133, opened 2024-02-16
pflog: Modernize packet parsing

fxlb:
Enable ND_LONGJMP_FROM_TCHECK and remove a 'trunc' label.
Add MAX_PFLOG_HDRLEN and use it for a test.
Use ND_ICHECK_U() in length tests and add an 'invalid' label.

Don't check the tructation with ND_TCHECK_SIZE(hdr), because the sizeof(hdr) depend on the OS. Use ND_TCHECK_LEN(hdr, hdrlen).
Increment ndo_ll_hdr_len only in non-truncation case.

### DIFF
--- a/print-pflog.c
+++ b/print-pflog.c
@@ -27,6 +27,7 @@
 
 #include "netdissect-stdinc.h"
 
+#define ND_LONGJMP_FROM_TCHECK
 #include "netdissect.h"
 #include "extract.h"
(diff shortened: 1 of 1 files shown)


infrastation:
"tructation"

fxlb:
> "tructation"

Fixed, thanks.

PR #1138, opened 2024-02-27
Make illumos build warning-free.

infrastation:
To the best of my understanding, the approach is correct. If you find that `-Wstrict-prototypes` should use a macro of its own instead of piggy-backing on `DIAG_OFF_DEPRECATION`, please make your point before long.


guyharris:
If the idea is to have a `DIAG_{OFF,ON}_` pair to cover all "the system header files do annoying things", it should probably be given a name of that sort.

Otherwise, I'd be inclined to have a `DIAG_{OFF,ON}_STRICT_PROTOTYPES` pair.

infrastation:
Updated as advised.

infrastation:
This time it should be correct.

guyharris:
Looks good.

Issue #1145, opened 2024-03-12
[Bug] Infinite Loop Vulnerability in tcpdump leading to Potential DDoS Conditions


msxfXF:
## Description
A flaw was discovered in the tcpdump utility version 5.0.0-PRE-GIT that causes the application to enter an infinite loop when processing certain packets. This behavior can be triggered by an attacker who can deliver specially crafted packets to be processed by tcpdump. The vulnerability resides in the way tcpdump reads packet data from a file or network stream.

When running tcpdump with the options -x -vvv -n -r followed by a Proof of Concept (poc) file, the application enters an infinite loop, resulting in tcpdump consuming excessive system resources, potentially leading to a Denial of Service (DoS) condition on the system where tcpdump is running.

## Info
Versions Affected:
tcpdump version 5.0.0-PRE-GIT
libpcap version 1.11.0-PRE-GIT (with TPACKET_V3)
OpenSSL 1.1.1f 31 Mar 2020
Compiled with AddressSanitizer/Clang.

## Build
build with build.sh
```
CC=afl-gcc CFLAGS="-fsanitize=address -g" ./build.sh
```

## Reproduce
```
tcpdump -x -vvv -n -r [poc file]
```

## POC
[tcpdump_poc.zip](https://github.com/the-tcpdump-group/tcpdump/files/14565979/tcpdump_poc.zip)

msxfXF:
![image](https://github.com/the-tcpdump-group/tcpdump/assets/38368353/1c57f97b-dd71-4f88-92b1-c576e1e51e47)

guyharris:
Stack trace after killing tcpdump with SIGABRT:

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	    0x7ff80c3da036 sigprocmask + 10
1   libsystem_platform.dylib      	    0x7ff80c441e83 longjmp + 23
2   tcpdump                       	       0x100b750dd hex_print_with_offset + 365
3   tcpdump                       	       0x100b68468 pretty_print_packet + 760
4   tcpdump                       	       0x100b64141 print_packet + 49 (tcpdump.c:3252)
5   tcpdump                       	       0x100c3794e pcapint_offline_read + 110 (savefile.c:684)
6   tcpdump                       	       0x100c178a9 pcap_loop + 57 (pcap.c:2967)
7   tcpdump                       	       0x100b624df main + 6143 (tcpdump.c:2685)
[...]

PR #1146, opened 2024-03-12
ppp: use the buffer stack for the de-escaping buffer.

guyharris:
This both saves the buffer for freeing later and saves the packet pointer and snapend to be restored when packet processing is complete, even if an exception is thrown with longjmp.

This means that the hex/ASCII printing in pretty_print_packet() processes the packet data as captured or read from the savefile, rather than as modified by the PPP printer, so that the bounds checking is correct.

That fixes issue #1145, which was caused by an exception being thrown by the hex/ASCII printer (which should only happen if those routines are called by a packet printer, not if they're called for the -X/-x/-A flag), which jumps back to the setjmp() that surrounds the packet printer.  Hilarity^Winfinite looping ensues.

Also, restore ndo->ndo_packetp before calling the hex/ASCII printing routine, in case nd_pop_all_packet_info() didn't restore it.

### DIFF
--- a/print-ppp.c
+++ b/print-ppp.c
@@ -37,6 +37,8 @@
 
 #include "netdissect-stdinc.h"
 
+#include <stdlib.h>
+
 #include "netdissect.h"
 #include "extract.h"
 #include "addrtoname.h"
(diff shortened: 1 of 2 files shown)


firexinghe:
Does CVE-2024-2397 affect the 4.9.x branch?

PR #1148, opened 2024-03-13
Update --version option to print 32/64-bit build and time_t size

fxlb:
Use only sizeof() and not SIZEOF_VOID_P/SIZEOF_TIME_T because these
macros could have invalid values in some macOS fat builds (See commit
170cbc8c6d484387ec7ae19ef9f4bd14dd0845e7).

This method should be future proof.

With this change we know:
whether this is a 32-bit or 64-bit build ;
whether the time_t size is 32-bit or 64-bit.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -3341,6 +3341,8 @@ print_version(FILE *f)
 	(void)fprintf (f, "Compiled with MemorySanitizer/Clang.\n");
 #  endif
 #endif /* __SANITIZE_ADDRESS__ or __has_feature */
+	(void)fprintf (f, "%zu-bit build, %zu-bit time_t\n",
+		       sizeof(void *) * 8, sizeof(time_t) * 8);
 }
 DIAG_ON_DEPRECATION
 


infrastation:
The output would not be well-formatted if only `SIZEOF_VOID_P` is not defined or if both macros are not defined. One possible way to address this would be to print each size on a separate line, another would be to arrange equivalent checks in CMake and to print the line unconditionally.

fxlb:
Yes. The PR is draft because the CMake part is not ready...

fxlb:
The #ifdef are here only to secure. The goal is to have the two macros defined all the time.

guyharris:
At least with CMake, the `SIZEOF` values could be 0, if somebody's doing a fat build on macOS and that includes both 32-bit and 64-bit instruction sets.  What should probably be done if one of them is zero is to do something such as

```
#if SIZEOF_VOID_P == 0
#undef SIZEOF_VOID_P
#if defined(__i386__) || defined(__ppc__)
  #define SIZEOF_VOID_P 4
#else
  #define SIZEOF_VOID_P 8
#endif
#if SIZEOF_TIME_T == 0
#undef SIZEOF_TIME_T
#if defined(__i386__) || defined(__ppc__)
  #define SIZEOF_TIME_T 4
#else
  #define SIZEOF_TIME_T 8
#endif
```

for the benefit of people doing fat builds on old versions of macOS that supported 32-bit processors (yes, somebody filed a bug fairly recently to fix a bug when building for Mac OS X Tiger(!)).

Apple relies on switching to 64-bit-only code to deal with 2038; 32-bit apps are probably not going to work past the overflow date.

fxlb:
To avoid all this complication (and potential errors), it is perhaps simpler to forget the macros in `tcpdump.c` and just use `sizeof(void *)` and `sizeof(time_t)`.

fxlb:
Or, better, to define in an header file:
#define SIZEOF_VOID_P sizeof(void *)
and
#define SIZEOF_TIME_T sizeof(time_t)

infrastation:
It is better when the solution space matches the problem space.

infrastation:
Would it be useful to print the size of `off_t` as well?

guyharris:
> Would it be useful to print the size of `off_t` as well?

That affects libpcap, so maybe it belongs in the `pcap_lib_version()` string.  For tcpdump, the only question is whether there's a `pcap_dump_ftell64()` or not; if not, the "file size written so far" is a `long`, which might be 32-bit (32-bit UN\*Xes, Windows) or 64-bit (64-bit UN\*Xes), otherwise it's `int64_t`, which will always be 64-bit.

fxlb:
> if somebody's doing a fat build on macOS

Do you have a link to better undertand such a build?

fxlb:
> somebody filed a bug

Where can we find it?

guyharris:
> > somebody filed a bug
> 
> Where can we find it?

[Libpcap pull request #1203, "Allow build on OS X pre 10.5"](https://github.com/the-tcpdump-group/libpcap/pull/1203).

fxlb:
Rebased without update for now. (The Autoconf & CMAKE part was committed)

guyharris:
Using `sizeof`, as this currently does, completely avoids any problem with fat binaries, etc..

fxlb:
> completely avoids any problem with fat binaries

Yes, I prefer to avoid any depandancy in relation to Ap**e, as they make their changes without any concertation.

infrastation:
So commit 170cbc8 will need to be reverted, correct?

fxlb:
> So commit [170cbc8] will need to be reverted, correct?

I'd prefer to keep it to have the same log when building libpcap / tcpdump for tests, like:

```
./configure [...] | grep 'checking size of'
[...]
checking size of void *... 4
checking size of time_t... 8
```

PR #1150, opened 2024-03-19
Update some tests files if the packet time is > 2038-01-19 03:14:07 UTC

fxlb:
This will avoid, for now, "[Error converting time]" messages even
when replacing 0x80000000 mask test on tv_sec by a test on tv_sec
negative in ts_date_hmsfrac_print() and ts_unix_print().

With this change, the outputs remain the same when building on 64-bit or
32-bit with 32 or 64 time_t size.

The updated time is: 2038-01-01 00:00:00.000000 (Un*x time 2145916800).


fxlb:
Any specific test files with some packet dates > 2038-01-19 03:14:07 UTC should be placed in a ".tests" file for conditional testing.

PR #1151, opened 2024-03-19
Add Github CI for Haiku r1beta4 using build_matrix.sh

korli:
clang 17 doesn't work as expected on x86.


infrastation:
Thank you for preparing these changes. Do you know if there is a feasible way to implement this using Buildbot and a standalone VM?

korli:
Sorry, I don't know much about buildbot.

infrastation:
For reference, on my old PC with 4 cores and ccache tcpdump build matrix in a Haiku VM took almost 70 minutes. Given a budget of 2000 minutes per month, that would be 28 builds per month.

Let me have a look and think about it.

korli:
Github CI does it in 26 minutes: https://github.com/korli/tcpdump/actions/runs/8349002774  What's the problem with Github actions?

fxlb:
Please don't use 'git merge', always use 'git rebase' (See CONTRIBUTING.md)

kallisti5:
https://www.haiku-os.org/guides/virtualizing/google

PR #1152, opened 2024-03-25
Disregard setlinebuf(3), always use setvbuf(3).

infrastation:
Let's confirm this passes CI as expected.


guyharris:
That dates back to...

...at least tcpdump 3.4, from the Old Days of "BSD vs. AT&T", with BSD Unix having `setlinebuf()` and System V having `setvbuf()`.  `setvbuf()` is in C90 (and probaby C89), so I don't expect this to have a problem, and don't see any need to continue to support platforms with `setlinebuf()` but not `setvbuf()`.

(There's still the problem that Microsoft's C library implements line buffering as full buffering; TShark works around that by implementing line buffering as "do an `fflush(stdout)` after printing the entire packet dissection.)

PR #1157, opened 2024-03-28
Include <fcntl.h> unconditionally

fxlb:
It is used for open() with is in POSIX.1-2001, POSIX.1-2008.

libpcap has included <fcntl.h> unconditionally since at least 1999.


infrastation:
Similar logic applies to `<net/if.h>`, which is also in POSIX.1-2001: libpcap sometimes checks for it, but includes the header regardless of `HAVE_NET_IF_H`, and it just works. tcpdump has checks and minds `HAVE_NET_IF_H`.

fxlb:
I'll take a look.

PR #1161, opened 2024-03-29
Include <net/if.h> unconditionally

fxlb:
It is used for if_indextoname() which is in POSIX.1-2001, POSIX.1-2008.

libpcap has included <net/if.h> unconditionally since at least 1999.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -344,7 +344,6 @@ cmake_pop_check_state()
 # Header files.
 #
 check_include_file(rpc/rpc.h HAVE_RPC_RPC_H)
-check_include_file(net/if.h HAVE_NET_IF_H)
 if(HAVE_RPC_RPC_H)
     check_include_files("rpc/rpc.h;rpc/rpcent.h" HAVE_RPC_RPCENT_H)
 endif(HAVE_RPC_RPC_H)
--- a/cmakeconfig.h.in
+++ b/cmakeconfig.h.in
@@ -60,9 +60,6 @@
 /* Define to 1 if you have the `rpc' library (-lrpc). */
 #cmakedefine HAVE_LIBRPC 1
 
-/* Define to 1 if you have the <net/if.h> header file. */
-#cmakedefine HAVE_NET_IF_H 1
-
 /* Define to 1 if you have the `openat' function. */
 #cmakedefine HAVE_OPENAT 1
 
--- a/configure.ac
+++ b/configure.ac
@@ -30,7 +30,7 @@ if test "$ac_cv_prog_cc_c99" = "no"; then
 fi
 AC_LBL_C_INIT(V_CCOPT, V_INCLS)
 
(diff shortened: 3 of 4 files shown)


fxlb:
Error in Windows:
`C:\projects\tcpdump\print-sll.c(31): fatal error C1083: Cannot open include file: 'net/if.h': No such file or directory`

fxlb:
Conditions on `_WIN32` would be no better than on `HAVE_NET_IF_H`. Closing.

infrastation:
When I suggested that, I forgot about Windows, as often happens.

infrastation:
Perhaps it would be useful to leave a comment somewhere to explain the purpose of the check for `<net/if.h>` if it remains.

That said, the only sound use case for `if_indextoname()` is either a live packet capture or decoding of a savefile that was written on the same host with the same index-to-name interface mapping (in other words, capture on Linux, decode on a different Linux or BSD/Haiku/illumos, and it will "work" and will print invalid interface names). So I suspect the interface name printing in `print-sll.c` should hinge on neither `HAVE_NET_IF_H` nor `_WIN32`, but on `__linux__`, then the header check can go away because `<net/if.h>` is present on Linux.

fxlb:
I have not a clear view of the "any" interface outside Linux. Please do the change.

guyharris:
> I have not a clear view of the "any" interface outside Linux.

macOS has one (macOS's libpcap also supports "all" for the same purpose); it's based on an undocumented mechanism, although both Apple libpcap code (licensed under the patent-clause-including APSL, unfortunately) and [non-Apple code](https://github.com/pandax381/pktap_demo) use it.

Unless and until we add pktap support, we don't need to worry about it.  *Apple's* tcpdump writes out a pcapng file when capturing on the "any" device and provides separate Interface Description Blocks (IDBs) for each interface, allowing a program reading the file to report the right interface names, even when run on a different machine.

The *right* way to handle printing interface names when reading a capture file is to add full pcapng support to libpcap, and, in tcpdump, do it with pcapng files and get the interface names from the file; that way, the right interface names are printed, even if you're not running tcpdump on the machine on which you did the capture.  That also doesn't require using `if_indextoname()`.

For now, as per what Denis said, given that the printing only works when you're doing a live capture or reading a live capture on the same machine (running the same OS, so don't dual-boot to FreeBSD or whatever and expect it to read a capture from when you were running Linux), we might as well just restrict that feature to Linux.

infrastation:
Not just the same OS, but the same booted instance of the OS with the same subset of interfaces: unconfigure a VLAN or a tunnel interface or unplug a USB network adapter or unload a PCI network card driver and the live index-to-name mapping no longer has all the indices that may have appeared in the packet capture. This data can expire very fast, unless in a stable environment with many identical hosts where, for example, ifindex 1 is always `lo` and ifindex 2 is always `eth0`, and there is nothing else. But for a live capture it is very user-friendly.

PR #1162, opened 2024-03-29
Fix incompatible pointer types with time functions calls on Windows

fxlb:
On Windows in a struct timestamp, tv_sec is a long not a 64-bit time_t.
```
The problem shows:
listening on \Device\NPF_Loopback, link-type NULL (BSD loopback), snapshot length 262144 bytes
    1  [localtime() or gmtime() couldn't convert the date and time].052255
       IP 10.0.0.10 > 224.0.0.251: igmp v2 report 224.0.0.251
    2  [localtime() or gmtime() couldn't convert the date and time].792000
       IP 10.0.0.10.138 > 10.0.0.255.138: NBT UDP PACKET(138)

The warnings with clang-cl were:
util-print.c(253,18): warning: incompatible pointer types passing
  'const long *' to parameter of type 'const time_t *'
      (aka 'const long long *') [-Wincompatible-pointer-types]
  253 |                 tm = localtime(&tv->tv_sec);
      |                                ^~~~~~~~~~~

util-print.c(255,15): warning: incompatible pointer types passing
  'const long *' to parameter of type 'const time_t *'
      (aka 'const long long *') [-Wincompatible-pointer-types]
  255 |                 tm = gmtime(&tv->tv_sec);
      |                             ^~~~~~~~~~~
```

### DIFF
--- a/util-print.c
+++ b/util-print.c
@@ -243,16 +243,32 @@ ts_date_hmsfrac_print(netdissect_options *ndo, const struct timeval *tv,
 	struct tm *tm;
 	char timebuf[32];
 	const char *timestr;
+#ifdef _WIN32
+	time_t sec;
+#endif
 
 	if (tv->tv_sec < 0) {
 		ND_PRINT("[timestamp < 1970-01-01 00:00:00 UTC]");
 		return;
 	}
 
+#ifdef _WIN32
+	/* on Windows tv->tv_sec is a long not a 64-bit time_t. */
+	sec = tv->tv_sec;
+#endif
+
 	if (time_flag == LOCAL_TIME)
+#ifdef _WIN32
+		tm = localtime(&sec);
+#else
 		tm = localtime(&tv->tv_sec);
+#endif
 	else
+#ifdef _WIN32
+		tm = gmtime(&sec);
+#else
 		tm = gmtime(&tv->tv_sec);
+#endif
 
 	if (date_flag == WITH_DATE) {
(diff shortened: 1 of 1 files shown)


guyharris:
> On Windows in a struct timestamp, tv_sec is a long not a 64-bit time_t.

...which means that, on Windows, the libpcap API has a Y2.038K problem that cannot be fixed without a source-incompatible and binary-incompatible change (on *all* platforms, 32-bit and 64-bit) to the libpcap API.

We'll need to add new APIs to fully support pcapng anyway; those APIs should avoid using `struct timeval`, so that they won't have this problem.  `struct timespec` is in C11, POSIX, and [Visual Studio apparently back to at least VS 2015](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/timespec-get-timespec32-get-timespec64-get1?view=msvc-140), so that's probably what we should do.

For now, this change is the right short-term fix.

guyharris:
Unfortunately, some of those platforms can't be tested for another 3 days (I guess we've been making too many changes lately), but, given that the changes are Windows-only, if they succeed on one UN\*X they'll probably succeed on all UN\*Xes, and it passed on AppVeyor, so that's enough.

fxlb:
Yes, it's a Windows only change.
(That's why I've modified the Gisle proposal)

infrastation:
Buildbot now runs Linux and FreeBSD 14.0 on AMD64. I am going to reduce the footprint on Cirrus CI in April.

PR #1164, opened 2024-04-02
[TEST ONLY] NTP: test a p_ntp_time_fmt() update

fxlb:
Fail on 32-bit?


fxlb:
More test to do with 64-bit time_t.

PR #1166, opened 2024-04-03
Makefile.in: Update the .c.o build rule

fxlb:
Don't use the hacks for the rather old SunOS 4.

It's helps to do VPATH builds, e.g. 32-bit, 64-bit in two directories.
(https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -64,12 +64,8 @@ RANLIB = @RANLIB@
 
 DEPENDENCY_CFLAG = @DEPENDENCY_CFLAG@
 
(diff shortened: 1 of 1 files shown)


fxlb:
@infrastation Please, could you test on your Solaris 11.4 CBE local VM?

infrastation:
Testing...

infrastation:
Build matrix passes using Sun C 5.15 and GCC 11.2 and Clang 11.0. Local libpcap is the latest master branch (with the same change already applied).

fxlb:
Thank you!

PR #1176, opened 2024-04-19
Print NTP kiss codes for NTP debugging

demsbjf8:
NTPv4 (RFC 5905, section 7.4) and SNTPv4 (RFC 4330, section 8) formalize that if the stratum value is zero, the refid _may_ contain a printable ("designed for character displays and log files"), left justified, zero filled ASCII string for status reporting and debugging ("kiss code"). Some kiss codes are defined in the RFCs, but the list may be modified or extended in the future, and unregistered kiss codes, e.g., for experimentation and development, are possible (and are being seen in the field).

Thus, if the stratum value is zero AND the refid field begins with a printable character, assume the refid field contains a kiss code and print it.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -24,6 +24,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       NetFlow: Use tcp_flag_values[] for TCP flags.
       NTP: Check that the entire extension field is in the capture even
         if it isn't printed
+      NTP: Print kiss codes relevant for NTP debugging.
       MPTCP: parse MPC data_len field, print flags from MP_CAPABLE option.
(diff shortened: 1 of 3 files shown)


fxlb:
Thanks for your proposal.
The RFC states "Kiss codes are encoded in four-character ASCII strings".
This should be added to the comment and the commit message, like:
`s/left justified, zero filled ASCII string/four-character, left justified, zero filled ASCII string/`

demsbjf8:
Thanks for considering this proposal!

Just force-pushed an update as per your suggestion.

13.09.2024 17:28:36 fxlb ***@***.***>:

fxlb:
CHANGES updated, rebased, merged. Thanks!

demsbjf8:
Thanks to you for picking this up!

PR #1183, opened 2024-05-16
Get rid of support for libpcap prior to 1.0.

guyharris:
This lets us remove a bunch of configure-time and compile-time tests.

WinPcap 4.1.3 is based on libpcap 1.0, but doesn't export all of the new APIs, so it won't work with code that uses any of the other APIs, which tcpdump does, so don't test with WinPcap.


fxlb:
```
Also in the man page man page, we could remove:
              The -D flag will not be supported if tcpdump was built  with  an
              older  version of libpcap that lacks the pcap_findalldevs(3PCAP)
              function.

And replace:
              If  the  -D flag is supported, an interface number as printed by
              that flag can be used as the interface argument, if no interface
              on the system has that number as a name.

by:
              An interface number as printed by the -D flag can be used as the
              interface argument, if no interface on the system has that number
              as a name.
```

guyharris:
> Also in the man page man page ...

Done.

fxlb:
man page, remove ?
```
              The  -U  flag will not be supported if tcpdump was built with an
              older version of libpcap that lacks  the  pcap_dump_flush(3PCAP)
              function.
```

guyharris:
> These changes need an entry in the change log and a note in the "Dependency on libpcap" section of README.md, also possibly in INSTALL.md.

I've done all three.

INSTALL.md
```diff
@@ -1,18 +1,37 @@
...
+If your system does not provide libpcap, or provides a version that does
+not support all of the libpcap 1.0 APIs, you will need to download the
+source for The Tcpdump Group version of libpcap; see [this
+file](README.md) for the location, and build and install that version.
+Either install libpcap with `make install` or make sure both the libpcap
+and tcpdump source trees are in the same directory.
+
+### On Windows
+
+You will need to intall both Npcap and the Npcap SDK; see [this
```
gvanem, line 23: Typo; `intall`.

guyharris: Fixed.

PR #1184, opened 2024-05-19
CI: Add TinyCC (aka TCC) support

fxlb:
Identification examples: tcc-0.9.27, tcc-0.9.28rc

There are currently warnings such as:
print-802_11.c:3317: warning: assignment discards qualifiers from pointer target type
print-esp.c:373: warning: function might return no value: 'hexdigit'
So use: TCPDUMP_TAINTED=yes

Works on linux-amd64.
Does not work on linux-armv7l with BUILD_LIBPCAP=yes / CMAKE=yes
(Segmentation fault, libpcap dynamically linked). Works with 0.9.28rc.

TinyCC can be found at https://bellard.org/tcc/,
https://repo.or.cz/r/tinycc.git or as package on some distros.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -35,6 +35,12 @@ print_cc_version
 # later warnings in the same matrix subset trigger an error.
 
 case `cc_id`/`os_id` in
+tcc-*/*)
+    # print-802_11.c:3317: warning: assignment discards qualifiers from pointer
+    #   target type
+    # print-esp.c:373: warning: function might return no value: 'ldp_pdu_print'
+    TCPDUMP_TAINTED=yes
+    ;;
 *)
     ;;
 esac
--- a/build_common.sh
+++ b/build_common.sh
@@ -63,7 +63,7 @@ print_sysinfo() {
 cc_version_nocache() {
     : "${CC:?}"
     case `basename "$CC"` in
-    gcc*|egcc*|clang*)
+    gcc*|egcc*|clang*|tcc*)
         # GCC and Clang recognize --version, print to stdout and exit with 0.
         "$CC" --version
         ;;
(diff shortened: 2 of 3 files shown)


fxlb:
> Something is wrong with linking in ...

With the setup: `BUILD_LIBPCAP=yes` / `CMAKE=yes`.

guyharris:
> print-802_11.c:3317: warning: assignment discards qualifiers from pointer target type

I cannot see why that would be reported.  That particular line assigns to `last_presentp`, declared with

    const nd_uint32_t *presentp, *last_presentp;

so, unless tcc doesn't apply `const` to both of those variables, that shouldn't cause a problem.

The rest of the `for` statement only "assigns to" something is assigning `last_presentp + 1` to `last_presentp` in the `last_presentp++` clause.

Am I missing something here?

fxlb:
With the setup: BUILD_LIBPCAP=yes / CMAKE=no: libpcap is staticallly linked.
With the setup: BUILD_LIBPCAP=yes / CMAKE=yes: libpcap is dynamically linked.
So the builds are not the same.
```
$ ldd /dev/shm/tcpdump_build_matrix.oI29k1Wu/bin/tcpdump
        linux-vdso.so.1 (0x00007ffd72fe8000)
        libpcap.so.1 => /dev/shm/tcpdump_build_matrix.oI29k1Wu/lib/libpcap.so.1 (0x00007f18cc800000)
```

guyharris:
This appears to be crashing with a SIGSEGV in https://ci.tcpdump.org/#/builders/18/builds/1304 (a build for pull request #1185) with tcc and dynamic linking with a built-from-source libpcap rather than the system libpcap.

guyharris:
See [this thread on a Raspberry Pi forum](https://forums.raspberrypi.com/viewtopic.php?t=329647&start=25), wherein [somebody says](https://forums.raspberrypi.com/viewtopic.php?t=329647&start=25#p1973747):

> The problem is that tcc doesn't produce position independent code suitable for use as shared object files. The Segmentation fault you are getting in the example code is because the pointer to the string literal that gets passed to puts() isn't correctly adjusted hence the pointer that puts() receives is pointing to memory that isn't in the program's address space.
> 
> I've forgotten a lot about how relocation is handled in elf files but from what I can tell looking at the generated code, as long as you don't use string literals or global variables in your library code then it will probably work. Basically only function addresses are relocated properly in shared objects produced by tcc.

fxlb:
Good catch.
tcc version 0.9.27 is quite old (2017).
With current tcc version 0.9.28rc builded on linux-armv7l, it's OK.
Test with:
`MATRIX_CC=/home/buildbot-worker/tmp/tinycc/tcc CFLAGS='-I/home/buildbot-worker/tmp/tinycc/include -L/home/buildbot-worker/tmp/tinycc' LDFLAGS=-L/home/buildbot-worker/tmp/tinycc MATRIX_BUILD_LIBPCAP=yes MATRIX_SMB=no MATRIX_CRYPTO=no ./build_matrix.sh`

guyharris:
> With current tcc version 0.9.28rc builded on linux-armv7l, it's OK.

The "rc" is encouraging, as in "if it's a release candidate, hopefully a new release will come out relatively soon".

fxlb:
Hopefully ... In the meantime, linux-armv7l now run tcc 0.9.28rc.

PR #1185, opened 2024-05-19
esp: mark unreached code as such.

guyharris:
Some compilers might not understand the declaration of the ndo_error member of an netdissect_options as pointing to a function that never returns; add a /* NOTREACHED */ comment in case they recognize that as a hint.

### DIFF
--- a/print-esp.c
+++ b/print-esp.c
@@ -368,6 +368,7 @@ static u_int hexdigit(netdissect_options *ndo, char hex)
 	else {
 		(*ndo->ndo_error)(ndo, S_ERR_ND_ESP_SECRET,
 				  "invalid hex digit %c in espsecret\n", hex);
+		/* NOTREACHED */
 	}
 }
 


guyharris:
> The same does not work in [this build](https://ci.tcpdump.org/#/builders/17/builds/1198):

I guess `/* NOTREACHED */` isn't handled by tcc, either, so I guess we just have to live with it producing bogus warnings.

fxlb:
Because the macro NORETURN is used for ndo_error(), we could apply this patch:
```patch
diff --git a/funcattrs.h b/funcattrs.h
index 21f3cc1da..28e4108b5 100644
--- a/funcattrs.h
+++ b/funcattrs.h
@@ -52,13 +52,14 @@
     || ND_IS_AT_LEAST_GNUC_VERSION(2,5) \
     || ND_IS_AT_LEAST_SUNC_VERSION(5,9) \
     || ND_IS_AT_LEAST_XL_C_VERSION(10,1) \
-    || ND_IS_AT_LEAST_HP_C_VERSION(6,10)
+    || ND_IS_AT_LEAST_HP_C_VERSION(6,10) \
+    || __TINYC__
   /*
    * Compiler with support for __attribute((noreturn)), or GCC 2.5 and
    * later, or some compiler asserting compatibility with GCC 2.5 and
    * later, or Solaris Studio 12 (Sun C 5.9) and later, or IBM XL C 10.1
    * and later (do any earlier versions of XL C support this?), or HP aCC
-   * A.06.10 and later.
+   * A.06.10 and later, or current TinyCC.
    */
   #define NORETURN __attribute((noreturn))
 
```

guyharris:
This appears to be crashing in https://ci.tcpdump.org/#/builders/18/builds/1304 with tcc and dynamic linking with a built-from-source libpcap rather than the system libpcap.

guyharris:
That appears to be a tcc issue, being discussed in #1184, so I'll commit this (it doesn't appear to do any harm, and we use `/* NOTREACHED */` elsewhere - either it's useful somewhere, even if only as a notation to humans, in which case we might as well use it, or it's not useful anywhere, in which case we might as well remove it everywhere).

guyharris:
> But there is another question:
> Why hexdigit() use ndo_error() and not e.g. ndo_warning() with an adapted return code?

As the comment says:

```
                 * Failed to initialize the cipher context.
                 * From a look at the OpenSSL code, this appears to
                 * mean "couldn't allocate memory for the cipher context";
                 * note that we're not passing any parameters, so there's
                 * not much else it can mean.
```

Perhaps we should review all cases where `ndo_error()` is passed `S_ERR_ND_MEM_ALLOC` and, if we can print a warning and fall back on not doing something that would use what would have been allocated, change them to do so.

fxlb:
In `hexdigit()` the ndo_error() code is `S_ERR_ND_ESP_SECRET`.

fxlb:
""invalid hex digit %c in espsecret\n""

guyharris:
See pull request #1186.

fxlb:
> Perhaps we should review all cases where `ndo_error()` is passed `S_ERR_ND_MEM_ALLOC` and, if we can print a warning and fall back on not doing something that would use what would have been allocated, change them to do so.

Yes.

PR #1186, opened 2024-05-19
esp: report a warning, not an error, for an invalid hex digit in a se…

guyharris:
…cret.

We already report a warning if the secret is too long; do so if there's an invalid hex digit in the secret.

See

https://github.com/the-tcpdump-group/tcpdump/pull/1185#issuecomment-2119211162

While we're at it, put the length of the string, in hex-digit pairs, into a size_t rather than an unsigned int, just in case (it's unlikely that it will overflow a 32-bit unsigned integer, but...).

### DIFF
--- a/print-esp.c
+++ b/print-esp.c
@@ -357,7 +357,7 @@ static void esp_print_addsa(netdissect_options *ndo,
 }
 
 
-static u_int hexdigit(netdissect_options *ndo, char hex)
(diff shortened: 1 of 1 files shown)


fxlb:
(*ndo->ndo_warning)(ndo, "secret is too big: %u\n", len);
len is now a size_t, format should be '%zu'.

fxlb:
I see, it's done.

PR #1188, opened 2024-05-21
erspan: add support for ERSPAN version III.

guyharris:


guyharris:
> This would be worth a change log entry. Also do you have any test captures?

Done.

tests/erspan-type-iii-1.out
```diff
@@ -0,0 +1,240 @@
...
+	erspan type3 session 101 bso Bad cos 0: 4b:b5:01:00:81:00 Unknown SSAP 0x18 > 00:00:ab:c5:00:00 Null Supervisory, Reject, rcv seq 111, Flags [Final], length 58
+   16  2017-02-12 03:55:41.583029 IP (tos 0x0, ttl 255, id 0, offset 0, flags [none], proto GRE (47), length 112)
+    192.168.1.172 > 192.168.1.249: GREv0, Flags [sequence# present], seq 47853, length 92
+	erspan type3 session 101 bso Bad cos 0: 4b:b5:01:00:81:00 Unknown SSAP 0x18 > 00:00:ab:c5:00:00 Null Supervisory, Reject, rcv seq 111, Flags [Final], length 58
+   17  2017-02-12 03:55:42.022149 IP (tos 0x0, ttl 255, id 0, offset 0, flags [none], proto GRE (47), length 108)
+    192.168.1.172 > 192.168.1.249: GREv0, Flags [sequence# present], seq 47854, length 88
+	erspan type3 session 101 bso Bad cos 0:  [|llc]
+   18  2017-02-12 03:55:42.048321 IP (tos 0x0, ttl 255, id 0, offset 0, flags [none], proto GRE (47), length 112)
+    192.168.1.172 > 192.168.1.249: GREv0, Flags [sequence# present], seq 47855, length 92
+	erspan type3 session 101 bso Bad cos 0: 4b:b6:ff:ff:ff:ff > 00:00:ab:c5:00:00, ethertype Unknown (0xffff), length 72:
```
fxlb, line 54: What are these `ethertype Unknown (0xffff)`?

guyharris: I don't know - and neither does Wireshark.

The capture came from [Wireshark issue #13400](https://gitlab.com/wireshark/wireshark/-/issues/13400).  Frame 18 *looks* as if it has an ARP frame inside a VLAN frame, with 81 00 03 61 08 06 in it, and there's an ff ff ff ff ff ff that might be a broadcast destination address, but it's not obviously an Ethernet packet, and it has an unknown frame type of 7 in the ERSPAN header.

guyharris: I've changed the ERSPAN III dissector to use the frame type field, and not to dissect the payload as Ethernet unless the type field's value is 0, in 4b6535c921d0323245ae8e75d9feb464f23ab06a.  That matches what Wireshark does.

fxlb: Thanks.

PR #1204, opened 2024-07-22
OSPF: print the Extended Link Opaque TLV and subTLVs

fenner:
This adds support for the OSPFv2 Extended Link Opaque TLV as defined in RFC7684 and subTLVs defined in RFC8665 for segment routing.


fenner:
The linux-armv7l  build failed because the disk was full on the worker.

guyharris:
> The linux-armv7l build failed because the disk was full on the worker.

Fixed by @infrastation.

PR #1209, opened 2024-08-04
Don't touch groups with -Z root.

marenamat:
In single-user namespaces, calling initgroups() is forbidden. This enables tcpdump to be compiled with forced privilege separation while keeping the ability to actually run in an isolated environment where the privilege separation is already done by other means.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -724,7 +724,10 @@ droproot(const char *username, const char *chroot_dir)
 				fprintf(stderr, "dropped privs to %s\n", username);
 		}
 #else
-		if (initgroups(pw->pw_name, pw->pw_gid) != 0 ||
+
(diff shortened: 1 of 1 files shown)


marenamat:
`unshare -nUfr tcpdump -Z root`

This command actually makes no sense to be run isolated but it reproduces the issue.

```
$ unshare -nUfr tcpdump -Z root
tcpdump: Couldn't change to 'root' uid=0 gid=0: Operation not permitted
$ unshare -nUfr ~/tcpdump/tcpdump -Z root
requested to not drop privs
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
…
```

The actual use case is to simulate routing between several virtual pseudomachines. It is all possible to be run by an unprivileged user but tcpdump built with default privilege separation refuses to work there because it's forbidden to call `setgroups` inside these not-even-containers.

PR #1210, opened 2024-08-05
[BGP] BGP Tunnel Encapsulation and SR-TE printing

fenner:
Print BGP SR-TE NLRI using the tunnel encapsulation, both from the base spec draft-ietf-idr-sr-policy-safi and using the types and names specified in the IANA registry for draft-ietf-idr-bgp-sr-segtypes-ext.

These documents have both recently been sent to the IESG for approval to be published as RFC.

While draft-ietf-idr-bgp-sr-segtypes-ext is proposed for Experimental status, the values from that document are just the types and names, there are no packet types.

See "IESG state" at
https://datatracker.ietf.org/doc/draft-ietf-idr-sr-policy-safi/
https://datatracker.ietf.org/doc/draft-ietf-idr-bgp-sr-segtypes-ext/


fenner:
I've refreshed the test file, because draft-ietf-idr-sr-policy-safi and draft-ietf-idr-bgp-sr-segtypes-ext are now in the RFC Editor queue. Technical changes at this time are extremely unlikely.

PR #1211, opened 2024-08-08
Fix Memory Leak and Improve Error Handling in `find_interface_by_number` Function

RoseSecurity:
## What and Why

This pull request addresses a potential memory leak and improves error handling in the `find_interface_by_number` function within `tcpdump.c`. The issue was identified when the `pcap_findalldevs` or `pcap_findalldevs_ex` function call failed, leading to an early exit without freeing the allocated memory for the device list (`devlist`). This pull request ensures proper memory management by freeing resources before exiting the function on error.

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -1164,12 +1164,12 @@ _U_
 #endif
         status = pcap_findalldevs(&devlist, ebuf);
         if (status < 0) {
-		      if (devlist != NULL) {
-			        pcap_freealldevs(devlist);
-		      }
-		      error("%s", ebuf);
-	        return NULL;
-      }
+  		      if (devlist != NULL) {
+	  		        pcap_freealldevs(devlist);
+		        }
+		        error("%s", ebuf);
+	          return NULL;
+        }
 
 	/*
 	 * Look for the devnum-th entry in the list of devices (1-based).
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -1162,15 +1162,14 @@ _U_
 		free(host_url);
 	} else
 #endif
-      status = pcap_findalldevs(&devlist, ebuf);
-
-    if (status < 0) {
-		  if (devlist != NULL) {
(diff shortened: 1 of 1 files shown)


tcpdump.c
```diff
@@ -1162,17 +1162,27 @@ _U_
 		free(host_url);
 	} else
 #endif
-	status = pcap_findalldevs(&devlist, ebuf);
-	if (status < 0)
+    status = pcap_findalldevs(&devlist, ebuf);
```
guyharris, line 1165: Please don't change indentation.

RoseSecurity: I apologize, my linter took over for me on that one. I hope this is better!

tcpdump.c
```diff
@@ -1162,17 +1162,26 @@ _U_
 		free(host_url);
 	} else
 #endif
-	status = pcap_findalldevs(&devlist, ebuf);
-	if (status < 0)
-		error("%s", ebuf);
+        status = pcap_findalldevs(&devlist, ebuf);
+        if (status < 0) {
+  		      if (devlist != NULL) {
```
guyharris, line 1167: Still needs the indentation adjusted.

tcpdump.c
```diff
@@ -1162,17 +1162,26 @@ _U_
...
+	  		        pcap_freealldevs(devlist);
+		        }
+		        error("%s", ebuf);
+	          return NULL;
```
guyharris, line 1171: Note that `error()` never returns, so that `return` statement will never be executed.

tcpdump.c
guyharris: Also, note that if `pcap_findalldevs()` returns an error indication, `devlist` will be null.

guyharris:
Already fixed in the main branch in  11e8f0a2ecaf6e391ab412b1b23ece0ddac2dd0cand in the 4.99 branch in 830c52e09f41ac82a25a201bf1fd042395f6c8b4.

PR #1212, opened 2024-08-28
msdp: do some additional bounds checks.

guyharris:
Make sure we don't run past either 1) the end of the packet data (using ND_TCHECK_LEN() for fields we don't fetch and print) or 2) the end of the TLV (with a check of length before fetching the T and L, and checks of L before fetching any further data).

Don't advance sp past the T and L, or decrement length for the T and L, before parsing the V, so we don't have to subtract 3 from sp in some cases.

Add some comments.

Define ND_LONGJMP_FROM_TCHECK to do new-style checks.

### DIFF
--- a/print-msdp.c
+++ b/print-msdp.c
@@ -20,6 +20,7 @@
 
 #include <config.h>
 
+#define ND_LONGJMP_FROM_TCHECK
 #include "netdissect-stdinc.h"
 
 #include "netdissect.h"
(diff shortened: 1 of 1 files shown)


fxlb:
Because there are now some ND_TCHECK*() macros uses, we should add `#define ND_LONGJMP_FROM_TCHECK` before `#include "netdissect.h"`.

guyharris:
> Because there are now some ND_TCHECK*() macros uses, we should add `#define ND_LONGJMP_FROM_TCHECK` before `#include "netdissect.h"`.

Done.

guyharris:
@fenner - do these changes look reasonable?

We might want to backport them to the 4.99 branch to get them into the next release.

PR #1223, opened 2024-09-28
Fix '-tt' option printing when time > 2106-02-07T06:28:15Z

fxlb:
Currently the printing with '-tt' option (unix time) is incorrect.

Some examples:
1) test: time_2106_overflow-tt
0.000000 IP 192.168.1.11.43966 > 209.87.249.18.53: UDP, length 56
Should be:
4294967296.000000 IP 192.168.1.11.43966 > 209.87.249.18.53: UDP, length 56
2) test: time_2107-tt
28315904.000000 IP 192.168.1.11.43966 > 209.87.249.18.53: UDP, length 56
Should be:
4323283200.000000 IP 192.168.1.11.43966 > 209.87.249.18.53: UDP, length 56

Two build examples:
64-bit build: tv->tv_sec has type '__time_t' (aka 'long').
32-bit build with _TIME_BITS=64: tv->tv_sec has type '__time64_t'
  (aka 'long long').

Using 'unsigned' cast is incorrect for these 64-bit data.

Thus convert to 'int64_t' and print with '"%" PRId64'.

Add two test cases (existing pcapng printed with -tt).


guyharris:
> Thus convert to 'long long' and print with %lld.

Or convert to `int64_t` and print with `"%" `PRId64"`.

fxlb:
> Or convert to `int64_t` and print with `"%" PRId64`.

Yes! Updated.

PR #1232, opened 2024-10-21
Add --skip option to skip some packets before writing or printing

fxlb:
With this change, we can write/print some contiguous packets from a file.
We can also skip some packets doing a live capture.

The '--skip 0' option is allowed to help some loop in a shell script.

Examples:

Skip 3 packets when printing:
tcpdump -#n --skip 3 -r in.pcap

Write the sixth packet, if any:
tcpdump --skip 5 -c 1 -r in.pcap -w out.pcap

Write up to 5 packets after skipping 3:
tcpdump --skip 3 -c 5 -r in.pcap -w out.pcap


infrastation:
Would piping `editcap` achieve the same?

infrastation:
"By default, it reads all packets from the infile and writes them to the outfile in pcapng file format. Use '-' for infile or outfile to read from standard input or write to standard output, respectively."

guyharris:
> Would piping `editcap` achieve the same?

Only if you have Wireshark installed.

infrastation:
I've been thinking about adding a variable of the same effect to the build matrix: if you know which setup number has a problem, you can just skip N setups before it and retry as close to the context of last failure as possible.

PR #1249, opened 2024-12-01
Warn that options -A, -x[x] and -X[X] are mutually exclusive

fxlb:
[skip ci]


infrastation:
...and update the man page date in the same go?

PR #1250, opened 2024-12-07
Fix two undefined behaviors for the pcap_loop() call

fxlb:
Limit the --skip argument to INT_MAX.
Limit The sum of -c and --skip arguments to INT_MAX.

Fix the regression in 3eab64d3: The '--skip 0' option is allowed to get
the first packet in some loop e.g. in a shell script.

The errors were:

tcpdump.c:2696:8: runtime error: implicit conversion from type 'int' of
  value -1 (32-bit, signed) to type 'u_int' (aka 'unsigned int')
  changed the value to 4294967295 (32-bit, unsigned)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior tcpdump.c:2696:8

tcpdump.c:2696:8: runtime error: implicit conversion from type 'u_int'
  (aka 'unsigned int') of value 4294967295 (32-bit, unsigned) to type
  'int' changed the value to -1 (32-bit, signed)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior tcpdump.c:2696:8

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -2056,7 +2056,7 @@ main(int argc, char **argv)
 
 		case OPTION_SKIP:
 			packets_skipped = parse_u_int("packet skip count",
-			    optarg, NULL, 1, UINT_MAX, 0);
+			    optarg, NULL, 0, INT_MAX, 0);
 			break;
 
 #ifdef HAVE_PCAP_SET_TSTAMP_PRECISION
@@ -2096,6 +2096,12 @@ main(int argc, char **argv)
 		warning("-A and -X[X] are mutually exclusive. -A ignored.");
 	if (ndo->ndo_xflag && ndo->ndo_Xflag)
 		warning("-x[x] and -X[X] are mutually exclusive. -x[x] ignored.");
+
+	if (cnt != -1)
+		if ((int)packets_skipped > (INT_MAX - cnt))
+			// cnt + (int)packets_skipped used in pcap_loop() call
+			error("Overflow (-c count) %d + (--skip count) %d", cnt,
+			      (int)packets_skipped);
 
 	if (Dflag)
 		show_devices_and_exit();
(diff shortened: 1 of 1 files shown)


guyharris:
> From: man pcap_loop "A value of -1 or 0 for cnt is equivalent to infinity"

Also from `man pcap_loop`:

```
BACKWARD COMPATIBILITY
       In libpcap versions before 1.5.0, the behavior when cnt was 0 was
       undefined; different platforms and devices behaved differently, so code
       that must work with these versions of libpcap should use -1, not 0, as
       the value of cnt.
```

In 1.5.0, I introduced the `PACKET_COUNT_IS_UNLIMITED()` and converted all the pcap-*.c files to use it, so that *any* value <= 0 is treated as "unlimited", but, before that, some pcap-*.c files were, as I remember, explicitly testing against -1. I think there may have been an email to the list reporting that issue (I can't find an obvious issue in GitHub for it).

guyharris:
The count argument to `pcap_loop()` is an `int`, so:

We should probably have the argument to `-c` and to `--skip` both be `uint64_t` values, defaulting to 0, and give an error if either are greater than `INT_MAX` or, if neither are greater than `INT_MAX`, their sum is greater than `INT_MAX`.

Then the argument to `pcap_loop()` would be:

- -1, if `--c` wasn't specified;
- the sum of the values of `-c` and `--skip``, otherwise.

tcpdump.c
```diff
@@ -2056,7 +2056,7 @@ main(int argc, char **argv)
 
 		case OPTION_SKIP:
 			packets_skipped = parse_u_int("packet skip count",
```
guyharris, line 2058: Maybe just make `packets_skipped` an `int`, and fetch it with `parse_int()`?

(And maybe rename it as `packets_to_skip`?)

guyharris: Yes, leave it as is for now.  In the long run, all counts should be unsigned and 64-bit, and the capture/read API should have callback routines that return a value, so they can report errors, do their own counting and say "the loop is done", etc..

PR #1252, opened 2024-12-08
Add routines to parse numeric command-line arguments, and use them.

guyharris:
They do various error checks; use them instead of strto*() or atoi().

### DIFF
--- a/tcpdump.c
+++ b/tcpdump.c
@@ -165,11 +165,7 @@ The Regents of the University of California.  All rights reserved.\n";
 #endif
 
 static int Bflag;			/* buffer size */
-#ifdef HAVE_PCAP_DUMP_FTELL64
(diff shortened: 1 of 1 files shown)


guyharris:
> The current code uses strtoul() to obtain the value of the `--skip` option. This allows you to enter hex values 0xnnnnn, which is useful for testing purposes. This update does not.

Fixed.

guyharris:
> Thanks. Should we allow other flags to have also hexa input values?

We could, but I'd treat that as another change; this one's supposed to preserve the UI (other than possibly catching out-of-range errors), and allowing at least some other flags to take hex (or octal...) values would be a UI enhancement that we should probably not separately in a separate commit.

Most people have 10 fingers, so allowing hex values in packet counts might not be useful, but hex byte counts/sizes and file offsets could be useful.

PR #1254, opened 2024-12-10
autotools, CMake: fix issues with snprintf test and sanitizers.

guyharris:
Avoid trying to cast negative values to unsigned types, or doing shifts of signed types, in order not to have the test program fail if we're building with undefined-behavior sanitizers enabled.

See the-tcpdump-group/libpcap#1396 for the equivalent libpcap issue.


guyharris:
> `s/llibpcap/libpcap/` in the commit message.

Done.

PR #1262, opened 2024-12-21
tests: Fix some tests with invalid microsecond packet timestamps

fxlb:
If the timestamp for packets are given in seconds and microseconds, a microsecond value > 999999 is invalid. Set it to 999999.


infrastation:
In real world the input data is not under control.  Why not print invalid timestamp values as "invalid" and let the malformed test files flex the code path?

infrastation:
Ah, alright.

PR #1267, opened 2024-12-22
RSVP improvements to tests as well as P2MP support

mstigge:
This PR makes as few improvements in the RSVP dissector.

1. Add a test pcap for the most common protocol messages, for P2P as well as P2MP.
2. Adjust rendering of the P2MP ID.
3. P2MP Secondary ERO and RRO support was missing, add that.

All three are in separate commits and I suggest merging them separately to the master branch. Steps 2 and 3 are modifying the expected test output added in step 1, illustrating the effect of their changes.

Initially I planned to make 3 separate PRs, but GitHub's stacking support seems to be lacking somewhat, so doing a single PR instead with separate commits. Please indicate if a different commit structure is preferred in your project.


mstigge:
The build failure on openbsd-mips64 was a timeout in an ipv6 test which looks to me like a case of flaky test infra instead of being related to my changes.

PR #1274, opened 2025-01-07
BOOTP: Use an uint16_t variable to get the result of a GET_BE_U_2()

fxlb:
This will fix this Visual Studio warning:
print-bootp.c(1009,18): warning C4242: =: conversion from uint16_t to uint8_t, possible loss of data

### DIFF
--- a/print-bootp.c
+++ b/print-bootp.c
@@ -1006,22 +1006,24 @@ rfc1048_print(netdissect_options *ndo,
 				 * URI: URI of the SZTP bootstrap server.
 				 */
 				while (len >= 2) {
-					suboptlen = GET_BE_U_2(bp);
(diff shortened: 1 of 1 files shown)


print-bootp.c
```diff
@@ -1005,28 +1006,31 @@ rfc1048_print(netdissect_options *ndo,
 				 * uri-length: 2 octets long; specifies the length of the URI data.
 				 * URI: URI of the SZTP bootstrap server.
 				 */
+				uint16_t suboptlen2;
```
infrastation, line 1009: Placing this declaration inside the `while()` eliminates the need for the additional curly braces.

infrastation:
The next build of libpcap on macOS should pass.

PR #1280, opened 2025-02-06
Reimplement the tests similarly to libpcap.

infrastation:
This looks complete to me less the following imperfections:
* The existing code has some provisions for Windows, which these changes try to preserve on paper, but have not been tested at all.
* The existing exit status code handling code uses `>> 8` on Windows and special macros on POSIX, although in libpcap `>> 8` seems to work correctly on POSIX.  This may mean that at some point the Perl `system()` semantics converged, so there is no point in having two test methods. This is a separate function now, it may be usedful to move it `TESTlib.pm` after confirming the current semantics.
* The additional "this is not working right" block looked duplicate to me, so I removed it. I understand the key things that matter is whether tcpdump ran cleanly (0 code, one-line `stderr`) and whether `diff` ran cleanly. If any of that failed, just fail the test reliably and leave detailed interpretation to humans.
* The new code does not redact the path to the input and output files. It has no sense so long as the script prints the absolute path to the source tree right in the beginning.

Originally I tried to make the changes to `TESTrun` as a series of focused small commits, but soon discovered that addressing one lot of issues makes the next lot of issues apparent and so on, so this way the file ends up almost entirely replaced anyway, but many parts get changed several times over and eventually deleted. Replacing the entire file at once seems to incur less noise.


infrastation:
Also this does not include any support for tcpdump timeout yet: I would like to improve that a bit in libpcap first. Other than that, this is going to be merged soon unless anyone objects.

infrastation:
That's a good point, thank you. Updated.

infrastation:
Found a couple more imperfections, so there will be another revision.

PR #1299, opened 2025-02-22
Fix "ip broadcast" with -f and other assorted improvements

infrastation:
This includes a follow-up to the-tcpdump-group/libpcap#1454.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -729,7 +729,7 @@ fi
 AC_CHECK_FUNCS(pcap_set_optimizer_debug)
 
 #
-# bpf_dump() moved from tcpdump to libpcap in libpcap 0.6, but not all
+# bpf_dump() moved from tcpdump to libpcap in libpcap 0.6, but some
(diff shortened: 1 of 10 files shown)


configure.ac
```diff
@@ -88,7 +93,11 @@ PKG_PROG_PKG_CONFIG
 #
 # Do we have the brew command from Homebrew?
 #
-AC_PATH_PROG([BREW], [brew])
+case "$host_os" in
```
guyharris, line 96: Homebrew also supports Linux...

...but I'm not sure whether anybody would use it rather than the native package manager that almost all if not all Linux distributions have, so there's probably no reason to test for it on Linux.

guyharris: But there's probably no problem with testing for ti.

infrastation: This eliminates some tests that are never true, as far as the developers of these optional dependencies see it.

PR #1302, opened 2025-03-17
DHCP: Fix a comment about formats for (un)signed longs/shorts data

fxlb:
l - unsigned longs (32 bits)
L - longs (32 bits)
s - unsigned shorts (16 bits)

Also:
DHCP: Fix print format for client-last-transaction-time option

### DIFF
--- a/print-bootp.c
+++ b/print-bootp.c
@@ -518,7 +518,7 @@ static const struct tok tag2str[] = {
 	{ TAG_TZ_STRING,	"aTZSTR" },
 	{ TAG_FQDN_OPTION,	"bFQDNS" },	/* XXX 'b' */
 	{ TAG_AUTH,		"bAUTH" },	/* XXX 'b' */
-	{ TAG_CLIENT_LAST_TRANSACTION_TIME, "LLast-Transaction-Time" },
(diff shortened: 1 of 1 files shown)


infrastation:
There are no tests for this code path, correct?

infrastation:
A time interval is always a non-negative amount of time, so an unsigned integer is very appropriate.

infrastation:
`s` stands for `uint16_t`, which seems consistent with `l` for `uint32_t` and `L` for `int32_t`. Are the option declarations the only aspect that is off?

infrastation:
Option 2 is a time zone offset, thus can be negative. Option 91 is a number of seconds since an event in the past, thus cannot be negative.

print-bootp.c
```diff
@@ -401,8 +401,8 @@ bootp_print(netdissect_options *ndo,
  * The first character specifies the format to print:
  *     i - ip address (32 bits)
  *     p - ip address pairs (32 bits + 32 bits)
- *     l - long (32 bits)
- *     L - unsigned long (32 bits)
+ *     l - unsigned long (32 bits)
+ *     L - long (32 bits)
  *     s - short (16 bits)
```
infrastation, line 406: unsigned short (16 bits)

PR #1305, opened 2025-03-18
DHCP: Add support for option 108 "IPv6-Only Preferred" (RFC 8925)

fxlb:
Add a test file with DHCP option 108 in both the request and the reply, provided by Bill.


infrastation:
I wonder how many more options that currently use `l`, `L` and `i` (zero or more 32-bit values) are specified as strictly 32-bit.

infrastation:
(The same stands for `p`, `s` and `B`.)

infrastation:
If a boolean option is specified as exactly one octet and it is not exactly one octet (0, 2, 3...), that means the option is malformed and ideally should be flagged in addition to being printed (if it is malformed, the usual next move is to see what's inside). Looks like a potential space for future improvements.

PR #1307, opened 2025-03-22
DHCP: Fix printing IP address pairs options

fxlb:
'p' is used by the options Policy Filter (21) and Static Route (33).
They have a minimum length of 8, and the length MUST be a multiple of 8.
Check that.

Add a test file for option 33 (lengths: 8, 16, 24, 3, 0), build with scapy.


print-bootp.c
```diff
@@ -707,6 +707,14 @@ rfc1048_print(netdissect_options *ndo,
 
 		case 'p':
 			/* IP address pairs */
+			/* this option should be N x 8 bytes long */
+			if (len < 8 || len % 8 != 0) {
+				ND_PRINT("%s[ERROR: length != N x 8 bytes]",
```
infrastation, line 712: Could be just `nd_print_invalid();`.

infrastation: Alright, one improvement at a time is good enough.

PR #1316, opened 2025-04-03
tcpdump.c: Fix a warning when HAVE_FORK and HAVE_VFORK are not defined

fxlb:
The warning was:
tcpdump.c:226:13: warning: unused function 'child_cleanup'
  [-Wunused-function] static void child_cleanup(int);


infrastation:
What is the supported platform that has neither of those?

infrastation:
At the very least it makes the forward declaration consistent.

PR #1317, opened 2025-04-03
Require config.h to be from the tcpdump build.

infrastation:


infrastation:
Could be done later if trivial.

infrastation:
Thank you.

PR #1329, opened 2025-06-17
ipv6: Support new TSO without HBH

gentoo-root:
Currently, BIG TCP IPv6 inserts a hop-by-hop extension header with a jumbo payload option to reflect the real length of the packet bigger than 65535 bytes. New kernels will drop this extension header and just calculate the packet length from skb->len, like it's currently done for BIG TCP IPv4.

Reflect the future kernel change in tcpdump and support parsing such packets.

Kernel ref: https://lore.kernel.org/netdev/20250617144017.82931-1-maxim@isovalent.com/


borkmann:
Cc @infrastation do you have a chance to take a look please? Thx

borkmann:
> Thinking back to the IPv4 code, it should probably be:
> 
> ```
>         if (len == 0) {
>                 uint8_t nh = GET_U_1(ip->ip_p);
> 
>                 if (nh == IPPROTO_TCP) {
>                         /* we guess that it is a TSO send */
>                         len = length;
>                         presumed_tso = 1;
>                 }
>         } else
> ```
> 
> To better show invalid non-TCP packets of length 0.

That would break the goal of having BIG TCP for encapsulation (vxlan, geneve). I think the IPv4 side is fine as it is, it is really only about addressing the IPv6 side to be similar to how the IPv4 dissector prints the length when it presumes TSO.

borkmann:
> Would it be correct to say that you are on the way to generalize BIG TCP into BIG IP, and so far this mechanism covers TCP, VXLAN and GENEVE only?

Not quite, the mechanism in here is still mainly for TCP flows. Today, the Linux kernel networking stack is doing GRO/GSO/TSO also for TCP traffic over vxlan/geneve tunnels (TSO in this case only for most modern 100G NICs). This works similarly with regards to the aggregation as described above (max aggregating into 64k super-sized packet). The patch series for the kernel which we'd like to merge is so that also TCP traffic over tunnels can benefit from the same BIG TCP optimization as it exists for non-tunneled traffic already. So in that sense its a not a generalization into BIG IP, but its still targeted at TCP, thus BIG TCP (just that in this case the outer header has UDP, the inner TCP).

gentoo-root:
> Thus "length %u [was 0, presumed TSO]" in commit https://github.com/the-tcpdump-group/tcpdump/commit/3465ec4a0c94f3abe774dfdb9874f966403e20ce is probably wrong.

I just pushed a new version of this patchset, that keeps the existing behavior for IPv4, allowing zero-len GSO packets ("presumed TSO"). I also see that Wireshark has the same behavior. I added new tests to affirm that behavior, but the zero-len packet is manually crafted, it's not a real-life capture.

If, however, we are sure that "was 0, presumed TSO" is wrong, and this is the consensus, I can simply drop this behavior and the corresponding test ipv4_tcp_http_xml_tso.

fxlb:
I propose to commit first #1404 to have consistent  `presumed BIG TCP` messages with IP and IPv6.
It ensures the next header is TCP or UDP (for Geneve/VXLAN) to minimize problem (ICMP, etc.).
It keeps for now `presumed TSO` (the unclear case).

print-ip6.c
```diff
@@ -466,14 +466,23 @@ ip6_print(netdissect_options *ndo, const u_char *bp, u_int length)
...
+					 * a Jumbo Payload option, but only
+					 * if it's actually >64k. Otherwise,
+					 * we have no payload, so set the
+					 * length to the IPv6 header length,
 					 * and change the snapshot length
 					 * accordingly.
 					 */
-					len = sizeof(struct ip6_hdr);
-					nd_change_snaplen(ndo, bp, len);
+					if (length > 65535) {
```
fxlb: Proposed change:
```patch
diff --git a/print-ip6.c b/print-ip6.c
index 1ae7ce02f..8ce15176a 100644
--- a/print-ip6.c
+++ b/print-ip6.c
@@ -474,7 +474,12 @@ ip6_print(netdissect_options *ndo, const u_char *bp, u_int length)
 					 * and change the snapshot length
 					 * accordingly.
 					 */
-					if (length > 65535) {
+					if (length > 65535 &&
+					    (nh == IPPROTO_TCP || nh == IPPROTO_UDP)) {
+						/* we guess that it is an offload
+						   because next header is TCP or
+						   UDP (for VXLAN or Geneve)
+						 */
 						len = length;
 						if (ndo->ndo_vflag)
 							ND_PRINT("[real length %u, presumed BIG TCP] ",

```

fxlb:
> could you please take another look given this small change is unfortunately blocking bigger work on the Linux kernel side

Update proposal done.

@gentoo-root @borkmann

PR #1333, opened 2025-06-26
fix: do not overwrite files while Cflag_count reaches Wflag

ManzeonYip:
When the number of files reaches the specified value, avoid cyclic packet capturing from overwriting the already captured files.


infrastation:
The currently documented behaviour is this:
```
       -W filecount
              Used in conjunction with the -C option, this will limit the num-
              ber  of  files  created to the specified number, and begin over-
              writing files from the beginning,  thus  creating  a  'rotating'
              buffer.
```

PR #1336, opened 2025-07-01
ICMP: make sure we don't run past the end of a TLV for RFC 5837 objects.

guyharris:
As reported to tcpdump-security.

While we're at it, fix the check for the interface name length - it includes the one-octet length of the length, as well as the length of the name, so it must be at least 1.

Add the capture file from the report.


guyharris:
> The reason for this difference is currently unknown.

One or the other is probably computing it incorrectly. (There's some other issue you reported a while ago where tcpdump is doing some ESP(?) decryption correctly and *Shark isn't - I need to look into that at some point.)

PR #1352, opened 2025-08-09
ICMP6 RA: recognise PIO PD Preferred flag

bonktree:
This is a 1-bit flag in the Prefix Information RA option defined by [RFC9762](https://tools.ietf.org/html/rfc9762).
If a compliant RA-accepting node receives a PIO with the P flag set, it is expected to ignore the A flag, if set, and make use of DHCPv6 Prefix Delegation to obtain a prefix and assign addresses from that prefix as described in [RFC9663](https://tools.ietf.org/html/rfc9663), not individual addresses from the PIO-specified prefix.


infrastation:
Do you have any packets with this bit set as a material for a new test?

PR #1357, opened 2025-08-31
Treat DLT_IPV4 and DLT_IPV6 like DLT_RAW.

guyharris:
LINKTYPE_IPV4 and LINKTYPE_IPV6 were assigned by error, as DLT_RAW suffices for raw IP packets; there's no need to have link-layer types for "IPv4 only, IPv6 is an error" and "IPv6 only, IPv4 is an error".

Just hand DLT_IPV4 and DLT_IPV6 packets to the same dissector as DLT_RAW packets.

Update tests to reflect that.


guyharris:
At this point my inclination is not to make this change, to document `LINKTYPE_IPV4` and `LINKTYPE_IPV6` as for use *only* when the other IP version is to be treated as an error, and indicate that, in *all* other circumstances, `LINKTYPE_RAW` should be used.

The history of `LINKTYPE_IPV4` and `LINKTYPE_IPV6` is in [this tcpdump-workers thread](https://seclists.org/tcpdump/2009/q1/193). In [my reply to that message](https://seclists.org/tcpdump/2009/q1/194), I noted the existence of `LINKTYPE_RAW`, and Darren Reed said, in [his reply to my reply](https://seclists.org/tcpdump/2009/q1/195), "And it sounds like DLT_RAW is what to use here - ok.", so it's not clear `LINKTYPE_IPV4` and `LINKTYPE_IPV6` were ever used. Not sure why I went ahead and created them, but here we are.

guyharris:
> > I used to assume the purpose of `DLT_IPV4` and `DLT_IPV6` is to indicate that the other IP protocol is not meant to appear in the capture.
> 
> I thought so too.
> 
> Currently tshark prints an error when there is an IPv4 packet in a 'Raw IPv6' capture file. But no error with an IPv6 packet in a 'Raw IPv4' capture file.

For some reason, Wireshark's IPv4 dissector checks for a version of 6 and hands those packets off to the IPv6 dissector.  I'll have to see what the reason for that is.

guyharris:
Closing.

PR #1358, opened 2025-09-04
fixing missing dependencies in Makefile

Meiye-lj:
This PR fixes an issue in the Makefile. Specifically, previously, any modifications of files like varattrs.h would not trigger a rebuild of print-ascii.o. The PR fixes this by including them as additional dependencies.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -65,7 +65,10 @@ RANLIB = @RANLIB@
 DEPENDENCY_CFLAG = @DEPENDENCY_CFLAG@
 
 .c.o:
-	$(CC) $(FULL_CFLAGS) -c -o $@ $<
+	$(CC) $(FULL_CFLAGS) -MD -MP -c -o $@ $<
+	@mv $(@:.o=.d) $(@:.o=.d.tmp)
+	@sed -e 's|.*:|$@:|' <$(@:.o=.d).tmp >$(@:.o=.d)
+	@rm -f $(@:.o=.d).tmp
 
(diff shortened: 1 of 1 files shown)


Meiye-lj:
> Do you have done `touch .devel`? See CONTRIBUTING.md, line 57.

I see. Using touch .devel and make depend will indeed rebuild correctly. However, this PR generates .d directly in the script, allowing inexperienced developers to correctly rebuild their targets locally without having to execute touch .devel.

guyharris:
> In this project the default build mode is optimized for automatic builds that retrieve the source tree once, build it once, copy the binaries once and discard the source tree

...including downstream vendor builds (Linux distributions, \*BSDs, macOS builds at Apple, probably Solaris builds at Oracle, maybe AIX builds at IBM).

And in the era when libpcap and tcpdump *weren't* commonly provided by vendors, many users downloaded the source and built libpcap and tcpdump just so they'd have tcpdump handy to capture and analyze network traffic, which is another case like that.

The dependency tracking is mainly of use to people modifying libpcap or tcpdump.

I don't know whether any projects that do dependency tracking by default do it with an explicit "make depend" step or with flags to the compiler used when doing normal builds, and don't know how many such projects there are. Wireshark does that; it currently uses CMake exclusively, and I *think* CMake has modules, or whatever it calls them, for various toolchains that, among other things, the knowledge of how to automatically generate dependencies. I forget how that was done when it supported solely autotools for building.

PR #1362, opened 2025-09-23
instrument functions: Satisfy bfd inclusion guard

corubba:
In [2012](https://sourceware.org/git/?p=binutils-gdb.git;a=blobdiff;f=bfd/bfd-in.h;h=ae8149a2;hp=d88bcb6c;hb=df7b86aa;hpb=134fa82e) binutils added an inclusion guard to `bfd.h`, requiring to include the automake `config.h` beforehand. This guard is triggered when enabling instrumentation and development mode:

```
% git clone https://github.com/the-tcpdump-group/tcpdump.git
% cd tcpdump
% touch .devel
% ./autogen.sh
% ./configure --enable-instrument-functions --enable-smb --quiet
./mkdep -c gcc -m -M -s . -DHAVE_CONFIG_H -I. fptype.c tcpdump.c print-smb.c smbutil.c instrument-functions.c <libnetdissect src list>
In file included from ./instrument-functions.c:18:
/usr/include/bfd.h:35:2: error: #error config.h must be included before this header
   35 | #error config.h must be included before this header
      |  ^~~~~
```

Note than this does \*not\* happen on Debian-based systems, because they [remove](https://salsa.debian.org/toolchain-team/binutils/-/blob/38415eb8/debian/rules#L1045-1048) that guard since [2012](https://code.launchpad.net/~doko/binutils/pkg-2.23-debian). That's also why the Ubuntu-based Linux CI is not affected. But on distributions using the vanilla source, like Arch Linux, it prevents the debug build.

Also a small update to the CONTRIBUTING docs, adding the required <del>automake</del> autoreconf invocation.

### DIFF
--- a/instrument-functions.c
+++ b/instrument-functions.c
@@ -15,7 +15,15 @@
 #include <string.h>
 #include <stdlib.h>
 #include <unistd.h>
+
+/*
+ * The binutils bfd header has a weird inclusion guard, requiring this
+ * macro to be defined. Some distributions patch the guard out, others
+ * don't.
+ */
+#define PACKAGE "binutils-bfd.h-guard-workaround"
 #include <bfd.h>
+#undef PACKAGE
 
 /*
  * Generate instrumentation calls for entry and exit to functions.


guyharris:
> In [2012](https://sourceware.org/git/?p=binutils-gdb.git;a=blobdiff;f=bfd/bfd-in.h;h=ae8149a2;hp=d88bcb6c;hb=df7b86aa;hpb=134fa82e) binutils added an inclusion guard to `bfd.h`, requiring to include the automake `config.h` beforehand.

Given that bfd.h is, as far as I can tell, a *public header file*, exported for use by *programs that use libbfd*, why on *earth* would they do *that*?  There is no guarantee, for example, that all code using libbfd *even has a config.h file*.

> Note than this does *not* happen on Debian-based systems, because they [remove](https://salsa.debian.org/toolchain-team/binutils/-/blob/38415eb8/debian/rules#L1045-1048) that guard since [2012](https://code.launchpad.net/~doko/binutils/pkg-2.23-debian).

I don't know why they did that - or why they didn't push back on the binutils maintainers, if, as I suspect is the case, they found that it caused problems to *users* of binutils (as opposed to *developers*of binutils).

guyharris:
> My reading of the [original binutils PR 14072](https://sourceware.org/bugzilla/show_bug.cgi?id=14072) (and its follow-ups [14243](https://sourceware.org/bugzilla/show_bug.cgi?id=14243) and [15920](https://sourceware.org/bugzilla/show_bug.cgi?id=15920)) is that the inclusion guard is designed as some binutils-internal thing, and they didn't/don't expect other projects to directly use/include it.''

[The "How To Use BFD" section of the libbfd documentation](https://sourceware.org/binutils/docs/bfd.html#How-It-Works) says "To use the library, include bfd.h and link with libbfd.a." I don't know who first made an exported library out of it - the binutils developers, one or more random Linux distributions, or somebody else, but whoever it was should have fixed that, either by making bfd.h a usable external header (moving that guard somewhere else), making a new *suitable* header for use by arbitrary programs using libbfd, or *documenting* that annoying requirement.

> I deemed it best to just include the existing `config.h` since tcpdump is using autoconf,

Or, rather, that it can use either autoconf or CMake, and both of them, as used in tcpdump, define a config.h file that defines the `PACKAGE_VERSION` macro (neither of them define `PACKAGE`).

> Not sure why the `netbsd-mips64` ci-build failed,

One reason appears to be "the builder for it sometimes gets memory errors or something such as that, causing it to randomly crash".

corubba:
There was no answer/reaction on the ML.

So there you go, updated the PR: Instead of including the tcpdump `config.h`, temporarily define the checked-by-bfd.h and unused-by-tcpdump `PACKAGE` macro, with a comment about the "why" for posterity.

fxlb:
Thanks for the update.
In commit message "That's also why the Ubuntu-based Linux CI is not affected."

Linux **CI** ?

fxlb:
Yes I know what is a "CI". I just not understand this sentence in the context.

PR #1373, opened 2025-10-16
VXLAN: Add two tests with filters

fxlb:
It uses vxlan.pcap (UDP port 4789) with filters:
1) vxlan
2) vxlan 100

The ouputs are the same as witout filters because all packets are VXLAN ones and all with VNI 100.

### DIFF
--- a/tests/TESTLIST
+++ b/tests/TESTLIST
@@ -536,6 +536,8 @@ dhcp-option-108	dhcp-option-108.pcapng	dhcp-option-108.out	-v
 
 # VXLAN tests
 vxlan  vxlan.pcap  vxlan.out -e
+vxlan_filter_vxlan vxlan.pcap vxlan.out -e vxlan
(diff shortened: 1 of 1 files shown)


infrastation:
The tests I added earlier still have quite some space for improvement because the savefile consists of VXLAN packets only, this way the test does not reliably prove that `vxlan` actually matches a VXLAN packet, it could as well match any packet. Similarly, since all packets have VNI 100, the test does not reliably prove that `vxlan 100` matches VNI 100 rather than any packet or any VXLAN packet, and that `vxlan 101` matches VNI 101 rather than no packets. The file needs to be more diverse to provide a more reliable proof. To add to that, there are no apply tests for any protocols inside VXLAN, which is the more complicated part of VXLAN, and no tests for double- or triple-encapsulated VXLAN.

fxlb:
> no tests for double- or triple-encapsulated VXLAN

You mean there could be VXLAN within VXLAN?

infrastation:
Commit c4e2dec has removed all filter expressions from tcpdump tests.

PR #1382, opened 2025-11-02
fix #1190 support to display packet content as UTF-8

### LINKED ISSUE #1190: Add support to interpret/display packet content as UTF-8 in addition to ASCII
It would be great to support displaying the content of a packet as UTF-8 in addition to ASCII.

lotabout:
This PR adds a new --utf8 option to enable UTF-8 character support.

When UTF-8 support is enabled, tcpdump will detect and display UTF-8 characters in the payload as-is when using the `-A` option.
Note that in `-X` mode, if a multi-byte character spans across two lines, it will appear on the first line, and a spaces will be padded on the next line.

## Tests

- Verified with several random PCAP files dumped using `-A` and `-x`, ensuring their MD5 checksums remain identical without --utf8.
- Manually tested using a UTF-8 sample PCAP on macOS.

[utf8.pcap.zip](https://github.com/user-attachments/files/23289840/utf8.pcap.zip)


infrastation:
How did it build in Appveyor then?

lotabout:
@gvanem Could you please help me give it another try on Windows? I’ve added Markus Kuhn’s implementation as a replacement on Windows.

lotabout:
@gvanem Please help to try again with latest code, it turns out that locale should be set correctly for `mbrtowc` to work properly.

（left: tcpdump, right: cat. Compiled on Windows 11 with VS 2022, Shell: PowerShell with `$OutputEncoding = [System.Text.Encoding]::UTF8` set）
image

image

print-ascii.c
```diff
@@ -56,6 +63,69 @@
...
+		return 0;
+	}
+
+	if (len == 0) {
+		/* Null character */
+		return 0;
+	}
+
+	/* Check if the wide character is printable */
+#if defined(_WIN32) || defined(_WIN64)
```
guyharris: Are there any compilers for Windows that define `_WIN64`, but don't define `_WIN32`, on "modern" Windows? ("modern" here means "not 16-bit Windows".)

infrastation:
I agree that for this new feature it is more important to implement it safe than quick. In particular, it should be impossible to weaponise it as is sometimes done using ANSI escape codes, see CVE-2025-46394 for one example. So this requires a bit more attention before it is declared safe and ready.

PR #1384, opened 2025-11-11
Install sysusers.d config for 'tcpdump' user/group

bluca:
Ubuntu, Fedora and other distro create the tcpdump user/group pair by default, to be used for unprivileged access to tcpdump:

https://git.launchpad.net/ubuntu/+source/tcpdump/tree/debian/tcpdump.postinst https://src.fedoraproject.org/rpms/tcpdump/blob/rawhide/f/tcpdump-sysusers.conf

Add a sysusers.d config file and install it. This configuration file will create the 'tcpdump' user/group pair on first boot/factory reset/package installation, without requiring manual scripts or actions. This is very important for image-based OSes, but it is useful in other cases too, to allow using declarative configuration instead of scripts.

RPM now has built-in support for sysusers.d too.

https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html

This is just a one-line config file, so if a system doesn't support sysusers.d, it is a harmless no-op (and can be deleted/not packaged too).

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -369,6 +369,7 @@ EXTRA_DIST = \
 	send-ack.awk \
 	smbutil.c \
 	stime.awk \
+	tcpdump-sysusers.conf \
 	tcpdump.1.in
 
 TEST_DIST= `git -C "$$DIR" ls-files tests`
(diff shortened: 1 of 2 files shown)


fxlb:
Each distribution has its own compilation process and, possibly, its own patches. 
(e.g. https://salsa.debian.org/debian/tcpdump/-/tree/master/debian?ref_type=heads)

It is the role of package maintainers to adapt projects to their distribution.
This systemd configuration could conflict with their process.
Therefore, I am not in favor of this PR.

bluca:
> This systemd configuration could conflict with their process.

Hi,

Thanks for having a look!

Speaking as a distribution maintainer (not of tcpdump specifically, but many other packages), I can attest that this does not conflict with any process.

In fact, generally speaking most maintainers prefer when upstream provides such generic configurations, as they are applicable everywhere, and contribute to reducing meaningless small differences between distributions. Opinions might vary of course, but this seems to me to be the majority view.

If an override is needed, it is trivial to do so, and all packaging systems allow to ignore upstream-shipped files and adding our own, so this creates no problems.

Here's a few examples of upstream projects shipping sysusers.d configs:

https://github.com/util-linux/util-linux/blob/master/misc-utils/uuidd-sysusers.conf.in
https://gitlab.freedesktop.org/dbus/dbus/-/blob/main/bus/sysusers.d/dbus.conf.in?ref_type=heads
https://github.com/polkit-org/polkit/blob/main/data/polkit.conf.in
https://github.com/deepin-community/stunnel4/blob/master/debian/stunnel4.sysusers#L4
https://github.com/flatpak/flatpak/blob/main/system-helper/flatpak.conf.in
https://github.com/fwupd/fwupd/blob/main/data/motd/fwupd.sysusers
https://github.com/stefanberger/swtpm/blob/master/src/swtpm-sysusers.conf.in

Of course this is your project so it's entirely up to you. Thanks for your consideration!

fxlb:
> I agree this detail belongs entirely to the packaging method of every binary distribution.

Closing, then?

infrastation:
Yes, let's not merge this. Pseudo-user specifics are completely up to the packager, and on OSes other than Linux the new file would be just noise.

PR #1386, opened 2025-11-13
Add MaxLinear GSW1xx DSA driver support

peteraxis:
This driver uses mdio port for it raw packets.
This is dependent on patch in libpcap.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -908,7 +908,7 @@ check_function_exists(pcap_dump_ftell64 HAVE_PCAP_DUMP_FTELL64)
 # we should check for it, so that we can use it if it's present.
 #
 # So we check for pcap_open() and pcap_findalldevs_ex() if 1) this isn't
-# macOS or 2) the the libpcap we found is not a system library, meaning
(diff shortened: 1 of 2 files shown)


guyharris:
A Google search for

[make a github change depend on a change in a different github repository](https://www.google.com/search?client=safari&rls=en&q=make+a+github+change+depend+on+a+change+in+a+different+github+repository&ie=UTF-8&oe=UTF-8)

didn't turn up anything that looked easy to do.

Presumably the libpcap change is https://github.com/the-tcpdump-group/libpcap/pull/1578; given that this change won't build without that change, that's probably sufficient.

peteraxis:
> It would be more useful to put the packet diagrams into tcpdump-htdocs. On this note, what are the two vertical lines to the right? One spans from a field boundary to a field boundary and another spans from a middle of a field to a field boundary. The labels next to them do not seems relevant.

I tried to fill in similar information as from the version it was based on. 
> 
> What is the reason the vendor is not allocating an EtherType value?

I don't know, I am not affiliated with MaxLinear. It is currently not allocated with IANA.

PR #1397, opened 2026-01-14
Fix overlapping 16-bit prints in OAM FM Loopback and AIS/RDI Location…

amyvarg:
Fix buffer overrun in oam_print() when printing ATM OAM fields.

The previous code iterated over the loopback_id, source_id, and failure_location
arrays and, on certain indexes, printed a 16-bit value starting at that index.
This could result in a buffer overrun on the last iteration, as it could read
past the end of the array.

Update the loops to increment by 2 and only access valid 16-bit aligned elements,
ensuring no out-of-bounds access occurs.

This change was made to address issues reported by Coverity static analysis.

No functional change except improved safety.


infrastation:
Thank you for reporting this bug. The commit message is empty, please make it right-sized.

I do not have a packet to test this code path and Coverity Scan is down right now, but code-wise it indeed looks an off-by-one error with no security implications.  For example, for the 12-byte Loopback Location ID the loop would be as follows:
```
i == 0, i % 2 == 0
i == 1, i % 2 == 1, prints bytes 1-2 (bug: byte 0 never printed)
i == 2, i % 2 == 0
i == 3, i % 2 == 1, prints bytes 3-4
i == 4, i % 2 == 0
i == 5, i % 2 == 1, prints bytes 5-6
i == 6, i % 2 == 0
i == 7, i % 2 == 1, prints bytes 7-8
i == 8, i % 2 == 0
i == 9, i % 2 == 1, prints bytes 9-10
i == 10, i % 2 == 0
i == 11, i % 2 == 1, prints bytes 11-12 (bug: out of bounds)
```
So in the original commit 27f414a the correct condition would be `idx % 2 == 0` in all three instances.

infrastation:
Also instead of "ATM OAM F5 Fault Management Loopback" please just say "ATM OAM F4 Fault Management" because:
* Loopback Location ID is only one of the three Fault Management function types affected by this bug.
* The function handles ATM OAM F4 only.

The latter is not obvious because `oam_print()` prints "OAM F5, vpi 96, vci 12288", but note the `VCI_OAMF4SC` and `VCI_OAMF4EC` case in `atm_print()`, which is the path into `oam_print()`, which means the output is supposed to show VCI 3 or 4. This is easy to see using the `-e` flag:
```
tcpdump -nev -tttt -r atm_pr1397.pcap -xx
reading from file atm_pr1397.pcap, link-type SUNATM (Sun raw ATM), snapshot length 18546945
2018-12-11 00:36:17.1159821166 (invalid us) Tx: VPI:0 VCI:4 OAM F5, vpi 96, vci 12288, payload [ user data, uncongested, SDU 0 ], clp 0, length 65618
	cell-type Fault Management (1), func-type Loopback (8)
	Loopback-Indicator Reply, Correlation-Tag 0x18020109
	Location-ID 011b 9e1b 0001 0101 5a01 0000 
	Source-ID   c329 1acd f08b 2f10 2709 646f 
	cksum 0x181 (incorrect)
	0x0000:  d300 0004 80e6 0300 0018 3018 0201 091b
	0x0010:  011b 9e1b 0001 0101 5a01 0000 c329 1acd
	0x0020:  f08b 2f10 2709 646f 0800 45d6 0143 00fe
	0x0030:  81b6 0c02 0101 0201 8102 0101 3001 01ac
```
Here "VPI:0" and "VCI:4" come from `atm_print()`, "vpi 96" and "vci 12288" come from `oam_print()`. The former takes VPI and VCI values from `sunatm_if_print()`, which parses [DLT_SUNATM](https://www.tcpdump.org/linktypes/LINKTYPE_SUNATM.html), note the bytes 1 (VPI 0x00) and 2-3 (VCI 0x0004) in the hex dump. The latter does not receive the values, instead it parses the provided buffer:
```c
    vpi = (cell_header>>20)&0xff;
    vci = (cell_header>>4)&0xffff;
    payload = (cell_header>>1)&0x7;
```
This method of parsing corresponds to the [ATM UNI](https://en.wikipedia.org/wiki/Asynchronous_Transfer_Mode#Cell_structure) header encoding, which is different, which is why the two functions print different values, which is a separate bug.

That bug manifests through the "OAM F5" default label in `oam_print()`, which normally would be dead code or even better would not be present in the first place (along with the supposedly duplicate printing of VPI and VCI), but in this case it makes its way into the output and causes more confusion. You are not supposed to fix that bug, but please mind not to copy the erroneous output.

amyvarg:
> Forgot to mention: when you have the next revision ready, please make it one nice commit (6 so far) with one nice commit message. Thank you.

Oops. I posted my most recent one before I saw this. I'm still new to using github but I'll see if I can get it figured out.

infrastation:
Thank you. I am going to denote the other bug separately. It probably won't be fixed anytime soon, but at lest the code will be a bit less misleading if anyone bases their work/research on it in future.

PR #1398, opened 2026-01-16
Build on QNX.

infrastation:
The `SA_RESTART` workaround has not been tested in any regard, other than it unblocks building of tcpdump and the resulting executable seems to work as expected.


infrastation:
This revision avoids `SA_RESTART` only, not the entire `sigaction()` branch.

PR #1403, opened 2026-01-28
introduce a set of SECURITY reporting details

mcr:
This adds a SECURITY.md file that details our views about security reporting.


SECURITY.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Use of LLMs ("AI")
+
+We do not accept reports generated by LLMs.
+We do not consent to your using our project to help train your LLM to do reports.
+
+## Reporting
+
+Send an e-mail to security@tcpdump.org.
+This is a closed list, and which you will receive communication from the project members.  If you have a spam filter that requires any action on our behalf to confirm emails, then we will ignore you.
```
guyharris: "and which you will receive..." or "*from* which you will receive..."?

infrastation:
IMO, it would be wrong to bake these details separately into every repository (where the contents would depend on the commit/release the user is looking at). There has been a [dedicated page](https://www.tcpdump.org/security.html) exactly for this kind of information, so it would be better to clarify the current policy there, once. Arguably, the READMEs and man pages could be updated to point to the page instead of giving an e-mail address.

Also without [skip ci] every push generates roughly a dozen build requests in the CI.

PR #1431, opened 2026-04-19
tcp: add support for diagnostic payload

tuexen:
TCP RST diagnostic payload is currently specified in [draft-boucadair-tcpm-rst-diagnostic-payload-17](https://www.ietf.org/archive/id/draft-boucadair-tcpm-rst-diagnostic-payload-17.html). This ID is accepted as a WG document in TCPM.


fxlb:
Thanks for the PR.
Printing e.g. `[RST 9/0]` is a little cryptic. Why not  `[RST Not Authorized (9), pen 0]`?
Please add a pcap test file with some RST examples (See CONTRIBUTING.md item 10).

tuexen:
> Thanks for the PR. Printing e.g. `[RST 9/0]` is a little cryptic. Why not `[RST Not Authorized (9), pen 0]`?

I am not sure if the table is already stable. So what about initially: `[RST: code 9, pen 0]`. As far as I can tell, people will in particular use this extension with pen specific codes and pen != 0. The `:` after the `RST` would distinguish the output from the ASCII string.
> Please add a pcap test file with some RST examples (See CONTRIBUTING.md item 10).

Will do.

tuexen:
@fxlb Are you fine with the PR?

fxlb:
> As far as I can tell, people will in particular use this extension with pen specific codes and pen != 0.

Really ?
```
8.  Security Considerations

   [...]

   The presence of vendor-specific reason codes may be used to
   fingerprint hosts.  Such a concern does not apply if the reason codes
   are taken from the IANA-maintained registry.  Implementers are, thus,
   encouraged to register new codes within IANA instead of maintaining
   specific registries.
```

tuexen:
> > As far as I can tell, people will in particular use this extension with pen specific codes and pen != 0.
> 
> Really ?
> 
> ```
> 8.  Security Considerations
> 
>    [...]
> 
>    The presence of vendor-specific reason codes may be used to
>    fingerprint hosts.  Such a concern does not apply if the reason codes
>    are taken from the IANA-maintained registry.  Implementers are, thus,
>    encouraged to register new codes within IANA instead of maintaining
>    specific registries.
> ```

As far as I know, Netflix is considering to use this with vendor specific codes for traffic, which is not customer facing. To be more precise, this is for application triggered RST segments using the `TCP_RST_REASON_CODE` socket option. I don't think in this case the problem is fingerprinting, the problem is more if the RST segments are delivered correctly to the receiver.
In general, I agree with the cited text for RST segments for which the sending is triggered by the TCP stack.

fxlb:
> I am not sure if the table is already stable.

I don't know but Wireshark currently use it (https://gitlab.com/wireshark/wireshark/-/blob/master/epan/dissectors/packet-tcp.c#L147-168).

> So what about initially: `[RST: code 9, pen 0]`

Why not for ndo_vflag == 1 and IANA messages registry use for  ndo_vflag > 1.

tuexen:
> > I am not sure if the table is already stable.
> 
> I don't know but Wiresherk currently use it (https://gitlab.com/wireshark/wireshark/-/blob/master/epan/dissectors/packet-tcp.c#L147-168).

If you don't care about follow-up commits, I can put the table in.
> 
> > So what about initially: `[RST: code 9, pen 0]`
> 
> Why not for ndo_vflag == 1 and IANA messages registry use for ndo_vflag > 1.

OK. Since the usage of the IANA table entries implies pen == 0, something like

1. `[RST: Unexpected SYN in the window]` for `pen==0` and IANA registered code and `ndo_vflag > 1`.
2. `[RST: code 1, pen 2238]` for other cases as long as `ndo_vflag >= 1`.

Sounds like a good solution for me.

tuexen:
@fxlb : Do you have more suggestions to improve the code?

fxlb:
1. `switch (code)` could be replaced by a `struct tok` and a tok2str() call.
(See CONTRIBUTING.md "Code style and generic remarks" item 8)

3. Even if the draft states "Value 0 is reserved and MUST NOT be used."
the "Reserved" (code 0) must be added. (Error, fuzzing, etc.)

4. s/"Illegal option length"/"Illegal Option"/

fxlb:
(There is a little inconsistency in the draft where all reason-code in 9.1 are decimal and all "5. Some Examples" reason-code values are hexa.)

tuexen:
> 1. `switch (code)` could be replaced by a `struct tok` and a tok2str() call.
>    (See CONTRIBUTING.md "Code style and generic remarks" item 8)

Done. Thanks for the hint. Much simpler.

> 2. Even if the draft states "Value 0 is reserved and MUST NOT be used."
>    the "Reserved" (code 0) must be added. (Error, fuzzing, etc.)

Done.

> 3. s/"Illegal option length"/"Illegal Option"/

Not done, since this is a mistake in the ID. It needs to be fixed there. RFC 9293 (and RFC 793) describes the situation of an illegal option length, not an illegal option. Will take care of getting it fixed in the ID. I also use a consistent way of capitalization, which needs also to be done in the ID.
Anyway, if the table changes in the future, I'll provide a patch to keep the code in sync with the spec.

tuexen:
> (There is a little inconsistency in the draft where all reason-code in 9.1 are decimal and all "5. Some Examples" reason-code values are hexa.)

I see. I don't see it as a problem, since natural numbers are the same, no matter to which base they are represented. I can see that you want IANA to deal with decimal numbers and show fields in a packet as hexadecimal. But I also see your point. Thanks for making me aware of the difference.

tuexen:
@fxlb Any additional suggestions?

fxlb:
> since natural numbers are the same

But mixing decimal and hexadecimal numbers can lead to errors, for example:
```
   Figure 4 illustrates an RST diagnostic payload that is returned by a
   peer that resets a TCP connection for a reason code 1234 defined by a
   vendor with the private enterprise number 32473 (0x7D9).

      |             0x33AA            |              0x4DE            |
      |                             0x7D9                             |
```
hexa(1234) == 0x4D2 (!=  0x4DE)
hexa(32473) == 7ED9 (!= 0x7D9)

tuexen:
> > since natural numbers are the same
> 
> But mixing decimal and hexadecimal numbers can lead to errors, for example:
> 
> ```
>    Figure 4 illustrates an RST diagnostic payload that is returned by a
>    peer that resets a TCP connection for a reason code 1234 defined by a
>    vendor with the private enterprise number 32473 (0x7D9).
> 
>        0                   1                   2                   3
>        0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
>       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
>       |             0x33AA            |              0x4DE            |
>       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
>       |                             0x7D9                             |
>       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
> ```
> 
> hexa(1234) == 0x4D2 (!= 0x4DE) hexa(32473) == 7ED9 (!= 0x7D9)

Thanks for letting me know. Will make sure the authors of the ID will fix it.

tuexen:
@fxlb Any more comments regarding the PR?

tuexen:
@fxlb: the editorial issues with the ID should be addressed by https://github.com/boucadair/draft-boucadair-tcpm-rst-diagnostic-payload/pull/19. I also updated the reference in the code to the WG document.

Any further actions required by me or are you fine with the PR?

fxlb:
> Any further actions required by me or are you fine with the PR?

Thanks for the updates.

"ABORT Process" in the new draft and "ABORT process" in the PR.
"Malformed Message" in the new draft and "Malformed message" in the PR.

Does the project need to be updated again? or the PR?

Also need a rebase from the 8 commits to one clean commit.

tuexen:
> > Any further actions required by me or are you fine with the PR?
> 
> Thanks for the updates.
> 
> "ABORT Process" in the new draft and "ABORT process" in the PR. "Malformed Message" in the new draft and "Malformed message" in the PR.
> 
> Does the project need to be updated again? or the PR?

https://github.com/boucadair/draft-boucadair-tcpm-rst-diagnostic-payload/pull/20/changes

> 
> Also need a rebase from the 8 commits to one clean commit.

Done.

tuexen:
@fxlb Anything else needed from my side?

fxlb:
Merged, thank you!

tuexen:
> Merged, thank you!

Thank you very much for your support.

PR #1434, opened 2026-04-27
isakmp: validate item_len before subtracting sizeof(ikev2_id)

Sebasteuo:
In ikev2_ID_print(), item_len is read from the packet and used
directly in the subtraction:

    idtype_len = item_len - sizeof(struct ikev2_id);

The caller only guarantees item_len > 4, but sizeof(struct ikev2_id)
is 8. When item_len is 5, 6, or 7, the subtraction underflows,
producing a large value that is then passed to nd_printjn() and
rawprint().

Add an explicit bounds check before the subtraction and print
"[payload too short]" when item_len is less than sizeof(struct ikev2_id).

Regression tests included:
- ikev2-id-normal.pcap: correct behavior (item_len=9)
- ikev2-id-short.pcap: underflow case (item_len=5)

Reported in: security@tcpdump.org thread "[security] ISAKMP: 
Integer underflow in idtype_len in ikev2_ID_print()"


fxlb:
Thanks for the PR.
Perhaps "Author: Tu Name <tu-email@ejemplo.com>" should be updated.

Sebasteuo:
@fxlb Thanks for the review. I have updated the author information 
and force-pushed the branch.

fxlb:
The test lines in TESTLIST have been moved to facilitate backport. Rebased, merged, thanks.

PR #1441, opened 2026-05-28
print-ospf: guard RI LSA TLV length before subtraction

TristanInSec:
The RI LSA TLV dispatch passes `tlv_length - 4` to
`ospf_print_ri_lsa_sid_label_range_tlv()` for TLV types 9 (SID/Label
Range) and 14 (SR Local Block) without checking `tlv_length >= 4`.

When `tlv_length` is 0–3, the unsigned subtraction wraps to a large
value, causing the callee's `while (tlv_length != 0)` loop to read
past the TLV boundary into adjacent packet data.

The fix adds a `tlv_length < 4` guard that prints a bogus-length
diagnostic and returns, consistent with other TLV handlers in the
same function (e.g. lines 818/847).


fxlb:
Thank you!

PR #1453, opened 2026-07-31
Add "geneve" as an option for -T.

alxn:
I have the sad situation where I'm running Geneve on a port other than `6081`, so I manually reworked the same pattern as `-T vxlan`.


fxlb:
Thank you for your suggestion. That sounds fine to me.
If no objection, it will be merged in a few days.

fxlb:
Merged, thanks.

PR #1456, opened 2026-08-17
Add AmigaOS 4 to the platforms list

polluks:
version 3.8.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ tcpdump compiles and works on at least the following platforms:
 In the past tcpdump certainly or likely worked on the following platforms:
 
 * 4.3BSD
+* AmigaOS 4
 * BSD/386, later BSD/OS
 * DEC OSF/1, later Digital UNIX, later Tru64 UNIX
 * DOS


infrastation:
What does `tcpdump -D` print?

guyharris:
So can the *current* code be compiled, with *no* changes, and run on AmigaOS 4?

If not, then it's not a platform *we* support, so we'd need to make whatever changes are required to make it compile and run.

Are there any changes to *libpcap* needed to make it compile and run? If so, what are *those* changes?

infrastation:
Is this the tcpdump that is a part of the [Roadshow TCP/IP stack for Amiga](http://roadshow.apc-tcp.de/index-en.php)? The SDK for the latter includes source code for libpcap 0.8.1 and tcpdump 3.8.1, and Roadshow implements BPF.

polluks:
indeed

infrastation:
So what does `tcpdump -D` print?

polluks:
1.RTL8139
2.loopback

guyharris:
OK, so it appears to have a libpcap with live capture support rather than a libpcap that supports only reading capture files.

I assume that "indeed" is a "yes" answer to "Is this the tcpdump that is a part of the [Roadshow TCP/IP stack for Amiga](http://roadshow.apc-tcp.de/index-en.php)?"

polluks:
Roadshow is part of AmigaOS 4; therefore tcpdump is officially part the operating system.

fxlb:
Result of `tcpdump --version  `?

polluks:
> Result of `tcpdump --version `?

See first posting
```
tcpdump version 3.8
libpcap version 0.8
Usage: tcpdump [-aAdDeflLnNOpqRStuUvxX] [-c count] [ -C file_size ]
 [ -E algo:secret ] [ -F file ] [ -i interface ] [ -r file ]
 [ -s snaplen ] [ -T type ] [ -w file ] [ -y datalinktype ]
 [ expression ]

infrastation:
Thank you.

PR #1457, opened 2026-08-18
Add a header file that defines a "byte-swap a 32-bit integer" macro.

guyharris:
Remove the macro from dissectors that use it, and have them include the header file.

Go with the renamed macro from libpcap, and declare the variables on which it's used to be uint32_t.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -269,6 +269,7 @@ HDR = \
 	appletalk.h \
 	ascii_strcasecmp.h \
 	atm.h \
+	byteswap.h \
 	chdlc.h \
 	compiler-tests.h \
 	cpack.h \
--- a/byteswap.h
+++ b/byteswap.h
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2026
+ *	The Tcpdump Group and contributors.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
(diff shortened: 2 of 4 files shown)


infrastation:
Please make the new file say
```
 * Copyright (c) 2026
 *      The Tcpdump Group and contributors.  All rights reserved.
```
and use a BSD-2-Clause or BSD-3-Clause. There's a lot of pending BSD clarification work pending, so let's at least not complicate it.

Also I can add a commit with byte swap builtins to libpcap, if you want to copy that in the same go.

guyharris:
> Please make the new file say

Done, with a [BSD 2-Clause "Simplified" License](https://spdx.org/licenses/BSD-2-Clause.html).

Should we use SPDX short identifiers, instead?

> Also I can add a commit with byte swap builtins to libpcap, if you want to copy that in the same go.

If you have that change ready, go ahead. Update the commit message, and the title and first comment of the pull request, as appropriate.

infrastation:
Short identifiers are being introduced, typically after it is clear the complete boilerplate is correct rather than copied and pasted from another file. This turned out to be a bigger job than I thought and due to a number of higher-priority interruptions even the clean-up in the-tcpdump-group/libpcap#1441 has not been completed yet. For now it would be sensible enough not to make it worse.

infrastation:
Also `s/header fine/header file/`.

guyharris:
> Also s/header fine/header file/.

Done.

infrastation:
Also `netdissect-stdinc.h` defines i386-specific `__ntohl()` and `__ntohs()`, which may have already been absorbed into compiler builtins.

